Issue #7, opened 2022-06-10
Timeout duration unit is nanosec


heqingy:
See this line: https://github.com/databricks/databricks-sql-go/blob/5fd0604f4034c1cd5bb9bf5a2ac6bbf2aa437780/driver.go#L143

nanosec is not very common for that parameter given the network latency is mostly at ms level and smaller values are not very meaningful

PR #8, opened 2022-06-10
Use millisecond as the unit of timeout

heqingy:
Fixing https://github.com/databricks/databricks-sql-go/issues/7
In the current implementation, the timeout parameter is converted using `strconv.Atoi` that can only parse max value of int32, while the unit is nanosec, which means the biggest timeout can only be ~2 seconds. 

This PR changes the timeout unit to be millisecond and updates the README accordingly.


arikfr:
Thanks!

PR #9, opened 2022-06-11
Escape Parameters

heqingy:
In the current implementation, the query
```sql
SELECT *
FROM table_name
WHERE string_column = ?;
```
with arg list `[]any{"a'b'c"}` will be translated into the following invalid query string:
```sql
SELECT *
FROM table_name
WHERE string_column = a'b'c;
```

This PR escapes string, time and turn the query into:
```sql
SELECT *
FROM table_name
WHERE string_column = 'a''b''c';
```

Referring to how python connector escape strings:
https://github.com/databricks/databricks-sql-connector/blob/a7af04c8a0df274b443fcc5ca6b40457019ab5ce/src/databricks/sql/utils.py#L126

### DIFF
--- a/escaper.go
+++ b/escaper.go
@@ -22,7 +22,7 @@ func EscapeArg(arg driver.NamedValue) (string, error) {
 	case string:
 		return fmt.Sprintf("'%v'", strings.ReplaceAll(v, "'", "''")), nil
 	case time.Time:
-		return v.Format(KTimeFmt), nil
+		return "'" + v.Format(KTimeFmt) + "'", nil
 	default:
 		return "", fmt.Errorf("unsupported parameter type %T for value %v", v, v)
 	}
--- a/escaper.go
+++ b/escaper.go
@@ -1,0 +1,29 @@
+package dbsql
+
+import (
+	"database/sql/driver"
+	"fmt"
+	"strings"
+	"time"
+)
+
+const (
+	KTimeFmt = "2006-01-02T15:04:05.999-07:00"
+)
+
+func EscapeArg(arg driver.NamedValue) (string, error) {
+	switch v := arg.Value.(type) {
+	case int64:
+		return fmt.Sprintf("%v", v), nil
+	case float64:
+		return fmt.Sprintf("%v", v), nil
+	case bool:
+		return fmt.Sprintf("%v", v), nil
+	case string:
+		return fmt.Sprintf("'%v'", strings.ReplaceAll(v, "'", "''")), nil
+	case time.Time:
+		return v.Format(KTimeFmt), nil
+	default:
+		return "", fmt.Errorf("unsupported parameter type %T for value %v", v, v)
+	}
+}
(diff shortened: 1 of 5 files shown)


escaper.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"database/sql/driver"
+	"fmt"
+	"strings"
+	"time"
+)
+
+const (
+	KTimeFmt = "2006-01-02T15:04:05.999-07:00"
```
arikfr, line 11: What's the `K` prefix for?

escaper.go
```diff
@@ -0,0 +1,29 @@
...
+	case float64:
+		return fmt.Sprintf("%v", v), nil
+	case bool:
+		return fmt.Sprintf("%v", v), nil
+	case string:
+		return fmt.Sprintf("'%v'", strings.ReplaceAll(v, "'", "''")), nil
+	case time.Time:
+		return "'" + v.Format(KTimeFmt) + "'", nil
+	default:
+		return "", fmt.Errorf("unsupported parameter type %T for value %v", v, v)
```
arikfr, line 27: Should we just fallback to some default? Or in theory these are the only types supported in Golang?

arikfr (commented):
Looks good, thanks! Added some questions.

escaper.go
heqingy: I used to name constants with leading `k`/`K`. what's the preferred way of naming constants here?

escaper.go
heqingy: > // Value is a value that drivers must be able to handle.
// It is either nil, a type handled by a database driver's NamedValueChecker
// interface, or an instance of one of these types:
//
//   int64
//   float64
//   bool
//   []byte
//   string
//   time.Time
//
// If the driver supports cursors, a returned Value may also implement the Rows interface
// in this package. This is used, for example, when a user selects a cursor
// such as "select cursor(select * from my_table) from dual". If the Rows
// from the select is closed, the cursor Rows will also be closed.
https://github.com/golang/go/blob/master/src/database/sql/driver/driver.go#L50

I've supported the listed ones except for `//   []byte` (I doubt if anyone is using it in DBSQL)

escaper.go
arikfr: Looking at other places where constants were used in the code and general golang guidance, I think this should be `TimeFmt` or even `timeFmt`.

escaper.go
arikfr: But `Value` is in fact `any` so the end user might pass other types. I wonder if we should just fallback to to converting unknown types to strings?

I tried to see what other drivers are doing and MySQL for example, while also returns an error when stumbling into an unfamiliar type it seems to support more types than this list (https://github.com/go-sql-driver/mysql/blob/bcc459a906419e2890a50fc2c99ea6dd927a88f2/connection.go#L198).

But anyway, I'm fine with keeping it as is and changing if we get further feedback.

PR #12, opened 2022-06-17
add preview to the readme file

moderakh:
add preview text the readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-# Preview Go Databricks SQL Driver
+# Go Databricks SQL Driver (Beta)
 
 Databricks SQL driver for Go's [database/sql](https://golang.org/pkg/database/sql) package.
 
-**NOTE: This Driver is in Preview mode.**
+**NOTE: This Driver is in Beta mode.**
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,9 @@
-# Go Databricks SQL Driver
+# Preview Go Databricks SQL Driver
 
 Databricks SQL driver for Go's [database/sql](https://golang.org/pkg/database/sql) package.
 
+**NOTE: This Driver is in Preview mode.**
```
arikfr, line 5: I'm not sure "Preview Mode" will be clear to people? Even if they are familiar with the [Databricks definitions](https://docs.databricks.com/release-notes/release-types.html) - we don't really map to any of the preview phases...

Maybe we should call it beta?

Something like: "Note: this driver is in beta."?

README.md
```diff
@@ -1,7 +1,9 @@
-# Go Databricks SQL Driver
+# Preview Go Databricks SQL Driver
```
arikfr, line 1: We should put the label at the end of the title.

moderakh: changed to beta and moved to the end.

README.md
moderakh: changed to beta.

arikfr (approved):

PR #18, opened 2022-07-25
Fix: FetchResults truncates rows returned by the server

susodapop:
## Description

Apache Hive thrift always sends `HasMoreResults==false` even when more results are available. This change updates the Next() method to continue fetching pages of results until an empty page is received.

This quirk of behaviour also affects the [Databricks NodeJS connector](https://github.com/databricks/databricks-sql-nodejs/issues/21) and we work around it in a similar way there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ## 0.1.4 (2022-07-30)
 
 - Fix: Could not fetch rowsets greater than the value of `maxRows` (#18)
+- Updated default user agent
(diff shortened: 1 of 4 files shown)


yunbodeng-db (approved):
Any tests?

susodapop:
I added an e2e test for this behaviour. Double-checked that it _failed_ prior to this fix and now passes. I also ran all the unit tests locally with `go test` and confirmed they pass.

yunbodeng-db (approved):

PR #22, opened 2022-09-10
[PECO-306] bug fix for escape characters breaking E2E examples

mattdeekay:
Since release v0.1.3, When running `examples/list/list.go`, the `SHOW TABLES IN @p1` queries were all failing with messages like these:
```
2022/09/09 19:16:14 error in querying database 000_demo_db: Error running query: [PARSE_SYNTAX_ERROR] org.apache.spark.sql.catalyst.parser.ParseException: 
[PARSE_SYNTAX_ERROR] Syntax error at or near ''000_demo_db''(line 1, pos 15)

== SQL ==
SHOW TABLES IN '000_demo_db'
---------------^^^
```

After this change, these examples work as it had before release v0.1.3:
```
2022/09/09 19:18:58 Tables in 000_demo_db:
2022/09/09 19:18:58   bi0_mbpartner
```

Signed-off-by: Matthew Kim <11141331+mattdeekay@users.noreply.github.com>

### DIFF
--- a/escaper.go
+++ b/escaper.go
@@ -20,7 +20,7 @@ func EscapeArg(arg driver.NamedValue) (string, error) {
 	case bool:
 		return fmt.Sprintf("%v", v), nil
 	case string:
-		return fmt.Sprintf("'%v'", strings.ReplaceAll(v, "'", "''")), nil
+		return fmt.Sprintf("%v", strings.ReplaceAll(v, "'", "''")), nil
 	case time.Time:
 		return "'" + v.Format(TimeFmt) + "'", nil
 	default:
--- a/statement.go
+++ b/statement.go
@@ -107,11 +107,10 @@ func statement(tmpl string, args []driver.NamedValue) (string, error) {
 		} else {
 			re = regexp.MustCompile(fmt.Sprintf("@p%d%s", arg.Ordinal, `\b`))
 		}
-		escaped, err := EscapeArg(arg)
+		val, err := EscapeArg(arg)
 		if err != nil {
 			return "", err
 		}
-		val := fmt.Sprintf("%v", escaped)
 		stmt = re.ReplaceAllString(stmt, val)
 	}
 	return stmt, nil
--- a/escaper_test.go
+++ b/escaper_test.go
@@ -11,7 +11,7 @@ func TestEscaper(t *testing.T) {
 		Value          driver.Value
 		ExpectedOutput string
 	}{
-		{Value: "a'b'c", ExpectedOutput: `'a''b''c'`},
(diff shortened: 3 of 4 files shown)


yunbodeng-db (commented):
What is the release v0.1.3? GoLang connector itself or one of the dependencies?
LGTM so long as this is not a reflection on the server change (if so, can we discuss how we would handle versioning due to server change?)

escaper.go
```diff
@@ -20,7 +20,7 @@ func EscapeArg(arg driver.NamedValue) (string, error) {
 	case bool:
 		return fmt.Sprintf("%v", v), nil
 	case string:
-		return fmt.Sprintf("'%v'", strings.ReplaceAll(v, "'", "''")), nil
```
moderakh, line 23: I see. so previously we were embracing string in a quotation mark. interesting find @mattdeekay

statement.go
```diff
@@ -107,11 +107,10 @@ func statement(tmpl string, args []driver.NamedValue) (string, error) {
 		} else {
 			re = regexp.MustCompile(fmt.Sprintf("@p%d%s", arg.Ordinal, `\b`))
 		}
-		escaped, err := EscapeArg(arg)
+		val, err := EscapeArg(arg)
```
moderakh, line 110: @mattdeekay I know invoking `EscapeArg` was not introduced in your PR.

So this is merely a "non-blocking" question (so I understand the context/impact of this change), in which scenario we escape ARG? is this for supporting SDK side query parametrization?

is `EscapeArg(arg)` invoked in any scenario other than query parametrization?

moderakh (approved):
Thanks @mattdeekay for fixing the broken behaviour. LGTM. 

I also have a question for you.
@mattdeekay I know invoking `EscapeArg` was not introduced in your PR.
So this is merely a "non-blocking" question (so I understand the context/impact of this change), in which scenario we escape ARG? is this for supporting SDK side query parametrization?

is `EscapeArg(arg)` invoked in any scenario other than query parametrization?

mattdeekay:
> What is the release v0.1.3? GoLang connector itself or one of the dependencies? LGTM so long as this is not a reflection on the server change (if so, can we discuss how we would handle versioning due to server change?)

This is the release of GoLang connector itself (not a server change)

statement.go
mattdeekay: I checked and saw that `EscapeArg(arg)` is only involved in query parametrization across the Go SDK

susodapop:
It looks like the `build-and-test` check is failing because of environment issues. I will update our Github `DATABRICKS_DSN` variable for this repository and we can retry the check.

PR #24, opened 2022-09-21
[PECO-271] Support cancel operation

mattdeekay:
### Description
Supports cancel operation (e.g. for long-running queries)

#### Testing with synchronous query execution
Modified `statement.go: 149 (func exec)` to Cancel instead of Close and tested via IDE
Screen Shot 2022-09-21 at 11 37 12 AM

#### Testing with asynchronous query execution
When manually overriding `session.ExecuteStatement` to RunAsync: true in code and running the above query, able to confirm that cancel operation work as intended and cancels the query execution.
Screen Shot 2022-09-21 at 1 11 27 PM

#### (Earlier Caveat - no longer applies)
`session.ExecuteStatement` (`s.hive.client.ExecuteStatement`) appears to work synchronously because the SDK will not return a response until the query is fully executed. I was unable to test cancellation while query was executing, only the cancellation before FetchResults

### DIFF
--- a/hive/operation.go
+++ b/hive/operation.go
@@ -119,8 +119,8 @@ func (op *Operation) Close(ctx context.Context) error {
 	return nil
 }
 
-// Cancel cancels operation
-func (op *Operation) Cancel(ctx context.Context) error {
+// private method to cancel operation
+func (op *Operation) cancel(ctx context.Context) error {
 	req := cli_service.TCancelOperationReq{
 		OperationHandle: op.h,
 	}
--- a/hive/operation.go
+++ b/hive/operation.go
@@ -118,3 +118,21 @@ func (op *Operation) Close(ctx context.Context) error {
 	op.hive.log.Printf("close operation: %v", guid(op.h.OperationId.GUID))
 	return nil
 }
+
+// Cancel cancels operation
+func (op *Operation) Cancel(ctx context.Context) error {
+	req := cli_service.TCancelOperationReq{
+		OperationHandle: op.h,
+	}
+
+	resp, err := op.hive.client.CancelOperation(ctx, &req)
+	if err != nil {
+		return err
+	}
+	if err := checkStatus(resp); err != nil {
+		return err
+	}
+
+	op.hive.log.Printf("cancel operation: %v", guid(op.h.OperationId.GUID))
+	return nil
+}


kravets-levko:
> appears to work synchronously because the SDK will not return a response until the query is fully executed. 

Have you tried `runAsync` option? In Nodejs I noticed the opposite behavior: cancelling works fine while query is executing, but sometimes fails if fetching already started 😅

hive/operation.go
```diff
@@ -118,3 +118,21 @@ func (op *Operation) Close(ctx context.Context) error {
 	op.hive.log.Printf("close operation: %v", guid(op.h.OperationId.GUID))
 	return nil
 }
+
+// Cancel cancels operation
+func (op *Operation) Cancel(ctx context.Context) error {
```
moderakh, line 123: we added a new cancel API here. Does this follow the go spec? https://pkg.go.dev/database/sql
is the method signature of this method defined in the go spec we follow? could you please share the link to the cancel signature from the go spec?

moderakh (commented):
@mattdeekay The PR descriptions says:

> When manually overriding session.ExecuteStatement to RunAsync: true in code and running the above query, able to confirm that cancel operation work as intended and cancels the query execution.

Could you please provide sample code for how you are overriding RunAsync in the ExecuteStatement?

mattdeekay:
> @mattdeekay The PR descriptions says:
> 
> > When manually overriding session.ExecuteStatement to RunAsync: true in code and running the above query, able to confirm that cancel operation work as intended and cancels the query execution.
> 
> Could you please provide sample code for how you are overriding RunAsync in the ExecuteStatement?

In `session.go`: 
```
func (s *Session) ExecuteStatement(ctx context.Context, stmt string) (*Operation, error) {
	req := cli_service.TExecuteStatementReq{
		SessionHandle: s.h,
		Statement:     stmt,
		RunAsync:  true
	}
	resp, err := s.hive.client.ExecuteStatement(ctx, &req)

```

moderakh:
> func (s *Session) ExecuteStatement(ctx context.Context, stmt string) (*Operation, error) {
> 	req := cli_service.TExecuteStatementReq{
> 		SessionHandle: s.h,
> 		Statement:     stmt,
> 		RunAsync:  true
> 	}



> > @mattdeekay The PR descriptions says:
> > > When manually overriding session.ExecuteStatement to RunAsync: true in code and running the above query, able to confirm that cancel operation work as intended and cancels the query execution.
> > 
> > 
> > Could you please provide sample code for how you are overriding RunAsync in the ExecuteStatement?
> 
> In `session.go`:
> 
> ```
> func (s *Session) ExecuteStatement(ctx context.Context, stmt string) (*Operation, error) {
> 	req := cli_service.TExecuteStatementReq{
> 		SessionHandle: s.h,
> 		Statement:     stmt,
> 		RunAsync:  true
> 	}
> 	resp, err := s.hive.client.ExecuteStatement(ctx, &req)
> ```

`session.go` is an implementation class. seems to me you are changing the implementation? is that right?

Think about it from customer/end-user point of view, can end user pass RunAsync=true from our GO SDK public API?

mattdeekay:
> > > @mattdeekay The PR descriptions says:
> > > > When manually overriding session.ExecuteStatement to RunAsync: true in code and running the above query, able to confirm that cancel operation work as intended and cancels the query execution.
> > > 
> > > 
> > > Could you please provide sample code for how you are overriding RunAsync in the ExecuteStatement?
> > 
> > 
> > In `session.go`:
> > ```
> > func (s *Session) ExecuteStatement(ctx context.Context, stmt string) (*Operation, error) {
> > 	req := cli_service.TExecuteStatementReq{
> > 		SessionHandle: s.h,
> > 		Statement:     stmt,
> > 		RunAsync:  true
> > 	}
> > 	resp, err := s.hive.client.ExecuteStatement(ctx, &req)
> > ```
> 
> `session.go` is an implementation class. seems to me you are changing the implementation? is that right?
> 
> Think about it from customer/end-user point of view, can end user pass RunAsync=true from our GO SDK public API?

Yep, I was changing the implementation but this was just to test that query execution would work async. Of course right now, end user cannot pass RunAsync=true but I already have a separate work item (PECO-341). When that work item is finished, the end result will be that `session.go > ExecuteStatement` will be passed RunAsync argument

mattdeekay:
> > > @mattdeekay The PR descriptions says:
> > > > When manually overriding session.ExecuteStatement to RunAsync: true in code and running the above query, able to confirm that cancel operation work as intended and cancels the query execution.
> > > 
> > > 
> > > Could you please provide sample code for how you are overriding RunAsync in the ExecuteStatement?
> > 
> > 
> > In `session.go`:
> > ```
> > func (s *Session) ExecuteStatement(ctx context.Context, stmt string) (*Operation, error) {
> > 	req := cli_service.TExecuteStatementReq{
> > 		SessionHandle: s.h,
> > 		Statement:     stmt,
> > 		RunAsync:  true
> > 	}
> > 	resp, err := s.hive.client.ExecuteStatement(ctx, &req)
> > ```
> 
> `session.go` is an implementation class. seems to me you are changing the implementation? is that right?
> 
> Think about it from customer/end-user point of view, can end user pass RunAsync=true from our GO SDK public API?

Yep, I was changing the implementation but this was just to test that query execution would work async. Of course right now, end user cannot pass RunAsync=true but I already have a separate work item (PECO-341). When that work item is finished, the end result will be that `session.go > ExecuteStatement` will be passed RunAsync argument

hive/operation.go
mattdeekay: As discussed offline, `func Cancel()` would introduce a new Cancel API that does not follow the go spec in database/sql. Instead, will make this a private implementation method and already have work items for Async Execution and Context Cancellation support that will utilize this private method.

moderakh: Thanks @mattdeekay please create a work item internally to track what we discussed for following database/sql spec for cancel.

moderakh (approved):
Thanks @mattdeekay please create a work item internally to track what we discussed for following database/sql spec for cancel.

other than that LGTM

PR #35, opened 2022-11-08
DSN Parsing, ExecContext, SessionParams

andrefurlan-db:
## Implemented DSN parsing

`token:supersecret2@example.cloud.databricks.com:8000/sql/1.0/endpoints/12346a5b5b0e123a?catalog=default&schema=system&timeout=100&maxRows=1000&timezone=America/Sao_Paulo`

where `catalog`, `schema`, `timeout`, `maxRows` are part of server protocol and anything else (such as timezone) are considered session parameters.

## Implemented ExecContext
For queries that return a number of rows affected instead of data.

## Implemented SessionParams 
Session params will be set using `SET param = value` at session opening

Also fixed a couple of bugs I found while testing and updated tests and examples


connection.go
```diff
@@ -187,52 +197,54 @@ func (c *conn) executeStatement(ctx context.Context, query string, args []driver
...
+func (c *conn) pollOperation(ctx context.Context, opHandle *ts.TOperationHandle) (*ts.TGetOperationStatusResp, error) {
+	var statusResp *ts.TGetOperationStatusResp
 	pollSentinel := sentinel.Sentinel{
 		OnDoneFn: func(statusResp any) (any, error) {
 			return statusResp, nil
 		},
 		StatusFn: func() (sentinel.Done, any, error) {
 			var err error
-			statusResp, err = c.client.GetOperationStatus(ctx, &cli_service.TGetOperationStatusReq{
+			statusResp, err = c.client.GetOperationStatus(context.Background(), &ts.TGetOperationStatusReq{
```
andrefurlan-db, line 219: this is a big weird bug that was fixed

rcypher-databricks (approved):

rcypher-databricks (approved):

PR #37, opened 2022-11-09
implementing statement

andrefurlan-db:
Basic implementation of statement.

There is no support in databricks for transactions or query parameters. So for now we are returning not supported errors.


connection.go
```diff
@@ -63,7 +76,14 @@ func (c *conn) IsValid() bool {
...
 
+// ExecContext executes a query that doesn't return rows, such
+// as an INSERT or UPDATE.
+//
+// ExecContext honors the context timeout and return when it is canceled.
 func (c *conn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	if len(args) > 0 {
+		return nil, fmt.Errorf("databricks: query parameters are not supported")
+	}
 	_, opStatusResp, err := c.runQuery(ctx, query, args)
```
mattdeekay, line 87: Is there a difference between `cli_service.TExecuteStatementResp.GetModifiedRowCount()` vs `cli_service.TGetOperationStatusResp.GetNumModifiedRows()`? I was using the former in my implementation

connection.go
```diff
@@ -118,26 +145,26 @@ func (c *conn) runQuery(ctx context.Context, query string, args []driver.NamedVa
...
 				return nil, statusResp, err
 			}
 			switch statusResp.GetOperationState() {
 			// terminal states
 			// good
-			case ts.TOperationState_FINISHED_STATE:
+			case cli_service.TOperationState_FINISHED_STATE:
 				// return handle to fetch results later
 				return exStmtResp, opStatus, nil
 			// bad
```
mattdeekay, line 170: Nit: a lot of nesting of cases here, no functional problems but I'll pull this out

connection.go
andrefurlan-db: haha, good question. Can you check and submit a PR if this needs to be changed?

connection.go
andrefurlan-db: yes. But it is still pretty readable, so I left the way it is to not complicate further. If it bothers you feel free to make the changes!

mattdeekay (approved):

rcypher-databricks (approved):

PR #41, opened 2022-11-10
Rewrite

andrefurlan-db:
This PR does a very large refactor (aka rewrite) of the driver do prepare for general availability.

## Design Considerations
* Any request to external systems must return as soon as the context deadline is exceeded
* Any request to external systems that do not take a context (open/close session) must have a timeout
* Packages that are not meant to be public to users must be in /internal
* No extension of the standard interface as context cancellation fulfills the requirements; thus, concepts such as “operation” or “cursor” are not needed.
* Contract testing instead of full integration testing (DUST). This enables all tests to run locally and in Github actions. And they are fast. This is done by creating a test thrift server using httptest package.
* Implement RoundTripper interface and wrap TCLIServiceClient to have access to HTTP response headers and status. This will enable retry.
* Fully declarative configuration with defaults instead of constants spreaded around

## New features
* Support for operation cancellation and timeout by using context.
* Direct Results. If the resultsset is small and the query takes less than 5 seconds, there will only be one request made, shortcutting 3 or 4 requests from the standard thrift protocol.
* Support for session parameters (maxRows, timeout, catalog, schema, and any other session params in SET such as timezone, ansi_mode, etc).
* Fully configurable logging

## Regressions
* Parameterized queries are not supported anymore. Client side string interpolation is prone to errors and Databricks will soon support it from the server side anyway.
* DSN does not start with "databricks" anymore. So one can just do `username:password@host:port/path/to/db?param=foo`

## Testing
There are 3 levels of testing:
* Unit: see internal/config/config_test.go
* Contract: see rows_test.go
* End to End with test server: see driver_e2e_test.go

## Bug fixes
* Many with regards to how to query, as synchronous queries were never really supported in the server.
* Some data type supported conversions where fixed

## Examples

```go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"log"
	"os"
	"time"

	_ "github.com/databricks/databricks-sql-go"
)

func main() {
	// Opening a driver typically will not attempt to connect to the database.
	db, err := sql.Open("databricks", <DSN>)

	if err != nil {
		// This will not be a connection error, but a DSN parse error or
		// another initialization error.
		log.Fatal(err)
	}
	defer db.Close()

	ctx, cancel1 := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel1()
	var res int
	err := db.QueryRowContext(ctx, `SELECT id FROM RANGE(100000000) ORDER BY RANDOM() + 2 asc`).Scan(&res)
	if err != nil {
		fmt.Printf("err: %v\n", err1)
	} else {
		fmt.Printf("result: %d\n", res)
	}
}

```

```go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"log"
	"os"
	"strconv"

	dbsql "github.com/databricks/databricks-sql-go"
)

func main() {
	port, err := strconv.Atoi(os.Getenv("DATABRICKS_PORT"))
	if err != nil {
		log.Fatal(err.Error())
	}
	connector, err := dbsql.NewConnector(
		dbsql.WithServerHostname(os.Getenv("DATABRICKS_HOST")),
		dbsql.WithPort(port),
		dbsql.WithHTTPPath(os.Getenv("DATABRICKS_HTTPPATH")),
		dbsql.WithAccessToken(os.Getenv("DATABRICKS_ACCESSTOKEN")),
		dbsql.WithSessionParams(map[string]string{"timezone": "America/Sao_Paulo", "ansi_mode": "true"}),
	)
	if err != nil {
		// This will not be a connection error, but a DSN parse error or
		// another initialization error.
		log.Fatal(err)
	}
	db := sql.OpenDB(connector)
	defer db.Close()
	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()

	res := &time.Time{}
	if qerr := db.QueryRowContext(ctx, `select now()`).Scan(res); qerr != nil {
		fmt.Printf("err: %v\n", qerr)
	} else {
		fmt.Println(res)
	}

}

```

## TODO
* Testing is still in a minimum, even though we added a lot more tests than previously
* Retry. The interfaces are all there, we just need to write the logic to retry now.


mattdeekay (approved):
LGTM

PR #53, opened 2022-11-16
End2end

andrefurlan-db:
- Added some good end 2 end tests with way to read server responses from json files. One can now just create an example, set the `recordResults` flag in client, and all server responses will be saved to JSON.
- Fix a few bugs regarding maxRows (can't be zero) and timezone.
- Added DSN parsing of userAgentEntry 
- Adds some docs
- Improved a few logs
- Added a very comprehensive example under example/workflow


internal/config/config.go
```diff
@@ -46,60 +49,76 @@ func (c *Config) DeepCopy() *Config {
...
+		QueryTimeout:   ucfg.QueryTimeout,
+		UserAgentEntry: ucfg.UserAgentEntry,
+		Location:       loccp,
+		SessionParams:  sessionParams,
 	}
 }
 
-func (ucfg UserConfig) FillDefaults() UserConfig {
+func (ucfg UserConfig) WithDefaults() UserConfig {
 	if ucfg.MaxRows == 0 {
```
mattdeekay, line 122: Could maybe do `<= 0` in case someone tries to have a negative MaxRows

testserver.go
```diff
@@ -21,31 +18,14 @@ func (h *thriftHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		protocolFactory = thrift.NewTBinaryProtocolFactoryConf(tcfg)
-	case "header":
-		protocolFactory = thrift.NewTHeaderProtocolFactoryConf(tcfg)
-	default:
-		panic(fmt.Errorf("invalid protocol specified %s", cfg.ThriftProtocol))
-	}
-	if cfg.ThriftDebugClientProtocol {
-		protocolFactory = thrift.NewTDebugProtocolFactoryWithLogger(protocolFactory, "client:", thrift.StdLogger(nil))
-	}
+	protocolFactory := thrift.NewTBinaryProtocolFactoryConf(tcfg)
```
mattdeekay, line 28: Curious - will we need the other protocols?

mattdeekay (approved):

testserver.go
andrefurlan-db: not for the test server, I guess :-)

internal/config/config.go
andrefurlan-db: ah, good one

PR #56, opened 2022-11-22
Async

andrefurlan-db:
experimentation on one approach of supporting true async apis


db.go
```diff
@@ -0,0 +1,149 @@
...
+func newContextWithExecution(ctx context.Context, exc *Execution) context.Context {
+	return context.WithValue(ctx, driverctx.ExecutionContextKey, exc)
+}
+
+func excFromContext(ctx context.Context) *Execution {
+	excId, ok := ctx.Value(driverctx.ExecutionContextKey).(*Execution)
+	if !ok {
+		return nil
+	}
+	return excId
```
mattdeekay, line 148: Nit: Since it's the `Execution` pointer and not Id being returned, perhaps use `exc` instead of `excId`

andrefurlan-db:
one idea to improve would be `rs, exc, err = db.CheckExecution(ogCtx, exc)`. This would make a single call and make the loop simpler

internal/config/config.go
```diff
@@ -20,7 +20,7 @@ type Config struct {
 	TLSConfig     *tls.Config // nil disables TLS
 	Authenticator string      //TODO for oauth
 
-	RunAsync                  bool // TODO
+	RunAsync                  bool
```
aldld, line 23: Is there a particular reason why this is set at the config level, rather than on a per-query level? Or rather, is there any reason not to provide different methods for callers to use depending on whether they want the sync vs. async behaviour?

examples/asyncWorkflow/main.go
```diff
@@ -0,0 +1,151 @@
...
+	db.SetMaxOpenConns(2)
+
+	// the "github.com/databricks/databricks-sql-go/driverctx" has some functions to help set the context for the driver
+	ogCtx := dbsqlctx.NewContextWithCorrelationId(context.Background(), "asyncWorkflow-example")
+
+	// for _, v := range []string{"1", "2", "3", "4", "5", "6", "7", "8", "9", "10"} {
+	// i := v
+	// go func() {
+	// _, exc, err := db.QueryContext(ogCtx, fmt.Sprintf("select %s", i))
+	rs, exc, err := db.QueryContext(ogCtx, `SELECT id FROM RANGE(100) ORDER BY RANDOM() + 2 asc`)
```
aldld, line 69: To make sure I'm understanding the semantics of this correctly:

1. If the query has completed (direct results), then `rs` will be set and the caller can fetch the results
2. Otherwise, the caller can use `exc` to get the operation ID(?), and use that to poll for query status?

aldld (commented):
Took a quick look over, mainly focused on the interface. Looking forward to this!

internal/config/config.go
andrefurlan-db: no. It is the fastest way I got the PoC to work :-) . I have to think about the methods in the interface, maybe it does make sense to do as you say

examples/asyncWorkflow/main.go
andrefurlan-db: hi @aldld , that's the idea... there is more to the execution than the ID, but it is just data that can be saved and used later.
I am not very happy with the way to check the results yet with different methods to query, check, get results. There must be a simpler way to merge check and get results in the same call. See the example I added so you have an idea of what I am talking about.

PR #57, opened 2022-11-23
feat: enable username and password auth

andrefurlan-db:
In case there are instances of Databricks where username and password are supported, we should also support it.


rcypher-databricks (approved):

yunbodeng-db (commented):
Maybe it has been the case but should we support username/passwd? We should support PAT only like other open source SDKs.

andrefurlan-db:
@yunbodeng-db, the case in know is community edition. There's even an open PR for it here. I don't think it hurts.

PR #60, opened 2022-11-30
P&R improvements

andrefurlan-db:
Added a retryable http client. Now we also re-use clients across connections. Not need to create one client per connection.

Fixed a problem where queries would wait intervalTime before firing
Fixed a problem where on context done the query would not be cancelled.


rcypher-databricks (approved):

connection.go
```diff
@@ -250,41 +261,56 @@ func logBadQueryState(log *logger.DBSQLLogger, opStatus *cli_service.TGetOperati
...
+		isOpen := resp.DirectResults != nil && resp.DirectResults.CloseOperation == nil
+		return hasHandle && isOpen
 	}
 
-	exStmtResp, ok := res.(*cli_service.TExecuteStatementResp)
-	if !ok {
-		return exStmtResp, errors.New("databricks: invalid execute statement response")
+	select {
+	default:
+	case <-ctx.Done():
```
mattdeekay, line 293: Thanks for explaining this logic to me offline. Would be nice to revisit sentinel in the future to perhaps have it work for non-polling operations.

mattdeekay (approved):
Discussed offline - retryableHttp from hashicorp is another dependency we'll be adding but I think that since it's a thin wrapper around the standard http library while providing very useful retry functionality without much security concern, I'm good with this approach.

connection.go
andrefurlan-db: yeap. But maybe work for post GA.

PR #61, opened 2022-12-01
Don't ignore error in InitThriftClient

aldld:
`thrift.NewTHttpClientWithOptions` may return an error, in which case `tTrans` is `nil`, which causes the next line to panic with the message `interface conversion: thrift.TTransport is nil, not *thrift.THttpClient` on the next line.

Signed-off-by: Eric Bannatyne <eric@sigmacomputing.com>

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -212,6 +212,10 @@ func InitThriftClient(cfg *config.Config) (*ThriftServiceClient, error) {
 			Timeout:   cfg.ClientTimeout,
 		}
 		tTrans, err = thrift.NewTHttpClientWithOptions(endpoint, thrift.THttpClientOptions{Client: httpclient})
+		if err != nil {
+			return nil, err
+		}
+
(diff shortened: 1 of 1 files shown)


internal/client/client.go
```diff
@@ -212,6 +212,10 @@ func InitThriftClient(cfg *config.Config) (*ThriftServiceClient, error) {
 			Timeout:   cfg.ClientTimeout,
 		}
 		tTrans, err = thrift.NewTHttpClientWithOptions(endpoint, thrift.THttpClientOptions{Client: httpclient})
+		if err != nil {
+			return nil, err
```
aldld, line 216: I can replace this with a `break` if that's what's preferred here, since it looks like we check for errors immediately after this `switch` block.

andrefurlan-db (approved):
Thanks!

internal/client/client.go
andrefurlan-db: it is fine like this

PR #64, opened 2022-12-02
Close optimization for Rows

rcypher-databricks:
When a query returns direct results it is possible that all the query result rows are contained therein.  In this case the server can automatically close the operation.  This is indicated by a non-null CloseOperation member in the TSparkDirectResults structure. In this case calls to Rows.Close don't require any interaction with the server as the operation is already closed.
- added a NewRows function and updated connection to use it when creating Rows
- added logic in NewRows to check for an already closed operation
- updated Rows.Close so we don't make a server round trip when not necessary
- added a unit test for the close optimization logic


mattdeekay (approved):
LGTM

PR #65, opened 2022-12-02
Close operation after executing statement

rcypher-databricks:
When executing a query there is a rows object returned to the client and the client is then responsible for closing the row set when finished with it.  
Closing the rows causes the driver to close the operation on the server.

When executing a statement (ex. DROP TABLE ...) the client doesn't receive anything by which they can close the operation.  This results in an operation that is completed but not closed.  Eventually the open operation is closed due to inactivity but the driver should be closing the operation.

- updated connection.ExecContext to close the operation if 1) we receive an operation handle back from the server and 2) the operations status is not already closed.
- added unit test to check that we do/don't close the operation depending on the returned operation status.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -102,13 +102,14 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 		// we have an operation id so update the logger
 		log = logger.WithContext(c.id, driverctx.CorrelationIdFromContext(ctx), client.SprintGuid(exStmtResp.OperationHandle.OperationId.GUID))
 
-		// since we have an operation handle we can close the operation
-		if opStatusResp == nil || opStatusResp.GetOperationState() != cli_service.TOperationState_CLOSED_STATE {
+		// since we have an operation handle we can close the operation if necessary
+		alreadyClosed := exStmtResp.DirectResults != nil && exStmtResp.DirectResults.CloseOperation != nil
+		if !alreadyClosed && (opStatusResp == nil || opStatusResp.GetOperationState() != cli_service.TOperationState_CLOSED_STATE) {
 			_, err1 := c.client.CloseOperation(ctx, &cli_service.TCloseOperationReq{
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -90,21 +90,37 @@ func (c *conn) IsValid() bool {
...
 		return nil, errors.New(ErrParametersNotSupported)
 	}
 	exStmtResp, opStatusResp, err := c.runQuery(ctx, query, args)
 
 	if exStmtResp != nil && exStmtResp.OperationHandle != nil {
+		// we have an operation id so update the logger
 		log = logger.WithContext(c.id, driverctx.CorrelationIdFromContext(ctx), client.SprintGuid(exStmtResp.OperationHandle.OperationId.GUID))
+
+		// since we have an operation handle we can close the operation
+		if opStatusResp == nil || opStatusResp.GetOperationState() != cli_service.TOperationState_CLOSED_STATE {
```
andrefurlan-db, line 106: is this enough? I think we also have to check `directResult.CloseOperationResp`, no?

connection.go
```diff
@@ -90,21 +90,37 @@ func (c *conn) IsValid() bool {
...
+			_, err1 := c.client.CloseOperation(ctx, &cli_service.TCloseOperationReq{
+				OperationHandle: exStmtResp.OperationHandle,
+			})
+			if err1 != nil {
+				log.Err(err1).Msg("failed to close operation after executing statement")
```
andrefurlan-db, line 111: add a `databricks:` to the message to maintain consistency?

connection.go
rcypher-databricks: You're right. I tried it and the returned operation state can be 'Finished' (i.e. not closed) while the direct results have a TCloseOperationResp with a status of 'Success', indicating that the operation is in a closed state.

andrefurlan-db (approved):
👍

PR #71, opened 2022-12-08
Handle parsing negative years in dates

rcypher-databricks:
time.Time can represent dates with a negative year, however the time.ParseInLocation function does not handle parsing date strings with a negative year.
- Added our own parseInLocation function which handles parsing with negative years.
- Added unit tests for date/time parsing.


andrefurlan-db (approved):
nice

PR #74, opened 2022-12-10
fix thread safety issue

carl-db:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/databricks/databricks-sql-go/pull/74).
* __->__ #74

fix thread safety issue

This fixes a thread safety (and structural issue) with connector. Because connectors have a long lifetime (potentially process-lifetime), and are used by multiple threads it's not safe to share the same thrift client connection.

This changes the connector so that a new thrift client is created on each connection. The go DB library will perform pooling of connections, so we don't need to worry about it here.

Signed-off-by: Carl Verge <carl.verge@databricks.com>

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -15,8 +15,7 @@ import (
 )
 
 type connector struct {
-	cfg    *config.Config
-	client cli_service.TCLIService
+	cfg *config.Config
 }
 
 func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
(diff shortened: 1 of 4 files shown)


andrefurlan-db:
Hi @carl-db, thanks for the PR. Can you commit with -s to pass the DCO check?

connector.go
```diff
@@ -29,8 +28,13 @@ func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
 		schemaName = cli_service.TIdentifierPtr(cli_service.TIdentifier(c.cfg.Schema))
 	}
 
+	tclient, err := client.InitThriftClient(c.cfg)
```
andrefurlan-db, line 31: This is okay to fix the issue, but ideally I'd like to reuse the http client (not necessarily the thrift client). The http client has a connection pool and it is intended to be used across many go routines.

andrefurlan-db (approved):

PR #63, opened 2022-12-01
Integrate with Go SDK

nfx:
This PR integrates SQL connector with Databricks SDK for Go. Transitively, it enables the following authentication methods:

- Azure Active Directory Service Principal Auth with well-known environment variables
- Azure CLI Auth
- Google Cloud Platform Default Application Credentials with well-known environment variables
- Google Cloud Platform Credentials with well-known environment variables
- Databricks Personal Access Tokens with well-known environment variables
- Databricks Basic Authentication with well-known environment variables
- `~/.databrickscfg` integration with well-known environment variables

## To be decided:

- [ ] changing `WithUserAgentEntry` with `databricks.WIthProduct`
- [ ] field name for `sdkConfig`
- [ ] need to allow unified user agent string: `isv-product-name/0.0.0 databricks-sdk-go/0.0.1 go/1.19.3 os/darwin sql-connector/0.9.0`, because currently it fails with malformed request
- [ ] if we want to add configurability by `WarehouseName`/`ClusterName`.
- [ ] logging interface, as SQL connector uses non-standard-library `zerolog`, which is great library, but might be too heavy for an SDK.

Signed-off-by: Serge Smertin <serge.smertin@databricks.com>

PR #81, opened 2022-12-16
[feat] Adding Custom Authenticator and Http Conn Pool

andrefurlan-db:
This PR adds:
* the ability to pass in an Authenticator to the driver. If a token is passed, the PAT authenticator will be used
* reusability of HTTP connections, including increasing MaxIdleConnsPerHost to 10.
* better error message, including Databricks specific error headers.
* fixed a few bugs in the way, namely on the retryable client and failure to close operation when context is canceled.


auth/auth.go
```diff
@@ -0,0 +1,7 @@
+package auth
+
+import "net/http"
+
+type Authenticator interface {
```
moderakh, line 5: is this a public user facing API?

andrefurlan-db: It is a public facing interface. Only needed if user wants to have their own implementation.

connection.go
```diff
@@ -102,7 +102,7 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 		// since we have an operation handle we can close the operation if necessary
 		alreadyClosed := exStmtResp.DirectResults != nil && exStmtResp.DirectResults.CloseOperation != nil
 		if !alreadyClosed && (opStatusResp == nil || opStatusResp.GetOperationState() != cli_service.TOperationState_CLOSED_STATE) {
-			_, err1 := c.client.CloseOperation(ctx, &cli_service.TCloseOperationReq{
+			_, err1 := c.client.CloseOperation(context.Background(), &cli_service.TCloseOperationReq{
```
rcypher-databricks, line 105: nice catch

rcypher-databricks (approved):

andrefurlan-db:
This is a v1.1.0 release PR

auth/auth.go
moderakh: @andrefurlan-db I am planning to write the decision doc we discussed sometime end of this week / early next week. as this is a change in the public surface area, can we wait to close on that decision before merging this?

andrefurlan-db: This PR is target for version 1.1.0. It won't be merged this year.

auth/pat/pat.go
```diff
@@ -0,0 +1,20 @@
+package pat
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+type PATAuth struct {
```
nfx, line 10: why do you need to re-implement auth, if there's a concurrent [PR](https://github.com/databricks/databricks-sql-go/pull/63) on integrating with Go SDK, that solves the auth problem in a unified way across different tools?

nfx (changes requested):

moderakh (changes requested):
pending a one decision doc on the plan as a follow up internal discussion.

Issue #68, opened 2022-12-05
[feat] Report typed errors for easier integration


andrefurlan-db:

andrefurlan-db:
@aldld FYI

matterpale:
Yes 👍  @andrefurlan-db 
For starters, the three error messages that until recently resided on root in `err.go` file (i.e. not implemented; unsupported tx and query params) - I see that just yesterday they were hidden under `internal` 😞 
I would appreciate the opposite approach: turn them straight into sentinel errors and keep them exported for users to check against.

matterpale:
I am not suggesting sentinel errors are the ultimate way to go (+1 for types), it just seems a more sensible and easily achievable alternative to these error messages, at least until these features get supported.

prochac:
Would be nice to implement something similar like other go sql drivers do:

https://pkg.go.dev/github.com/lib/pq#Error
https://pkg.go.dev/github.com/go-sql-driver/mysql#MySQLError
https://pkg.go.dev/github.com/jackc/pgconn#PgError
https://pkg.go.dev/github.com/denisenkom/go-mssqldb#Error
https://pkg.go.dev/github.com/snowflakedb/gosnowflake#SnowflakeError
https://pkg.go.dev/github.com/vertica/vertica-sql-go#VError.SQLState

PR #93, opened 2023-01-11
Error types

mattdeekay:
### Issue
https://github.com/databricks/databricks-sql-go/issues/68

### Plan
The following DBSQL error categories will be exposed to clients:
**RequestError**
**ExecutionError**
**SystemFault**

The standard Databricks error expose fields `err`, `corrId`, `connId`. ExecutionError will also expose `sqlState`, `queryId`, and `errClass`. SystemFault will also expose `isRetryable`. 

### Remaining work
* compare errors with JDBC/ODBC
* investigate returned headers to provide granularity around errors (particularly SystemFault)
* warehouses vs clusters have slightly different errors

### DIFF
--- a/error/err.go
+++ b/error/err.go
@@ -1,16 +1,18 @@
 package error
 
 import (
+	"context"
 	"fmt"
+	"github.com/databricks/databricks-sql-go/driverctx"
 	"github.com/pkg/errors"
 )
 
 type DatabricksError struct {
-	msg          string
-	err          error
-	corrId       string
-	connId       string
-	queryId      string
+	msg    string
+	err    error
+	corrId string
+	connId string
+	//queryId      string
 	errCondition string
 	errType      DatabricksErrorType
 }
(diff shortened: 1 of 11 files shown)


error/err.go
```diff
@@ -0,0 +1,13 @@
+package error
+
+import "fmt"
+
+// ConnectionError is an error in the connection to the server
+type ConnectionError struct {
+	Msg string
```
rcypher-databricks, line 7: Have we decided on the types of errors to expose?
I don't think the struct members should be public

error/err.go
```diff
@@ -0,0 +1,13 @@
...
+	Err error
+}
+
+func (e *ConnectionError) Error() string {
```
rcypher-databricks, line 11: I think we need to implement Unwrap to work correctly with errors.Is/As

internal/config/config.go
```diff
@@ -169,14 +169,14 @@ func ParseDSN(dsn string) (UserConfig, error) {
 	}
 	parsedURL, err := url.Parse(fullDSN)
 	if err != nil {
-		return UserConfig{}, errors.Wrap(err, "invalid DSN: invalid format")
+		return UserConfig{}, &dbsqlerror.ConnectionError{Msg: "invalid DSN: invalid format", Err: err}
```
rcypher-databricks, line 172: Do we still want to add a stack trace?

PR #94, opened 2023-01-13
Update DriverVersion in Config to 1.0.1

mattdeekay:
## Description
In this PR, we update the driver version to 1.0.1 to reflect the latest cut. We also move the DriverVersion to a more visible const in `driver.go`. 

(Already merged) Most of the code in `driver.go:Open` is the same as `driver.go:OpenConnector` so these are merged.

### DIFF
--- a/connector_test.go
+++ b/connector_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var ExpectedDriverVersion = "1.0.1"
+
 func TestNewConnector(t *testing.T) {
 	t.Run("Connector initialized with functional options should have all options set", func(t *testing.T) {
 		host := "databricks-host"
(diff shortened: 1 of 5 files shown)


internal/config/config.go
```diff
@@ -152,7 +152,7 @@ func WithDefaults() *Config {
 		PingTimeout:               15 * time.Second,
 		CanUseMultipleCatalogs:    true,
 		DriverName:                "godatabrickssqlconnector", // important. Do not change
-		DriverVersion:             "0.9.0",
+		DriverVersion:             "1.0.1",
```
andrefurlan-db, line 155: maybe the best way is to have a version.go file at top level folder instead of the config? It is pretty hidden in here :-)

mattdeekay: How about a `constants.go` file? Then we can have DriverVersion, DriverName as constants in the top level folder and refer to them here?

connector_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var ExpectedDriverVersion = "1.0.1"
```
andrefurlan-db, line 13: is this always equal driver.DriverVersion?

andrefurlan-db (approved):
I think we can merge this. Let's see how it goes, but I feel like having a version package would make it easier to find and also use in any other package

PR #98, opened 2023-01-24
Escape and subtitute query arguments client side

nyergler:
Coming from other Go database drivers the lack of query argument support was a surprise.  This commit ports approximately what the Python Databricks library does in order to handle query arguments.

Feedback, suggestions, etc welcome!


nyergler:
Superseded by #100; apologies for the noise, just wanted to get changes onto separate branches so we can continue using our fork.

PR #99, opened 2023-01-24
Use the context logger, if available.

nyergler:
This allows use of a non-global logger for finer grained control.


andrefurlan-db:
Hi @nyergler, great call. Thanks for this. I'll test it out a bit. Maybe we just need to add some example of how to use it?

nyergler:
Superseded by #180.

PR #100, opened 2023-01-24
Escape and subtitute query arguments client side

nyergler:
Coming from other Go database drivers the lack of query argument support was a surprise. This commit ports approximately what the Python Databricks library does in order to handle query arguments.

Feedback, suggestions, etc welcome!

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -98,7 +98,11 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 
 	ctx = driverctx.NewContextWithConnId(ctx, c.id)
 	if len(args) > 0 {
-		return nil, errors.New(ErrParametersNotSupported)
+		q, err := SubstituteArgs(query, args)
+		if err != nil {
+			return nil, err
+		}
+		query = q
 	}
 	exStmtResp, opStatusResp, err := c.runQuery(ctx, query, args)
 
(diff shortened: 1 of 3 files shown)


andrefurlan-db (changes requested):
we explicitly did not add support for this. The feature is under development in spark and databricks. We'll add support for the driver once the backend is ready.

nyergler:
Got it, @andrefurlan-db . Is there an expected timeline for that development? I'm sure there's a lot in flight, just curious how far along that is.

nohajc:
I did a bit of digging and it seems the parametrized query feature has been released in Spark 3.4.0 (merged since December 2022). However, it is not exposed in the thrift server which is used by databricks-sql-go to connect to Spark.
https://issues.apache.org/jira/browse/SPARK-41271
https://github.com/apache/spark/pull/38864
https://github.com/apache/spark/blob/master/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/SparkExecuteStatementOperation.scala#L226

To me it looks like the feature is ready but not exposed to external clients.

kravets-levko:
Hi @nohajc! Th library got a support for named query parameters since v1.5.0. Can you please try it and let us know if it works for yoy? Thanks!

PR #117, opened 2023-03-24
Implementing error types

rcypher-databricks:
Updated driver error handling to standardize on three error types: driver error (DBDriverError), request error (DBRequestError), and sql execution error (DBExecutionError). All three types have accessors for stack trace, correlation id, and connection id.  DBExecutionError also has accessors for query Id and sql state.
The example in examples/error shows how to use errors.Is() and errors.As() to work with the error chain to determine error type and extract the state.


achettyiitr:
Any updates on this?

connection.go
```diff
@@ -217,16 +218,16 @@ func (c *conn) runQuery(ctx context.Context, query string, args []driver.NamedVa
 				cli_service.TOperationState_ERROR_STATE,
 				cli_service.TOperationState_TIMEDOUT_STATE:
 				logBadQueryState(log, statusResp)
-				return exStmtResp, statusResp, errors.New(statusResp.GetDisplayMessage())
+				return exStmtResp, statusResp, dbsqlerr.NewSystemFault(ctx, dbsqlerr.ErrInvalidOperationState, nil)
```
andrefurlan-db, line 221: request error?

andrefurlan-db (commented):
we have to be careful to not wrap context deadline exceeded with our errors

andrefurlan-db (commented):
SystemFault has a different connotation than any driver error (such as not implemented). We need to either change this name to something like DriverError or really try to separate Databricks's system faults from the rest.

doc.go
```diff
@@ -144,6 +144,52 @@ The result log may look like this:
 
 	{"level":"debug","connId":"01ed6545-5669-1ec7-8c7e-6d8a1ea0ab16","corrId":"workflow-example","queryId":"01ed6545-57cc-188a-bfc5-d9c0eaf8e189","time":1668558402,"message":"Run Main elapsed time: 1.298712292s"}
 
+# Errors
+
+There are three error types exposed via dbsql/errors
+
+	DBSystemFault - A fault caused by Databricks services
+
+	DBRequestError - An error that is caused by an invalid request. Example: permission denied, or the user tries to access a warehouse that doesn’t exist
```
andrefurlan-db, line 153: too many requests?
network error?

internal/config/config.go
```diff
@@ -195,21 +197,21 @@ func ParseDSN(dsn string) (UserConfig, error) {
 	}
 	parsedURL, err := url.Parse(fullDSN)
 	if err != nil {
-		return UserConfig{}, errors.Wrap(err, "invalid DSN: invalid format")
+		return UserConfig{}, dbsqlerr.NewRequestError(context.TODO(), dbsqlerr.ErrInvalidDSNFormat, err)
```
andrefurlan-db, line 200: does this makes sense? there is no request yet...but maybe any driver interaction is a "request"?

internal/config/config.go
```diff
@@ -195,21 +197,21 @@ func ParseDSN(dsn string) (UserConfig, error) {
...
 	}
 	ucfg := UserConfig{}.WithDefaults()
 	ucfg.Protocol = parsedURL.Scheme
 	ucfg.Host = parsedURL.Hostname()
 	port, err := strconv.Atoi(parsedURL.Port())
 	if err != nil {
-		return UserConfig{}, errors.Wrap(err, "invalid DSN: invalid DSN port")
+		return UserConfig{}, dbsqlerr.NewRequestError(context.TODO(), dbsqlerr.ErrInvalidDSNPort, err)
```
andrefurlan-db, line 207: should we have a context for this function then?

examples/error/main.go
```diff
@@ -0,0 +1,508 @@
...
+	"strconv"
+	"time"
+
+	dbsql "github.com/databricks/databricks-sql-go"
+	dbsqlctx "github.com/databricks/databricks-sql-go/driverctx"
+	dbsqlerr "github.com/databricks/databricks-sql-go/errors"
+	"github.com/joho/godotenv"
+)
+
+// NOTE: these tests print errors to the console and are for demonstration purposes, not unit test coverage
```
andrefurlan-db, line 19: we need to make this file into a real example of how to use the error interfaces we are exposing. I am assuming that this file was not meant to be merged as is.

internal/config/config.go
rcypher-databricks: ParseDSN is called from databricksDriver.OpenConnector(dsn string) which doesn't take a context.  We can't change OpenConnector() because it is part of the driver.DriverContext interface.

andrefurlan-db (approved):
I think this is good to go now

PR #119, opened 2023-04-10
Implemented IsRetryable() and RetryAfter() for DatabricksError

rcypher-databricks:
Added IsRetryable and RetryAfter functions to DBError interface.
Added an internal error type for retryable errors.
Updated client to insert a retryable error instance into the error chain.


andrefurlan-db (commented):
change retry after to time.duration, and add an example of how to use it

andrefurlan-db (approved):
looks good!

PR #120, opened 2023-04-19
Initial cloud fetch work

rcypher-databricks:
Updated arrow based code to handle arrow batches returned in the rowset from the server or a set of URLs to files in the cloud containing the batches.


databricks
mattdeekay, line 1: Could this file be removed? Not sure what it's for, perhaps the IDE tucked this in?

PR #125, opened 2023-05-17
Updated retry behaviour

rcypher-databricks:
- retry on 500 status codes if the server operation is idempotent
- retry on request errors except: too many redirects, invalid protocol scheme,  TLS cert verification failure

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -56,16 +56,7 @@ const (
 	cancelOperation
 )
 
-var idempotentClientMethods map[clientMethod]any = map[clientMethod]any{
-	closeSession:         struct{}{},
-	getResultSetMetadata: struct{}{},
-	getOperationStatus:   struct{}{},
-	closeOperation:       struct{}{},
-	cancelOperation:      struct{}{},
-	// fetchResults is treated as idempotent because the drivers fetching logic is robust enoug
-	// to fetch results in correct order
-	fetchResults: struct{}{},
-}
+var nonRetryableClientMethods map[clientMethod]any = map[clientMethod]any{executeStatement: struct{}{}}
 
 // OpenSession is a wrapper around the thrift operation OpenSession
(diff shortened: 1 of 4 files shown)


internal/client/client.go
```diff
@@ -34,9 +37,40 @@ type ThriftServiceClient struct {
...
+	closeSession
+	fetchResults
+	getResultSetMetadata
+	executeStatement
+	getOperationStatus
+	closeOperation
+	cancelOperation
+)
+
+var idempotentClientMethods map[clientMethod]any = map[clientMethod]any{
```
andrefurlan-db, line 59: can we do the opposite? Look at the ones that are not idempotent

andrefurlan-db: only not retry on executeStatement

internal/client/client.go
```diff
@@ -431,59 +447,121 @@ func (l *leveledLogger) Warn(msg string, keysAndValues ...interface{}) {
...
 		// TODO @mattdeekay: convert these to specific error types
 		if resp.Header != nil {
 			orgid = resp.Header.Get("X-Databricks-Org-Id")
 			reason = resp.Header.Get("X-Databricks-Reason-Phrase") // TODO note: shown on notebook
 			terrmsg = resp.Header.Get("X-Thriftserver-Error-Message")
 			errmsg = resp.Header.Get("x-databricks-error-or-redirect-message")
-			retryAfter = resp.Header.Get("Retry-After")
 			// TODO note: need to see if there's other headers
 		}
 		msg := fmt.Sprintf("orgId: %s, reason: %s, thriftErr: %s, err: %s", orgid, reason, terrmsg, errmsg)
```
andrefurlan-db, line 467: this error message is too cumbersome. As an user it would read strange. Better to keep the same logic as before.

andrefurlan-db (approved):

PR #131, opened 2023-06-05
Tweak retry behaviour

rcypher-databricks:
In a retryable error situation should not retry if the client method is not considered retryable.


andrefurlan-db (approved):
looks good. The generated code will cause questions from other people looking at this code, later we should add comments.

PR #136, opened 2023-06-23
Bug fix for query that doesn't produce rows

rcypher-databricks:
When executing a statement that doesn't produce rows (ex. CREATE TABLE ....) using DB.Query() instead of DB.Exec() the driver would panic. When DB.Query() is used we try to create a row object.  In this case there was a mismatch in the schema metadata returned by thrift and the arrow schema metadata. The thrift schema was indicating a single column named 'Result' while the arrow schema metadata was indicating no columns. Moved the code creating the arrow schema and column info into separate functions. Updated creation of the column info to handle mismatch between the arrow and thrift schemas.  


susodapop (approved):
LGTM

mattdeekay (approved):
Looks good!

PR #143, opened 2023-07-11
Improve error information when query terminates in unexpected state.

rcypher-databricks:
When a query terminated in either an unknown or an unexpected operation state the connection was creating a driver error with a static message. 
Updated to include the operation state in the error message for an unknown operation state situation. 
Updated the case of an unexpected operation state to include the state and the operation status display message. 
Updated unit tests based on the new error message.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -200,7 +200,7 @@ func (c *conn) runQuery(ctx context.Context, query string, args []driver.NamedVa
 			cli_service.TOperationState_ERROR_STATE,
 			cli_service.TOperationState_TIMEDOUT_STATE:
 			logBadQueryState(log, opStatus)
-			return exStmtResp, opStatus, errors.New(opStatus.GetDisplayMessage())
+			return exStmtResp, opStatus, unexpectedOperationState(opStatus)
 		// live states
 		case cli_service.TOperationState_INITIALIZED_STATE,
 			cli_service.TOperationState_PENDING_STATE,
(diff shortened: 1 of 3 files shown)


connection.go
```diff
@@ -248,18 +248,27 @@ func (c *conn) runQuery(ctx context.Context, query string, args []driver.NamedVa
...
 
 func logBadQueryState(log *logger.DBSQLLogger, opStatus *cli_service.TGetOperationStatusResp) {
 	log.Error().Msgf("databricks: query state: %s", opStatus.GetOperationState())
-	log.Error().Msg(opStatus.GetErrorMessage())
+	log.Error().Msg(opStatus.GetDisplayMessage())
+	log.Debug().Msg(opStatus.GetDiagnosticInfo())
+}
+
+func unexpectedOperationState(opStatus *cli_service.TGetOperationStatusResp) error {
+	return errors.WithMessage(errors.New(opStatus.GetDisplayMessage()), dbsqlerr.ErrUnexpectedOperationState(opStatus.GetOperationState().String()))
```
mattdeekay, line 267: Should we wrap this error in `dbsqlerrint.NewDriverError` or is that overkill?

rcypher-databricks: It isn't actually a driver error and it will be wrapped with an execution error back up the call chain so I don't think we should

mattdeekay (approved):

PR #150, opened 2023-07-25
Cloud Fetch tests

mattdeekay:
Unit and end-to-end tests:

- batchloader_test.go (unit)
  - test cloud fetch happy case where the server returns arrow batch bytes
  - test cloud fetch sad case where the server status code is 500
- fetcher_test.go (unit)
  - test concurrent fetcher with mock fetchable items to check if results are expected and whether the fetcher took the estimated amount of time
- cloudfetch_test.go (e2e)
  - test end to end functionality of cloud fetch by comparing actual results from execution with and without cloud fetch

### DIFF
--- a/internal/fetcher/fetcher_test.go
+++ b/internal/fetcher/fetcher_test.go
@@ -3,12 +3,15 @@ package fetcher
 import (
 	"context"
 	"github.com/databricks/databricks-sql-go/internal/config"
+	"math"
 	"testing"
+	"time"
 )
 
 // Create a mock struct for FetchableItems
 type mockFetchableItem struct {
 	item int
+	wait time.Duration
 	err  error
 }
 
@@ -18,6 +21,7 @@ type mockOutput struct {
 
 // Implement the Fetch method
 func (m *mockFetchableItem) Fetch(ctx context.Context, cfg *config.Config) ([]*mockOutput, error) {
+	time.Sleep(m.wait)
 	outputs := make([]*mockOutput, 5)
 	for i := range outputs {
 		sampleOutput := mockOutput{item: m.item}
@@ -34,7 +38,7 @@ func TestConcurrentFetcher(t *testing.T) {
 		cfg := &config.Config{}
 		inputChan := make(chan FetchableItems[*mockOutput], 10)
 		for i := 0; i < 10; i++ {
-			item := mockFetchableItem{item: i}
+			item := mockFetchableItem{item: i, wait: 1 * time.Second}
 			inputChan <- &item
 		}
 		close(inputChan)
@@ -45,6 +49,7 @@ func TestConcurrentFetcher(t *testing.T) {
 			t.Fatalf("Error creating fetcher: %v", err)
 		}
 
+		start := time.Now()
 		outChan, _, err := fetcher.Start()
 		if err != nil {
 			t.Fatalf("Error starting fetcher: %v", err)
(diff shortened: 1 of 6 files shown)


internal/rows/arrowbased/batchloader_test.go
```diff
@@ -0,0 +1,89 @@
...
+			cu := &cloudURL{
+				TSparkArrowResultLink: &cli_service.TSparkArrowResultLink{
+					FileLink: tc.server.URL,
+				},
+			}
+
+			ctx := context.Background()
+			cfg := config.WithDefaults()
+
+			resp, err := cu.Fetch(ctx, cfg)
```
mattdeekay, line 79: At this point, I think I am properly connecting to test server defined here but `resp` is always `nil` and `err` is "EOF"

internal/fetcher/fetcher_test.go
```diff
@@ -0,0 +1,69 @@
+package fetcher
+
+import (
+	"context"
+	"github.com/databricks/databricks-sql-go/internal/config"
+	"testing"
+)
+
+// Create a mock struct for FetchableItems
+type mockFetchableItem struct {
```
rcypher-databricks, line 10: For a more realistic test we should add a wait time to mockFetchableItem and have the Fetch method sleep for that time so that we get the order of items finishing to be non-deterministic.

internal/rows/arrowbased/batchloader_test.go
```diff
@@ -0,0 +1,89 @@
...
+	schema := arrow.NewSchema(fields, nil)
+
+	builder := array.NewRecordBuilder(mem, schema)
+	defer builder.Release()
+
+	builder.Field(0).(*array.Int32Builder).AppendValues([]int32{1, 2, 3}, nil)
+	builder.Field(1).(*array.StringBuilder).AppendValues([]string{"one", "two", "three"}, nil)
+
+	record := builder.NewRecord()
+
```
rcypher-databricks, line 36: add `defer record.Release()` here

internal/rows/arrowbased/batchloader_test.go
rcypher-databricks: When I try to run this I'm getting `runtime error: invalid memory address or nil pointer dereference`
Adding the line in `generateMockArrowBytes` seems to fix this and the test runs successfully.  Of course the test itself fails because there's no `expectedResponse` in the test case

internal/rows/arrowbased/batchloader_test.go
```diff
@@ -0,0 +1,89 @@
...
+func TestBatchLoader(t *testing.T) {
+	testTable := []struct {
+		name             string
+		server           *httptest.Server
+		expectedResponse []*sparkArrowBatch
+		expectedErr      error
+	}{
+		{
+			name: "cloud-fetch-happy-case",
+			server: httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
rcypher-databricks, line 57: If you have more than one test case isn't this going to be a problem? Since each test case is starting a server instance.

internal/rows/arrowbased/batchloader_test.go
mattdeekay: Could I ask what you added in `generateMockArrowBytes` to fix that?

internal/fetcher/fetcher_test.go
mattdeekay: Added as part of [this](https://github.com/databricks/databricks-sql-go/pull/150/commits/33fe9b15ee221575a4218b374f7753519612a7ad) commit

internal/rows/arrowbased/batchloader_test.go
mattdeekay: Changed this in the latest commit - httptest server is now created before the test cases are run

rcypher-databricks (approved):

PR #151, opened 2023-07-27
Expose useCloudFetch and maxDownloadThreads

mattdeekay:
Expose cloud fetch with max download threads options via DSN string and connector functional options.


connector.go
```diff
@@ -214,3 +214,17 @@ func WithAuthenticator(authr auth.Authenticator) connOption {
 		c.Authenticator = authr
 	}
 }
+
+// WithUseCloudFetch sets up the use of cloud fetch for query execution. Default is false.
+func WithUseCloudFetch(useCloudFetch bool) connOption {
```
rcypher-databricks, line 219: Let's name this just WithCloudFetch

errors/errors.go
```diff
@@ -28,6 +28,8 @@ const (
 	ErrBasicAuthNotSupported  = "invalid DSN: basic auth not enabled"
 	ErrInvalidDSNMaxRows      = "invalid DSN: maxRows param is not an integer"
 	ErrInvalidDSNTimeout      = "invalid DSN: timeout param is not an integer"
+	ErrInvalidUseCloudFetch   = "invalid format: useCloudFetch param is not a boolean"
```
rcypher-databricks, line 31: Let's add a helper function for invalid param values error messages.  That way they'll all be consistent.

mattdeekay: We talked offline, implementing a helper function directly accessed when throwing the error

rcypher-databricks (approved):

Issue #140, opened 2023-07-03
Connect: Expose option to set protocol


MovieStoreGuy:
## Context

The method `WithServerHostname` will configure the protocol used to `http` if the value provided is `localhost`, however, it can be possible to try connector a compliant databricks service via a local proxy that does the TLS termination.

## Diagram

service --> reverse proxy --> service

## Outcomes

I am open to either outcome of:
- New option that can parse a complete URL and apply the `WithProtocol`, `WithHostname`, `WithPort`, `WithHTTPPath` configuration values
- New option that allows forcing http connection via a new option `WithHTTPProtocol`

rcypher-databricks:
Just to verify: you are not using a dsn to connect, you are using NewConnector().

MovieStoreGuy:
Yes, sorry that is correct.

rcypher-databricks:
Version 1.4.0 has been released and allows for including http or https protocol when using WithServerHostname
ex. WithServerHostname("http:my.server.name")

rcypher-databricks:
Unless there are still problems with this we will close this issue.

PR #152, opened 2023-08-02
Better handling of bad connection errors and specifying server protocol.

rcypher-databricks:
Added a new error type that identifies as driver.ErrBadConn.  This can be added in to an error chain to signal that a connection should no longer be used and to retry with a new connection.

Updated ThriftServiceClient error handling. When a Thrift request fails we are now checking the error message for the term 'Invalid SessionHandle'.  If it is present we are adding a bad connection error to the error chain.  Searching in the error message is clumsy and fragile but there doesn't currently appear to be another way to get the information.

Updated the WithServerHostname function to handle host names prefixed by 'http:' or 'https:' to allow users to specify which protocol to use.  This is in response to github issue #140.


internal/client/client.go
```diff
@@ -283,6 +286,42 @@ func InitThriftClient(cfg *config.Config, httpclient *http.Client) (*ThriftServi
...
+// handler function for errors returned by the thrift client methods
+func handleClientMethodError(ctx context.Context, err error) dbsqlerr.DBRequestError {
+	if err == nil {
+		return nil
+	}
+
+	// If the passed error indicates an invalid session we inject a bad connection error
+	// into the error stack. This allows the for retrying with a new connection.
+	s := err.Error()
+	if strings.Contains(s, "Invalid SessionHandle") {
```
andrefurlan-db, line 298: we need to be very careful with session state. If the user expects some state - temp views, etc - and we just open a new session in the background, it may be very hard to understand and debug what is going on.

aldld: In the context of `database/sql`, if the user is using methods directly on the `sql.DB` object directly (e.g. `db.QueryContext(...)`) then they cannot rely on maintaining per-session state anyway because these methods might create new connections or reuse an arbitrary connection from the pool.

If a user does need to rely on per-session state, then doing so correctly requires explicitly grabbing a connection and then calling methods on the connection. In that case, it's the user's responsibility to manage the connection lifecycle, which includes handling errors and accounting for the possibility that a connection may end up in a bad state.

internal/client/client.go
```diff
@@ -283,6 +286,42 @@ func InitThriftClient(cfg *config.Config, httpclient *http.Client) (*ThriftServi
...
+		err = dbsqlerrint.NewBadConnectionError(err)
```
mattdeekay, line 299: Out of curiosity, looks like the NewBadConnectionError will be wrapped again in a NewRequestError. Would it be better to just return the error as is? Or is the client expecting a NewRequestError?

andrefurlan-db (approved):

PR #155, opened 2023-08-16
Named parameter support and update Thrift

rcypher-databricks:
Updated thrift client code to latest version.
Added support for named query parameters.
Not yet complete.  The correct sql type needs to be set in TSparkParameter struct.  Need to add support for date/time and intervals.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -101,6 +101,10 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 	defer log.Duration(msg, start)
 
 	ctx = driverctx.NewContextWithConnId(ctx, c.id)
+	if len(args) > 0 {
+		return nil, dbsqlerrint.NewDriverError(ctx, dbsqlerr.ErrParametersNotSupported, nil)
+	}
+
 	exStmtResp, opStatusResp, err := c.runQuery(ctx, query, args)
 
 	if exStmtResp != nil && exStmtResp.OperationHandle != nil {
@@ -140,6 +144,9 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 	msg, start := log.Track("QueryContext")
 
 	ctx = driverctx.NewContextWithConnId(ctx, c.id)
+	if len(args) > 0 {
+		return nil, dbsqlerrint.NewDriverError(ctx, dbsqlerr.ErrParametersNotSupported, nil)
+	}
 
 	// first we try to get the results synchronously.
 	// at any point in time that the context is done we must cancel and return
--- a/connection.go
+++ b/connection.go
@@ -3,6 +3,7 @@ package dbsql
 import (
 	"context"
 	"database/sql/driver"
+	"fmt"
 	"time"
 
 	"github.com/databricks/databricks-sql-go/driverctx"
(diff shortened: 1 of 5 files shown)


yunbodeng-db (commented):
To be consistent with odbc/jdbc (which actually will not have named param support), do we have positional param support on the roadmap?
I understand the server support is work in progress but i think the client side support should be very similar between two. We can introduce a switch to turn positional param on when the server is done.

Another question is: do we check server versions? I should find out the versions that named param and positional params are supported. Someone on BIX-internal must know.

mattdeekay:
> To be consistent with odbc/jdbc (which actually will not have named param support), do we have positional param support on the roadmap? I understand the server support is work in progress but i think the client side support should be very similar between two. We can introduce a switch to turn positional param on when the server is done.
> 
> Another question is: do we check server versions? I should find out the versions that named param and positional params are supported. Someone on BIX-internal must know.

We do have client-side positional param support with this change. The Go SQL spec allows users to add positional OR named parameters like this:
`c.ExecContext(ctx, query, args ...any)`

In order for customers to use named parameters, they might do this:
`c.ExecContext(ctx, "SELECT :param_a, :param_b FROM sample", NamedArg{Name: "param_a", Value: "test_value_a"}, NamedArg{Name: "param_b", Value: "test_value_b"})`

But the driver will receive `[]driver.NamedValue` for both positional and named. It's up to the driver to send the correct `TSparkParameter` to the thrift server.

connection.go
```diff
@@ -337,6 +334,83 @@ func (c *conn) executeStatement(ctx context.Context, query string, args []driver
 	return resp, err
 }
 
+func namedValuesToTSparkParams(args []driver.NamedValue) []*cli_service.TSparkParameter {
+	var ts []string = []string{"STRING", "DOUBLE", "BOOLEAN"}
```
mattdeekay, line 338: Should all number values be "DOUBLE"? What about INT, or DECIMAL?

internal/config/config.go
```diff
@@ -194,7 +194,7 @@ func WithDefaults() *Config {
 		DriverName:                "godatabrickssqlconnector", // important. Do not change
 		ThriftProtocol:            "binary",
 		ThriftTransport:           "http",
-		ThriftProtocolVersion:     cli_service.TProtocolVersion_SPARK_CLI_SERVICE_PROTOCOL_V6,
+		ThriftProtocolVersion:     cli_service.TProtocolVersion_SPARK_CLI_SERVICE_PROTOCOL_V7,
```
mattdeekay, line 197: I think I was able to get NamedParameters working against e2-dogfood SQL without updating the ThriftProtocolVersion. Wondering if we know what changing this would achieve?

connection.go
rcypher-databricks: At the moment TSparkParameterValue only support bool, double and string values.  I'm waiting for clarification on the question of losing precision when converted to double.  Also waiting for clarification on whether TSparkParameter.Type should correspond to the type in TSparkParameterValue or if it is used for converting the value.

mattdeekay: From the code I saw in DBR, `TSparkParameter.Type` should correspond to the data types listed [here](https://docs.databricks.com/en/sql/language-manual/sql-ref-datatypes.html). In DBR, there will be an attempt to cast these to the correct type. Floats can even be represented as a string so we could do this (pseudocode):
```
TSparkParameter{Type: "FLOAT", Value: TSparkParameter{stringValue: "1234.125"}, ...}
```
Or 
```
TSparkParameter{Type: "DECIMAL(6, 2)", Value: TSparkParameter{stringValue: "1234.12"}, ...}
```
Or
```
TSparkParameter{Type: "BOOLEAN", Value: TSparkParameter{stringValue: "false"}, ...}
```
I think "ARRAY", "MAP", "STRUCT", and "BINARY" are not supported here however

connection.go
```diff
@@ -100,9 +101,6 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 	defer log.Duration(msg, start)
 
 	ctx = driverctx.NewContextWithConnId(ctx, c.id)
-	if len(args) > 0 {
```
andrefurlan-db, line 103: maybe for this PR we avoid any user facing changes?

andrefurlan-db (approved):
looks good. Building blocks :-)

PR #158, opened 2023-08-29
[PECO-1016-2] Add handling for special types

nithinkdb:
In this PR, we add further handling to allow for special types to be set via the DBSQLParams variable.

### DIFF
--- a/parameters.go
+++ b/parameters.go
@@ -35,8 +35,6 @@ const (
 
 func (s SqlType) String() string {
 	switch s {
-	case Void:
-		return "VOID"
 	case String:
 		return "STRING"
 	case Date:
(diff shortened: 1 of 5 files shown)


parameters.go
```diff
@@ -0,0 +1,149 @@
...
+
+const (
+	String SqlType = iota
+	Date
+	Timestamp
+	Float
+	Decimal
+	Double
+	Integer
+	Bigint
```
rcypher-databricks, line 28: can we change these to BigInt, SmallInt, and TinyInt

rcypher-databricks (commented):
Let's add a new example showing how to run a parameterized query.
We should also add a test driver_e2e_test.go so that we have a test that exercises the interaction of our connection as a NamedValueChecker.

nithinkdb:
> Let's add a new example showing how to run a parameterized query. We should also add a test driver_e2e_test.go so that we have a test that exercises the interaction of our connection as a NamedValueChecker.

Created this follow up work item to track both of these: https://databricks.atlassian.net/browse/PECO-1048

rcypher-databricks (approved):
Looks good.  I think we might want to change the names of the SqlType constants but we can talk about that for the next PR.

PR #163, opened 2023-09-20
PECO-1054 Expose Arrow batches to users, part two

rcypher-databricks:
Updated FetchableItems interface to return an instance of OutputType instead of a slice of output type. 
Created interfaces SparkArrowBatch and SparkArrowRecord and implementations of each. This also changed the 1:1 ratio of batch instances to arrow records.  A SparkArrowBatch can contain multiple arrow records now. 
Created BatchIterator interface and implementation and switched arrowRowScanner to use BatchIterator instead of BatchLoader 
Created RowValues interface and implementation as a container for the currently loaded values for a set of rows. 
Updated the behaviour of fetchable items cloudURL and localBatch to de-serialize the arrow records as part of fetching, rather than carry around the raw bytes for later de-serialization. Also eliminated the cloud fetch code that was de-serializing the arrow batch then serializing each record individually to create one batch instance per record. 
Removed chunkedByteReader and replaced with io.MultiReader 
Normalized use of row number so that there is no need to track the index of the row in the current batch.


internal/rows/errors.go
```diff
@@ -7,6 +7,8 @@ var errRowsNilRows = "databricks: nil Rows instance"
 var errRowsUnknowRowType = "databricks: unknown rows representation"
 var errRowsCloseFailed = "databricks: Rows instance Close operation failed"
 var errRowsMetadataFetchFailed = "databricks: Rows instance failed to retrieve result set metadata"
+var errRowsOnlyForward = "databricks: Rows instance can only iterate forward over rows"
```
jadewang-db, line 10: general question, should we give more info about the error, eg, what is the row number which has error?

internal/rows/arrowbased/arrowRows.go
```diff
@@ -688,3 +670,54 @@ func (vcm *arrowValueContainerMaker) makeColumnValueContainer(t arrow.DataType,
...
+	arrowRecords []SparkArrowRecord
+}
+
+var _ SparkArrowBatch = (*sparkArrowBatch)(nil)
+
+func (b *sparkArrowBatch) Next() (SparkArrowRecord, error) {
+	if len(b.arrowRecords) > 0 {
+		r := b.arrowRecords[0]
+		// remove the record from the slice as iteration is only forwards
+		b.arrowRecords = b.arrowRecords[1:]
```
jadewang-db, line 687: just curious, how efficient is this [1:] operation in go?

jadewang-db (approved):
added some comments, but don't block on me.

andrefurlan-db (approved):

internal/rows/arrowbased/arrowRows.go
rcypher-databricks: Very fast the new slice is just pointing in the same memory as the original slice.

PR #164, opened 2023-09-20
[PECO-1015] Add support for staging operations to Go Driver

nithinkdb:
We need to add support for staging operations to the go driver. This PR enabled Get, Delete, and Put operations.


internal/config/config.go
```diff
@@ -84,24 +84,25 @@ func (c *Config) DeepCopy() *Config {
...
-	QueryTimeout      time.Duration // Timeout passed to server for query processing
-	UserAgentEntry    string
-	Location          *time.Location
-	SessionParams     map[string]string
-	RetryWaitMin      time.Duration
-	RetryWaitMax      time.Duration
-	RetryMax          int
-	Transport         http.RoundTripper
-	UseLz4Compression bool
+	Protocol                string
```
jadewang-db, line 87: what is the change here?

connection.go
```diff
@@ -129,6 +143,139 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
 
+func Succeeded(response *http.Response) bool {
+	if response.StatusCode == 200 || response.StatusCode == 201 || response.StatusCode == 202 || response.StatusCode == 204 {
+		return true
+	}
+	return false
+}
+
+func (c *conn) HandleStagingPut(presignedUrl string, headers map[string]string, localFile string) (driver.Result, error) {
+	if localFile == "" {
```
jadewang-db, line 154: where if localFile being used in this function?

connection.go
```diff
@@ -129,6 +143,139 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
+
+}
+
+func (c *conn) HandleStagingDelete(presignedUrl string, headers map[string]string) (driver.Result, error) {
+	client := &http.Client{}
+	req, _ := http.NewRequest("DELETE", presignedUrl, nil)
+	for k, v := range headers {
+		req.Header.Set(k, v)
+	}
+	res, err := client.Do(req)
```
jadewang-db, line 212: are we doing any retries for all the client.do calls?

connection.go
```diff
@@ -129,6 +143,139 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
+		return nil, err
+	}
+	defer res.Body.Close()
+	content, err := io.ReadAll(res.Body)
+
+	if err != nil || !Succeeded(res) {
+		return nil, fmt.Errorf("staging operation over HTTP was unsuccessful: %d-%s", res.StatusCode, content)
+	}
+
+	err = os.WriteFile(localFile, content, 0644)
```
rcypher-databricks, line 198: should add an annotation to turn off this warning

connector.go
```diff
@@ -25,6 +25,38 @@ type connector struct {
 	client *http.Client
 }
 
+type StagingCtx struct {
```
rcypher-databricks, line 28: I don't think it makes sense to define our own struct that implements context.Context.  There's already a mechanism for passing request-scoped data that transits processes and API boundaries.
Define a key type and values and use context.WithValue.

connector.go
```diff
@@ -25,6 +25,38 @@ type connector struct {
...
+	IsStagingOperation      bool
```
rcypher-databricks, line 29: What happens if the query is a GET/PUT/DELETE and the user doesn't set this to true?

internal/config/config.go
```diff
@@ -84,24 +84,25 @@ func (c *Config) DeepCopy() *Config {
...
+	QueryTimeout            time.Duration // Timeout passed to server for query processing
+	UserAgentEntry          string
+	Location                *time.Location
+	SessionParams           map[string]string
+	RetryWaitMin            time.Duration
+	RetryWaitMax            time.Duration
+	RetryMax                int
+	Transport               http.RoundTripper
+	UseLz4Compression       bool
+	StagingAllowedLocalPath string
```
rcypher-databricks, line 105: Where do we use StagingAllowedLocalPath?  Also in StagingCtx this member is a slice of string while here it is a single string.

internal/config/config.go
```diff
@@ -282,6 +284,11 @@ func ParseDSN(dsn string) (UserConfig, error) {
 		ucfg.Location, err = time.LoadLocation(timezone)
 	}
 
+	if params.Has("stagingAllowedLocalPath") {
```
rcypher-databricks, line 287: I think it would be cleaner to do:
```
if paths, ok := params.extract("stagingAllowedLocalPath"); ok {
    ucfg.StagingAllowedLocalPath = catalog
}
```

internal/config/config.go
```diff
@@ -282,6 +284,11 @@ func ParseDSN(dsn string) (UserConfig, error) {
...
+		ucfg.StagingAllowedLocalPath = params.Get("stagingAllowedLocalPath")
+	}
+	params.Del("stagingAllowedLocalPath")
```
rcypher-databricks, line 290: If we parse this out of the DSN we should also allow the user to set it as a connector option.

internal/config/config.go
nithinkdb: Undone

connection.go
nithinkdb: nope, that seems like overkill.

nithinkdb: I will leave it up to Raymond, if that's a necessary feature we can implement it, but i don't believe we do these retries in go or python

internal/config/config.go
nithinkdb: Now we've just moved it to the context.

rcypher-databricks (commented):
Looks good.  Just need to fix the build and linting errors.

connection.go
susodapop: Note that in Python we do retry these requests automatically. Presumably that happens automatically in Node as well because we use `axios`. So Jade's feedback makes sense here. Does the Go http library ever automatically retry?

connection.go
```diff
@@ -129,6 +163,151 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
+	stagingAllowedLocalPaths := driverctx.StagingPathsFromContext(ctx)
+	for i := range stagingAllowedLocalPaths {
+		path := stagingAllowedLocalPaths[i]
+		relativePath, err := filepath.Rel(path, localFile)
+		if err != nil {
+			return false
+		}
+		if !strings.Contains(relativePath, "../") {
+			return true
+		}
```
susodapop, line 262: Do you have unit or e2e tests that prove this is secure?

I'm thinking something along these lines: https://github.com/databricks/databricks-sql-python/blob/fcc262f9bd4f48432e599b4f4f8b9cf62f8ace73/tests/e2e/common/staging_ingestion_tests.py#L213-L228

Rather than manually checking for `../` I'd recommend using a method like Python's `os.path.commonpath` which verifies absolutely that the allowed local path isn't being gamed. It does this by actually resolving the `../` and comparing the output absolute path to the allowed local paths.

From what I'm seeing [here](https://stackoverflow.com/questions/28024731/check-if-given-path-is-a-subdirectory-of-another-in-golang) I think a little more is required beyond just using go's `Rel` method.

susodapop (changes requested):
Overall looks good but requires tests to verify the behaviour of allowed local paths.

connection.go
nithinkdb: I can add in that unit test. Unfortunately, I'm not sure what this analog would be. The article you linked says that we just need to look for the ../ right?

nithinkdb:
> Are these for staging operations or UC Volumes?

This is for UC Volumes

connection.go
nithinkdb: Outside of retriable libraries, no. https://github.com/hashicorp/go-retryablehttp -> we could hypothetically use this?

nithinkdb: At the very least, this seems like follow up PR material given that it is a little more involved. This PR has been sitting a bit, and doesn't really break anything. I think we should prioritize getting it in, unless we see customers complain about retry issues.

rcypher-databricks (approved):

PR #166, opened 2023-09-22
PECO-1054 Expose Arrow batches to users, part three

rcypher-databricks:
Added DBSqlRows and DBSQLArrowBatchIterator public interfaces. 
Added arrowRecordIterator which implements  DBSQLArrowBatchIterator. 
Moved closing the database operation from rows type into resultPageIterator as well as properties that are only used by resultPageIterator. 
Added GetArrowBatches function to rows and arrowRowScanner types. 
Added HasNext function to BatchIterator and SparkArrowBatch interfaces. 
Added example for accessing Arrow batches and updated doc.go


doc.go
```diff
@@ -233,6 +233,80 @@ Example usage:
...
+
+	func main() {
+		...
+		db := sql.OpenDB(connector)
+		defer db.Close()
+
+		conn, _ := db.Conn(context.BackGround())
+		defer conn.Close()
+
+		query := `select * from hive_metastore.main.taxi_trip_data`
```
andrefurlan-db, line 277: change to UC. `main.default.<>`

examples/arrrowbatches/main.go
```diff
@@ -0,0 +1,139 @@
...
+	})
+	if err != nil {
+		log.Fatalf("unable to run the query. err: %v", err)
+	}
+	defer rows.Close()
+
+	ctx2, cancel2 := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel2()
+
+	batches, err := rows.(dbsqlrows.DBSQLRows).GetArrowBatches(ctx2)
```
andrefurlan-db, line 120: is `DBSQL` something that we actually expose externally as an acronym? Any alternatives to this?

andrefurlan-db (commented):
just a couple of annoying comments to make sure we name things that we expose as interface correctly. The rest looks awesome!

andrefurlan-db (approved):
good job

PR #168, opened 2023-09-27
[PECO-1048] Add example for parameterized queries

nithinkdb:
We don't currently have an example for parameterized queries, we should add one.

### DIFF
--- a/examples/parameters/main.go
+++ b/examples/parameters/main.go
@@ -50,11 +50,11 @@ func main() {
 	:p_double AS col_double,
 	:p_float AS col_float,
 	:p_date AS col_date`,
-		dbsql.DBSqlParam{Name: "p_bool", Value: true},
-		dbsql.DBSqlParam{Name: "p_int", Value: int(1234)},
-		dbsql.DBSqlParam{Name: "p_double", Type: dbsql.SqlDouble, Value: "3.14"},
-		dbsql.DBSqlParam{Name: "p_float", Type: dbsql.SqlFloat, Value: "3.14"},
-		dbsql.DBSqlParam{Name: "p_date", Type: dbsql.SqlDate, Value: "2017-07-23 00:00:00"}).Scan(&p_bool, &p_int, &p_double, &p_float, &p_date)
+		dbsql.Parameter{Name: "p_bool", Value: true},
+		dbsql.Parameter{Name: "p_int", Value: int(1234)},
+		dbsql.Parameter{Name: "p_double", Type: dbsql.SqlDouble, Value: "3.14"},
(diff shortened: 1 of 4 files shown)


examples/parameters/main.go
```diff
@@ -0,0 +1,63 @@
...
+	var res string
+	err1 := db.QueryRowContext(ctx, `SELECT
+	:p_bool AS col_bool,
+	:p_int AS col_int,
+	:p_double AS col_double,
+	:p_floatfloat AS col_float`,
+		dbsql.DBSqlParam{Name: "p_bool", Value: true},
+		dbsql.DBSqlParam{Name: "p_int", Value: int(1234)},
+		dbsql.DBSqlParam{Name: "p_double", Type: dbsql.Double, Value: 3.14},
+		dbsql.DBSqlParam{Name: "p_float", Type: dbsql.Float, Value: 3.14}).Scan(&res)
```
rcypher-databricks, line 51: I think we need to have a param that requires the use of DBSqlParam.Type.  Maybe a date where the passed value is a string. And have a comment about how if Type is not specified we try to infer it.

rcypher-databricks (commented):
As part of this PR please add a section in doc.go explaining the feature and how to use it.
Including things like what happens when passing raw values or sql.NamedValue as parameters.

nithinkdb:
> As part of this PR please add a section in doc.go explaining the feature and how to use it. Including things like what happens when passing raw values or sql.NamedValue as parameters.

Adding in a future PR.

examples/parameters/main.go
```diff
@@ -0,0 +1,68 @@
...
+	var p_double float64
+	var p_float float32
+	var p_date string
+	err1 := db.QueryRowContext(ctx, `SELECT
+	:p_bool AS col_bool,
+	:p_int AS col_int,
+	:p_double AS col_double,
+	:p_float AS col_float,
+	:p_date AS col_date`,
+		dbsql.DBSqlParam{Name: "p_bool", Value: true},
```
andrefurlan-db, line 53: nothing to do with this PR, but why is this called DBSqlParam? Please, please standardize the names we expose as public interfaces. @rcypher-databricks what do you think?

andrefurlan-db: we should just call this Param. As it is already scoped to dbsql

rcypher-databricks: @nithinkdb I though we talked about dropping DBSql and just calling it Parameter.

nithinkdb: Sorry, forgot about this. Will do.

rcypher-databricks (approved):

PR #172, opened 2023-10-10
Logging improvements

rcypher-databricks:
Differentiated some cases where the logged messages were ambiguous as to which code path they came from. 
Updated client to be more consistent in what context information is included in logging messages. Also updated client methods to produce a duration log message even if there is an error.  Updated error handling to log the wrapped error message.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -390,7 +390,7 @@ func SprintGuid(bts []byte) string {
 	return fmt.Sprintf("%x", bts)
 }
 
-// Create an updated context and a logger that include query and connection id
+// Create an updated context and a logger that includes query and connection id
 func LoggerAndContext(ctx context.Context, c any) (*logger.DBSQLLogger, context.Context) {
 	connId := driverctx.ConnIdFromContext(ctx)
 	corrId := driverctx.CorrelationIdFromContext(ctx)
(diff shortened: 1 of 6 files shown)


internal/client/client.go
```diff
@@ -149,32 +161,36 @@ func (tsc *ThriftServiceClient) GetResultSetMetadata(ctx context.Context, req *c
...
+	log, ctx = LoggerAndContext(ctx, req)
+	msg, start := log.Track("ExecuteStatement")
 
 	// We use context.Background to fix a problem where on context done the query would not be cancelled.
 	resp, err := tsc.TCLIServiceClient.ExecuteStatement(context.Background(), req)
+	log, ctx = LoggerAndContext(ctx, resp)
+	logDisplayMessage(resp, log)
+	logExecStatementState(resp, log)
+
+	defer log.Duration(msg, start)
```
jadewang-db, line 175: should we move this to before line 170?

jadewang-db (approved):

internal/client/client.go
```diff
@@ -355,6 +390,95 @@ func SprintGuid(bts []byte) string {
 	return fmt.Sprintf("%x", bts)
 }
 
+// Create an updated context and a logger that include query and connection id
```
susodapop, line 393: typo

```suggestion
// Create an updated context and a logger that includes query and connection id
```

susodapop (approved):
LGTM

PR #174, opened 2023-10-16
[PECO-1179] Added handling for staging remove

nithinkdb:
Earlier when I had coded this, I accidentally introduced a bug where the user facing API was delete instead of remove. This was because under the hood we are doing a delete operation. This should fix the bug, we should definitely hotfix our release. I've tested it locally for each operation this time.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -493,7 +493,7 @@ func (c *conn) handleStagingGet(ctx context.Context, presignedUrl string, header
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


rcypher-databricks (approved):
We should definitely be testing all operations before releasing.

yunbodeng-db (commented):
I think we should have unit test or e2e test to reflect the change?
whenever I see a product code change without a test fix, it makes me wonder the test gaps and test coverage.

PR #175, opened 2023-10-17
[PECO-1178] Added .us domain to inference list for AWS

nithinkdb:
Currently, our list of supported aws domains only includes .com domains, we need to add a domain corresponding to .us.


yunbodeng-db (approved):
Thanks for doing it and keep the list in the alphabetic order.
Remember to sign your commit DCO.

jadewang-db (approved):

PR #180, opened 2023-11-30
Use the context logger, if available.

nyergler:
This allows use of a non-global logger for finer grained control.

### DIFF
--- a/internal/rows/rows.go
+++ b/internal/rows/rows.go
@@ -68,6 +68,7 @@ var _ driver.RowsColumnTypeLength = (*rows)(nil)
 var _ dbsqlrows.Rows = (*rows)(nil)
 
 func NewRows(
+	ctx context.Context,
 	connId string,
 	correlationId string,
 	opHandle *cli_service.TOperationHandle,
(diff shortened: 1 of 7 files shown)


internal/rows/rows.go
```diff
@@ -68,6 +68,7 @@ var _ driver.RowsColumnTypeLength = (*rows)(nil)
 var _ dbsqlrows.Rows = (*rows)(nil)
 
 func NewRows(
+	ctx context.Context,
```
rcypher-databricks, line 71: If the only thing we are using ctx for is to extract a logger.  I would rather pass in a logger as the last argument to NewRows and ensure that dbsqllog.AddContext handles a nil first argument.

logger/logger.go
```diff
@@ -123,9 +124,24 @@ func Err(err error) *zerolog.Event {
 	return Logger.Err(err)
 }
 
+// Ctx returns a DBSQLLogger from the provided context. If no logger is found,
+// the default logger is returned.
+func Ctx(ctx context.Context) *DBSQLLogger {
```
rcypher-databricks, line 129: If find this function name a bit confusing.  I know that naming things here is complicated by the use of the noun context and the type context but it would be good to have a function name that more clearly tells you what it is doing.  Maybe something like FromCtx?  I'm open to suggestions.

PR #192, opened 2024-03-11
Add snippet to docs for GET from staging

elise-prequel:
We recently started using the `GET` staging operation after using `PUT` in production for some time. I thought it might be useful to have a snippet with the expected syntax in the documentation. (There is also `REMOVE` but I only see tests for `GET` and `PUT`.)

Not sure yet why I'm triggering linter errors 🤔


andrefurlan-db (approved):
thanks!!

PR #199, opened 2024-04-01
Fix Spark parameter creation when passing a `nil`-value named parameter to a query 

### LINKED ISSUE #193: Issue while converting `nil` SQL parameter to `null`
I was trying to use a query like this. I expected my query to return all results if it's not filtered, and filter the result if there's a filter.

```sql
SELECT *
FROM some_table
WHERE ISNULL(:mmCode) OR name = :mmCode
```

and then I pass `sql.Named("mmCode", code)` to my query.
`code` is a `*string`. It can, of course, be nil.

## Expected behaviour

- If `name` is `nil`, then it would be converted to `NULL` in SQL and trigger the first case in my `WHERE` clause.
- If `name` is not `nil`, then it would become a string and trigger the second case in my `WHERE` clause.

## Actual behaviour

I noticed that the library tries to use `fmt.Sprintf("%s", param.Value)` in [parameters.go](https://github.com/databricks/databricks-sql-go/blob/main/parameters.go#L136-L139). If `param.Value` is `nil`, it becomes `"%s(<nil>)"` and my query won't work.

image

## Workaround

I'm now manually creating my `WHERE` clause depending of the values I have for each filter parameter, but being able to use `ISNULL` would and inject `nil` parameters make my code shorter and cleaner.

I tried to write a PR myself, but I'm not familiar with the Spark parameters API.

esdrasbeleza:
- When a `sql.NamedValue` has the field `Value` set to `nil`, the resulting `cli_service.TSparkParameter` will also have the value `nil` instead of `*cli_service.TSparkParameterValue{StringValue: *"%!s(<nil>"}`.
- Add the type `SqlVoid`, following the conventions used in the [NodeJS connector](https://github.com/databricks/databricks-sql-nodejs/blob/main/lib/DBSQLParameter.ts#L43-L51) and the [Python driver](https://github.com/databricks/databricks-sql-python/blob/f6fd7a7956a4dbc78ad36b5e079fe8d74176a0f1/src/databricks/sql/parameters/native.py#L319-L323).

Fix #193.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -13,12 +13,18 @@ jobs:
 
     steps:
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
+
+      - name: Set up Go Toolchain
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.20.x'
+          cache: false
 
       - name: Lint
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action@v4
         with:
-          version: v1.51
+          version: 'v1.51'
   build-and-test:
     name: Test and Build
     strategy:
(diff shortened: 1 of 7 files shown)


kravets-levko (approved):
LGTM 👍 Though, better if someone of codeowners make a final decision

kravets-levko:
@esdrasbeleza Can you please rebase and resolve conflicts? Thanks!

esdrasbeleza:
@kravets-levko got a bit lost in the process with merges, rebases, signoffs etc., but it's done 😅 
10+ years of git and it's never enough!

kravets-levko:
@esdrasbeleza Merged 👍 I'd like to ship one more PR with this release - waiting for a person to rebase their branch and we're good to go. After that I'll cut a release ASAP

Issue #201, opened 2024-04-08
Expose dbsql.connOption type


shelldandy:
That way we can programmatically add arguments to the `NewConnector` function instead of copy/pasting all of them across conditionals.

For extra context we'd like to not repeat code everywhere depending on whether or not we're using Oauth or AccessTokens since everything else is the same but we end up with 6 calls of the same pretty much code.

PR #202, opened 2024-04-08
Expose dbsql.connOption type

shelldandy:
Fixes: https://github.com/databricks/databricks-sql-go/issues/201

That way we can programmatically add arguments to the `NewConnector` function instead of copy/pasting all of them across conditionals.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -50,7 +50,6 @@ func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
 		},
 		CanUseMultipleCatalogs: &c.cfg.CanUseMultipleCatalogs,
 	})
-
 	if err != nil {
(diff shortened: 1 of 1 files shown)


shelldandy:
Just did a force push to add the sign-off message

shelldandy:
@kravets-levko could i get a review now? i added the sign off message :)

kravets-levko:
Hi @shelldandy! The code on its own looks good, but can you please explain the use case for this change? Maybe some examples how it may be useful? Thank you!

shelldandy:
@kravets-levko please review again, so an use case for example is if we're programatically creating a connector so if in this case we sometimes need to have a connector withtransport but because the type is not exposed currently we're unable to just add arguments before doing the `NewConnector` call:

image

Having this PR merged can help us DRY the code a whole lot.

kravets-levko (approved):
@shelldandy Thank you for the example, now I indeed see the value of exporting this type. Waiting for the tests to run, and if no issues - will merge. Thank you for your effort on this!

PR #209, opened 2024-04-10
feat: Update to Apache Arrow v15

candiduslynx:
Update the apache arrow version to the latest major one.
The v12 used here has some nasty issues like https://github.com/apache/arrow/issues/41156 that prevent the proper Apache Arrow utilization.


candiduslynx:
The change was done via
```sh
gomajor get github.com/apache/arrow/go/v15
go mod tidy
go mod vendor
```
however, I think automating it with renovate would be nice

candiduslynx:
@kravets-levko any chance to get this merged & released, too?
We use arrow v15 & there isn't a straightforward way to map v12 to v15 now (only custom converter), but on the wire it's the same.

candiduslynx:
Closing in favor of https://github.com/databricks/databricks-sql-go/pull/216

PR #215, opened 2024-04-11
Fix formatting of *float64 parameters

### LINKED ISSUE #214: Error while passing a named parameter with a value of type `*float64`
## Summary

If I have a variable typed as `*float64` as pass it down to a query as a named parameter, the query won't work because it will be valued as `"%!s(float64=3.289113e+06)"`.

## Details
I have the following named parameter:
```go
someFloat := float64(3289113)
namedParam := sql.Named("float64Param", &someFloat)
```

When [inferTypes](https://github.com/databricks/databricks-sql-go/blob/e82880f5e0583fcb1228080956d8ca27491f22c3/parameters.go#L89) is called, it's passed down as
```go
Parameter{
  Name: "float64Param",
  Type: SqlUnkown,
  Value: interface{}(float64) 3289113,
}
```
and the function converts it to
```go
Parameter{
  Name: "float64Param",
  Type: SqlString,
  Value: interface {}(string) "%!s(float64=3.289113e+06)", // wrong!
}
```

So if I call my query `SELECT * FROM table WHERE someField = :float64Param`, it won't work because `someField` won't match `"%!s(float64=3.289113e+06)"`.

esdrasbeleza:
Attempt to fix #214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Release History
 
+- Fix formatting of *float64 parameters
+
 ## v1.5.4 (2024-04-10)
 
 - Added OAuth support for GCP (databricks/databricks-sql-go#189 by @rcypher-databricks)
--- a/driver_e2e_test.go
+++ b/driver_e2e_test.go
@@ -504,6 +504,10 @@ func strPtr(s string) *string {
 	return &s
 }
 
+func float64Ptr(f float64) *float64 {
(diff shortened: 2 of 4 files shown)


rcypher-databricks (approved):

esdrasbeleza:
I signed-off my last commit, amended it and force-pushed it to comply with the rules 😄

PR #220, opened 2024-04-19
Fix typo in parameters.go

apmechev:
`Unkown` -> `Unknown`


jackyhu-db:
Thanks for fixing this, but this change will break existing client that uses `SqlUnkown`

apmechev:
makes sense, thanks

Issue #223, opened 2024-05-23
[Feature Request] Add an connOption to skipVerifyHost on NewConnector


madisonchamberlain:
When opening a new connection, I am looking for the ability to allow the driver to skip verifying the host. When we call `NewConnector` I was hoping to have a new connOption which allows the caller to specify that they want to skip verifying the host, but still use TLS encryption when passing data over the wire. The reason I am seeking this is because when we make a connection via private link, we want to avoid using the domain `*.[cloud.databricks.com](http://cloud.databricks.com/)`, because it would confuse our normal databrick connections; rather we want to use `<customer-id>.[privatelink.cloud.databricks.com](http://privatelink.cloud.databricks.com/)` and then on our end since we recognize the host, we dont need you to verify it on your end. The go driver cannot verify `privatelink.cloud.databricks.com` based on the certs it receives. 

There are a few thread about this in a channel shared between databricks and sigma, which may contain more relevant information. 
[Thread 1
](https://sigmacomputing.slack.com/archives/C02QNK01Q2K/p1715878215658859)[Thread 2](https://sigmacomputing.slack.com/archives/C02QNK01Q2K/p1716397003110209)
(channel is #sigma-databricks)

madisonchamberlain:
I was asked to propose a solution so I am proposing this https://github.com/databricks/databricks-sql-go/pull/224

PR #225, opened 2024-05-24
Add connection option WithSkipTLSHostVerify for privatelink host

jackyhu-db:
## Background
The driver cannot connect to the workspace whose hostname is an internal private link hostname, this is because its domain is not added into Databricks workspace certificate.  See https://github.com/databricks/databricks-sql-go/issues/223

## Description
This change adds connection option `WithSkipTLSHostVerify` which can disable verifying the hostname in the server certificate. In this mode, TLS is susceptible to machine-in-the-middle attacks. Please only use this option when the hostname is an internal private link hostname. 

Here is the usage
```go
connector, err := dbsql.NewConnector(
   dbsql.WithServerHostname("<hostname>"),
   dbsql.WithHTTPPath("<http_path>"),
   dbsql.WithAccessToken("<token>"),
   dbsql.WithSkipTLSHostVerify(),
)
```

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


internal/client/client.go
```diff
@@ -545,14 +546,20 @@ func RetryableClient(cfg *config.Config) *http.Client {
 	return retryableClient.StandardClient()
 }
 
-func PooledTransport() *http.Transport {
+func PooledTransport(cfg *config.Config) *http.Transport {
+	var tlsConfig *tls.Config
+	if (cfg.TLSConfig != nil) && cfg.TLSConfig.InsecureSkipVerify {
```
benc-db, line 551: why does this assign only if skipping verification?

jackyhu-db: I want to be backward compatible as much as possible. If it is not set (which will be 99% cases), it will be same as before. `TLSClientConfig` is still `nill`. So no change will be introduced.

benc-db (approved):

benc-db (approved):

kravets-levko (approved):
LGTM 👍

PR #216, opened 2024-04-12
feat: Update Apache Arrow to `v16`

candiduslynx:
Instead of https://github.com/databricks/databricks-sql-go/pull/209

PR #228, opened 2024-06-04
Revert update Apache Arrow to `v16`

kravets-levko:
Reverts databricks/databricks-sql-go#216

Unfortunately, updating Apache Arrow turned out to be a breaking change which created issues to some library users. We made a decision to revert this upgrade


yunbodeng-db (approved):

candiduslynx:
Hi @kravets-levko!
Is there any issue to track the upgrade of the library, as this got reverted?
Basically, is there a plan & outlined steps to roll the arrow/v16 once again?

kravets-levko:
Hi @candiduslynx! Yes, we indeed plan to re-do the upgrade at some point. But we are looking for the way to do this and avoid issues for our users. We don't expect any major changes to the library, so you can continue using `v1.5.5` for now. I'm sorry for all this, but please be patient, once we'll figure things out - I will let you know

datbth:
Is there any place that mentions the breaking changes?
So that we (as users) are aware of the issues when forking/upgrading ourselves?

candiduslynx:
Is this still on v12?
Wow, what a shame.

PR #232, opened 2024-06-07
Set spark param ordinals with NamedValue ordinals

yomikaze:
Querying with unnamed arguments raises this error:
```
ERR databricks: failed to run query error="[INVALID_PARAMETER_MARKER_VALUE.DUPLICATE_NAME] An invalid parameter mapping was provided: multiple parameter mappings were given with the name ''. SQLSTATE: 22023"
```

Unnamed arguments still have ordinal values, so this PR sets the `driver.NamedValue` ordinal value to the spark parameter.


kravets-levko:
Replaced with databricks/databricks-sql-go#247

Issue #242, opened 2024-08-05
Pass ctx properly in internal client.ExecuteStatement


shelldandy:
https://github.com/databricks/databricks-sql-go/blob/93222abd7f5be64af98bd90fe8d60972d365eda6/internal/client/client.go#L162-L171

This needs to actually pass `ctx` so custom authenticators can make use of the context.

The bug was introduced here:

https://github.com/databricks/databricks-sql-go/pull/60/files#diff-3b1f8149deb45d304c69c856bcd1b9b628961fb6000486cb39172a99cb8de514L91-R90

shelldandy:
Seems like this was introduced in this PR:

https://github.com/databricks/databricks-sql-go/pull/60

Issue #231, opened 2024-06-06
Azure managed identities authentication support


shelldandy:
Do we have any sort of timeline on when this feature would be integrated to the driver?

https://learn.microsoft.com/en-us/azure/databricks/dev-tools/auth/#--supported-authentication-types-by-azure-databricks-tool-or-sdk

<img width="723" alt="image" src="https://github.com/databricks/databricks-sql-go/assets/7088511/27df5bdb-fd0c-4e68-9c58-c8428de7e933">

shelldandy:
And if not is there any workaround to getting a client to connect maybe using the go-sdk?

kravets-levko:
Hi @shelldandy! Unfortunately, this type of auth is currently not supported, and there are no estimates for implementing it. You can try to implement it yourself as a custom auth provider, and then use it with `dbsql.WithAuthenticator(authenticator)`:

https://github.com/databricks/databricks-sql-go/blob/d70ab7c5b6dda71f1439bd3edc51a2d2022f7d92/examples/oauth/main.go#L35-L39

shelldandy:
@kravets-levko got this one working but had to do a small change of code, please see this PR:

https://github.com/databricks/databricks-sql-go/pull/244

PR #244, opened 2024-08-05
pass ctx properly in internal client.ExecuteStatement

shelldandy:
Fixes https://github.com/databricks/databricks-sql-go/issues/242
Fixes https://github.com/databricks/databricks-sql-go/issues/231

This needs to actually pass ctx so custom authenticators can make use of the context.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -167,8 +167,7 @@ func (tsc *ThriftServiceClient) ExecuteStatement(ctx context.Context, req *cli_s
 	log, ctx = LoggerAndContext(ctx, req)
 	msg, start := log.Track("ExecuteStatement")
 
-	// We use context.Background to fix a problem where on context done the query would not be cancelled.
-	resp, err := tsc.TCLIServiceClient.ExecuteStatement(context.Background(), req)
+	resp, err := tsc.TCLIServiceClient.ExecuteStatement(ctx, req)
 	log, ctx = LoggerAndContext(ctx, resp)
 	logDisplayMessage(resp, log)
 	logExecStatementState(resp, log)


shelldandy:
@kravets-levko can you take a look at this PR, seems like it was some leftover code from a couple of years ago, im not entirely sure if the comment i removed in there still applies

internal/client/client.go
```diff
@@ -167,8 +167,7 @@ func (tsc *ThriftServiceClient) ExecuteStatement(ctx context.Context, req *cli_s
 	log, ctx = LoggerAndContext(ctx, req)
 	msg, start := log.Track("ExecuteStatement")
 
-	// We use context.Background to fix a problem where on context done the query would not be cancelled.
-	resp, err := tsc.TCLIServiceClient.ExecuteStatement(context.Background(), req)
+	resp, err := tsc.TCLIServiceClient.ExecuteStatement(ctx, req)
```
kravets-levko, line 170: @andrefurlan-db @rcypher-databricks I need your help here. It's unclear to me what (and how) `context.Background()` should fix, and the comment actually doesn't help a lot. If there was any issue with using `ctx` itself - I would like to know what it was, and how can I check if it's still there. Thanks!

shelldandy: @kravets-levko any progress in here getting this merged? any way i can help?

kravets-levko: @shelldandy sorry for the delay. I did try the library a lot with both variants, and haven't notice any differences in behavior. Maybe, `ctx.Background` was needed to fix some very specific case, but I don't know what it is. I tend to merge your fix, just maybe will check it once more just to make sure I don't miss something obvious

shelldandy: @kravets-levko FWIW we've been using my fork for a while without problems on production so I'm also inclined to think that whatever the reason for the background use was has already been fixed

shelldandy: @kravets-levko bumping again to see if we can get this one merged. its needed for use with a custom authenticator that needs the `ctx` to get some values related to the authentication flow

kravets-levko (approved):

PR #247, opened 2024-10-02
[PECO-1962] Support positional query parameters

kravets-levko:
[PECO-1962]

Replaces databricks/databricks-sql-go#232

[PECO-1962]: https://databricks.atlassian.net/browse/PECO-1962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/parameters.go
+++ b/parameters.go
@@ -8,7 +8,9 @@ import (
 	"strings"
 	"time"
 
+	dbsqlerr "github.com/databricks/databricks-sql-go/errors"
 	"github.com/databricks/databricks-sql-go/internal/cli_service"
+	"github.com/pkg/errors"
 )
 
 type Parameter struct {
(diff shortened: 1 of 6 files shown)


parameters.go
```diff
@@ -183,10 +189,24 @@ func convertNamedValuesToSparkParams(values []driver.NamedValue) []*cli_service.
 		} else {
 			sparkParamType = sqlParam.Type.String()
 		}
-		sparkParam := cli_service.TSparkParameter{Name: &sqlParam.Name, Type: &sparkParamType, Value: sparkValue}
+
+		var sparkParamName *string
+		if sqlParam.Name != "" {
```
jackyhu-db, line 194: what if `sqlParam.Name` is `nil`

kravets-levko: The type of `Parameter.Name` is `string` and not `*string`, so it cannot be `nil`. And therefore we use an empty string as an indicator of ordinal parameter (it's not possible to have named parameter with empty name anyway)

jackyhu-db (approved):

PR #255, opened 2025-03-12
Propagate error in TFetchResults

vikrantpuppala:
Currently, an error thrown in TFetchResults within getNextPage in resultPageIterator will be swallowed and an EOF err will be returned instead. This PR solves #254

### DIFF
--- a/internal/rows/rows.go
+++ b/internal/rows/rows.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"database/sql"
 	"database/sql/driver"
-	"io"
 	"math"
 	"reflect"
 	"time"
@@ -454,10 +453,6 @@ func (r *rows) fetchResultPage() error {
 	if r.RowScanner != nil {
 		r.RowScanner.Close()
 		r.RowScanner = nil
-	}
-
-	if !r.ResultPageIterator.HasNext() {
-		return io.EOF
 	}
 
 	fetchResult, err1 := r.ResultPageIterator.Next()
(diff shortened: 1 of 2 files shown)


jackyhu-db (approved):

internal/rows/rows.go
```diff
@@ -456,10 +455,6 @@ func (r *rows) fetchResultPage() error {
 		r.RowScanner = nil
 	}
 
-	if !r.ResultPageIterator.HasNext() {
```
jackyhu-db, line 459: can you check the r.error and return r.error instead?

mdibaiee: @jackyhu-db I think it is possible to do a check like that, but I would like to clarify what is the use of `HasNext` if `Next` itself can just return `nil, io.EOF` to signify an end of the stream?

My question is more structural: why the additional complexity of `HasNext` when `Next` itself can interface and encapsulate the concept of `HasNext`? It is possible to now check for the error in `r.ResultPageIterator.err` and bubble that up, but I'm afraid any divergence between `Next`'s logic and `HasNext`'s logic in the future can lead to similar problems

vikrantpuppala: @jackyhu-db, we would also need to update the function signature to return (bool, error) and also make changes to https://github.com/databricks/databricks-sql-go/blob/main/internal/rows/arrowbased/arrowRecordIterator.go#L86 to also return the error, do we want to go down that route?

jackyhu-db (approved):

PR #261, opened 2025-04-07
[PECOBLR-314] handle thrift protocol version

shivam2680:
## Description
This PR introduces changes to handle thrift protocol version returned in TOpenSessionResp for supporting different features like cloud fetch, compression, parameterized queries.

### Protocol Version Support:
* [`connection.go`](diffhunk://#diff-25a7cf08cc8fc8eb57475835cc380524de2fd1215cda9c6915132ae86dbd80a7L288-R312): Added checks for server protocol version to conditionally support features such as direct results, LZ4 compression, cloud fetch, Arrow format, and parameterized queries. [[1]](diffhunk://#diff-25a7cf08cc8fc8eb57475835cc380524de2fd1215cda9c6915132ae86dbd80a7L288-R312) [[2]](diffhunk://#diff-25a7cf08cc8fc8eb57475835cc380524de2fd1215cda9c6915132ae86dbd80a7L305-R324)
* [`internal/thrift_protocol/protocol_feature_util.go`](diffhunk://#diff-6d5ab0961a6f4e5eee4a29f813d12b51e54f5f45ccaed7360eb6a685e193a1a4R1-R48): Introduced utility functions to check for feature support based on the server protocol version.

## Testing
* [`internal/client/testclient.go`](diffhunk://#diff-23ebebc8a40cb5a963a06a6f99116d364a623dcf4c17979f4d8109b62c024959R13-R15): Updated the test client to include a default server protocol version and a default implementation for opening a session. [[1]](diffhunk://#diff-23ebebc8a40cb5a963a06a6f99116d364a623dcf4c17979f4d8109b62c024959R13-R15) [[2]](diffhunk://#diff-23ebebc8a40cb5a963a06a6f99116d364a623dcf4c17979f4d8109b62c024959L42-R58)
* [`internal/config/config_test.go`](diffhunk://#diff-78b78dc33eda1ed8d509bfa5a00f66709af57bd30d971cbaec454571a91c3e62L651-R651): Changed the default protocol version in the test configuration.

### DIFF
--- a/internal/client/testclient.go
+++ b/internal/client/testclient.go
@@ -10,6 +10,9 @@ import (
 var ErrNotImplemented = errors.New("databricks: not implemented")
 
 type TestClient struct {
+	// Default server protocol version to use in tests
+	ServerProtocolVersion cli_service.TProtocolVersion
+
 	FnOpenSession           func(ctx context.Context, req *cli_service.TOpenSessionReq) (_r *cli_service.TOpenSessionResp, _err error)
 	FnCloseSession          func(ctx context.Context, req *cli_service.TCloseSessionReq) (_r *cli_service.TCloseSessionResp, _err error)
 	FnGetInfo               func(ctx context.Context, req *cli_service.TGetInfoReq) (_r *cli_service.TGetInfoResp, _err error)
@@ -39,7 +42,20 @@ func (c *TestClient) OpenSession(ctx context.Context, req *cli_service.TOpenSess
 	if c.FnOpenSession != nil {
 		return c.FnOpenSession(ctx, req)
 	}
-	return nil, ErrNotImplemented
+
+	// Default implementation for test client
+	resp := &cli_service.TOpenSessionResp{
+		Status:                &cli_service.TStatus{StatusCode: cli_service.TStatusCode_SUCCESS_STATUS},
+		ServerProtocolVersion: c.ServerProtocolVersion,
+		SessionHandle: &cli_service.TSessionHandle{
+			SessionId: &cli_service.THandleIdentifier{
+				GUID:   []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16},
+				Secret: []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16},
+			},
+		},
+	}
+
+	return resp, nil
 }
 func (c *TestClient) CloseSession(ctx context.Context, req *cli_service.TCloseSessionReq) (_r *cli_service.TCloseSessionResp, _err error) {
 	if c.FnCloseSession != nil {
--- a/internal/client/testclient.go
+++ b/internal/client/testclient.go
@@ -10,9 +10,6 @@ import (
 var ErrNotImplemented = errors.New("databricks: not implemented")
 
 type TestClient struct {
-	// Default server protocol version to use in tests
-	ServerProtocolVersion cli_service.TProtocolVersion
-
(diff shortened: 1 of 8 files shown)


jackyhu-db (approved):

internal/client/testclient.go
```diff
@@ -39,7 +42,20 @@ func (c *TestClient) OpenSession(ctx context.Context, req *cli_service.TOpenSess
...
+		ServerProtocolVersion: c.ServerProtocolVersion,
+		SessionHandle: &cli_service.TSessionHandle{
+			SessionId: &cli_service.THandleIdentifier{
+				GUID:   []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16},
+				Secret: []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16},
+			},
+		},
+	}
+
+	return resp, nil
```
vikrantpuppala, line 58: can you explain why is the test client behaviour changed from returning nil, err to resp, nil?

internal/thrift_protocol/protocol_feature_util.go
```diff
@@ -0,0 +1,46 @@
...
+// Supported in SPARK_CLI_SERVICE_PROTOCOL_V1 and above
+func SupportsDirectResults(version cli_service.TProtocolVersion) bool {
+	return version >= cli_service.TProtocolVersion_SPARK_CLI_SERVICE_PROTOCOL_V1
+}
+
+// SupportsLz4Compression checks if the server protocol version supports LZ4 compression
+// Supported in SPARK_CLI_SERVICE_PROTOCOL_V6 and above
+func SupportsLz4Compression(version cli_service.TProtocolVersion) bool {
+	return version >= cli_service.TProtocolVersion_SPARK_CLI_SERVICE_PROTOCOL_V6
+}
```
vikrantpuppala, line 16: wondering if we should add an arg to these functions and do `supports && use` here itself (and change the function names to something like UseLz4Compression). What's your opinion?

internal/thrift_protocol/protocol_feature_util.go
vikrantpuppala, line 1: should we add unit tests?

internal/thrift_protocol/protocol_feature_util.go
shivam2680: I feel the present implementation is fine. The purpose of util file is to check if a feature is supported by particular thrift version and not to check whether to use this feature. Further, this helps in better identifying what caused a feature to be disabled (thrift version/config).

gopalldb (dismissed):

internal/thrift_protocol/protocol_feature_util.go
vikrantpuppala: but would we ever do this separately? i.e. "check if a feature is supported by particular thrift version and not to check whether to use this feature". Can you elaborate on: "this helps in better identifying what caused a feature to be disabled (thrift version/config).", i do not see any logs etc. how is this captured?

shivam2680: > do not see any logs etc. how is this captured

Yes, I wanted to add logs in protocol util, but then I thought there will be too many logs because for every ExecuteStatement there might be multiple logs.
> but would we ever do this separately?

The protocol package only concerns itself with protocol capabilities, not application configuration.

internal/client/testclient.go
shivam2680: Thought I'll need this for adding tests in connection_test. But it's possible without this also. I just created a test session and assigned specific session.ServerProtocolVersion = version

internal/thrift_protocol/protocol_feature_util.go
shivam2680: added

vikrantpuppala (approved):

PR #267, opened 2025-05-05
Add schema to ArrowBatchIterator

vikrantpuppala:
Allow ability to get the arrow schema when fetching Arrow Batches.

Currently, the GetArrowBatches(ctx) method in the DBSQL driver does not expose schema information, this PR adds support to get Schema directly like the Arrow Flight interface.


jackyhu-db (approved):

donatobarone (approved):
Done a test on the FactSet side, change looks good!

PR #277, opened 2025-06-19
upgrade arrow from v12 to v18

jadewang-db:


go.mod
```diff
@@ -1,50 +1,47 @@
 module github.com/databricks/databricks-sql-go
 
-go 1.20
+go 1.22.0
```
vikrantpuppala, line 3: Based on previous conversations with @jackyhu-db, we do not want to update go versions since this would break customer workloads

PR #285, opened 2025-09-17
Prepare for v1.9.0 release

sreekanth-db:
Prepare for v1.9.0 release


sreekanth-db:
> why is this change a major version bump?

This is adding support for a new feature (query tags), so updating the minor version instead of patch version.  (https://semver.org/)

samikshya-db (approved):

PR #291, opened 2025-10-30
Token federation for Go driver (2/3)

madhav-db:
Adds token federation for databricks sql go driver

### DIFF
--- a/auth/tokenprovider/cached.go
+++ b/auth/tokenprovider/cached.go
@@ -11,9 +11,10 @@ import (
 
 // CachedTokenProvider wraps another provider and caches tokens
 type CachedTokenProvider struct {
-	provider TokenProvider
-	cache    *Token
-	mutex    sync.RWMutex
+	provider   TokenProvider
+	cache      *Token
+	mutex      sync.RWMutex
+	refreshing bool // prevents thundering herd
 	// RefreshThreshold determines when to refresh (default 5 minutes before expiry)
 	RefreshThreshold time.Duration
 }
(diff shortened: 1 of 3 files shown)


auth/oauth/oauth.go
```diff
@@ -85,6 +85,8 @@ var databricksAWSDomains []string = []string{
 }
 
 var databricksAzureDomains []string = []string{
+	".staging.azuredatabricks.net",
+	".dev.azuredatabricks.net",
```
vikrantpuppala: assume this is only for testing? we shouldn't probably add these?

madhav-db: Correct, had added them to test, removed now.

auth/tokenprovider/cached.go
```diff
@@ -0,0 +1,86 @@
...
+	mutex    sync.RWMutex
+	// RefreshThreshold determines when to refresh (default 5 minutes before expiry)
+	RefreshThreshold time.Duration
+}
+
+// NewCachedTokenProvider creates a caching wrapper around any token provider
+func NewCachedTokenProvider(provider TokenProvider) *CachedTokenProvider {
+	return &CachedTokenProvider{
+		provider:         provider,
+		RefreshThreshold: 5 * time.Minute,
```
vikrantpuppala: isn't this quite high?

auth/tokenprovider/exchange.go
```diff
@@ -0,0 +1,204 @@
...
+	returnOriginalTokenIfAuthenticated bool
+}
+
+// NewFederationProvider creates a federation provider that wraps another provider
+// It automatically detects when token exchange is needed and falls back gracefully
+func NewFederationProvider(baseProvider TokenProvider, databricksHost string) *FederationProvider {
+	return &FederationProvider{
+		baseProvider:                       baseProvider,
+		databricksHost:                     databricksHost,
+		httpClient:                         &http.Client{Timeout: 30 * time.Second},
```
vikrantpuppala: do we not have any http clients in sql go that can be used here rather than creating a new one?

auth/tokenprovider/cached.go
```diff
@@ -0,0 +1,86 @@
...
+func (p *CachedTokenProvider) GetToken(ctx context.Context) (*Token, error) {
+	// Try to get from cache first
+	p.mutex.RLock()
+	cached := p.cache
+	p.mutex.RUnlock()
+
+	if cached != nil && !p.shouldRefresh(cached) {
+		log.Debug().Msgf("cached token provider: using cached token for provider %s", p.provider.Name())
+		return cached, nil
+	}
```
vikrantpuppala: After releasing the read lock, another goroutine could modify `cached.ExpiresAt`, making the `shouldRefresh` check stale. Additionally, returning a pointer to the cached token is unsafe.

auth/tokenprovider/cached.go
```diff
@@ -0,0 +1,86 @@
...
+	return token, nil
+}
+
+// shouldRefresh determines if a token should be refreshed
+func (p *CachedTokenProvider) shouldRefresh(token *Token) bool {
+	if token == nil {
+		return true
+	}
+
+	// If no expiry time, assume token doesn't expire
```
vikrantpuppala: can we add more documentation on when this happens?

auth/tokenprovider/exchange.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// Use original token
+	return baseToken, nil
+}
+
+// needsTokenExchange determines if a token needs exchange by checking if it's from a different issuer
+func (p *FederationProvider) needsTokenExchange(tokenString string) bool {
+	// Try to parse as JWT
+	token, _, err := new(jwt.Parser).ParseUnverified(tokenString, jwt.MapClaims{})
```
vikrantpuppala: why do we do parse unverified?

auth/tokenprovider/exchange.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// Check if issuer is different from Databricks host
+	return !p.isSameHost(issuer, p.databricksHost)
+}
+
+// tryTokenExchange attempts to exchange the token with Databricks
+func (p *FederationProvider) tryTokenExchange(ctx context.Context, subjectToken string) (*Token, error) {
+	// Build exchange URL - add scheme if not present
+	exchangeURL := p.databricksHost
+	if !strings.HasPrefix(exchangeURL, "http://") && !strings.HasPrefix(exchangeURL, "https://") {
```
vikrantpuppala: Should we allow `http://` prefix, might be insecure for token exchange.

auth/tokenprovider/exchange.go
```diff
@@ -0,0 +1,204 @@
...
+		AccessToken string `json:"access_token"`
+		TokenType   string `json:"token_type"`
+		ExpiresIn   int    `json:"expires_in"`
+		Scope       string `json:"scope"`
+	}
+
+	if err := json.Unmarshal(body, &tokenResp); err != nil {
+		return nil, fmt.Errorf("failed to parse response: %w", err)
+	}
+
```
vikrantpuppala: should we add validations on token resp here

auth/tokenprovider/federation_test.go
```diff
@@ -0,0 +1,348 @@
...
+	assert.Equal(t, "databricks-token", token1.AccessToken)
+	assert.Equal(t, 1, callCount, "External provider should be called once")
+	assert.Equal(t, 1, exchangeCount, "Token should be exchanged once")
+
+	// Second call - should use cache
+	token2, err2 := cachedProvider.GetToken(ctx)
+	require.NoError(t, err2)
+	assert.Equal(t, "databricks-token", token2.AccessToken)
+	assert.Equal(t, 1, callCount, "External provider should still be called only once (cached)")
+	assert.Equal(t, 1, exchangeCount, "Token should still be exchanged only once (cached)")
```
vikrantpuppala: should we also add assert that token1 == token2

auth/tokenprovider/cached.go
```diff
@@ -0,0 +1,86 @@
...
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	// Double-check after acquiring write lock
+	if p.cache != nil && !p.shouldRefresh(p.cache) {
+		return p.cache, nil
+	}
+
+	log.Debug().Msgf("cached token provider: fetching new token from provider %s", p.provider.Name())
+	token, err := p.provider.GetToken(ctx)
```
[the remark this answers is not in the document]
madhav-db: Asked Claude to fix this and I think it did a fairly good job.
Refactored to prevent this issue:

  1. Added a refreshing bool flag to coordinate between goroutines
  2. Mark as refreshing and release the mutex BEFORE calling `provider.GetToken()`
  3. Implement wait-and-retry pattern for goroutines that arrive during refresh
  4. Re-acquire mutex only to update the cache after the fetch completes

  This prevents:
  - Holding the lock during slow HTTP calls
  - Potential deadlocks
  - Thundering herd when multiple goroutines try to refresh simultaneously

auth/tokenprovider/cached.go
madhav-db: Fixed.
Now I'm reading both the cached token and the `shouldRefresh` result while holding the read lock. I return a copy of the token to prevent modification and callers now receive independent token copies, so the cache is thread-safe.

auth/tokenprovider/cached.go
madhav-db: Done

auth/tokenprovider/federation_test.go
madhav-db: Done

auth/tokenprovider/exchange.go
madhav-db: I've kept support for http:// but added a warning log when it's used

auth/tokenprovider/exchange.go
```diff
@@ -0,0 +1,204 @@
...
+	req, err := http.NewRequestWithContext(ctx, "POST", exchangeURL, strings.NewReader(data.Encode()))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Accept", "*/*")
+
+	// Make request
+	resp, err := p.httpClient.Do(req)
```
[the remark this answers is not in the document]
madhav-db: I don't think so. Token exchange is a separate concern from sql query operations, and this http client meets those needs (like graceful fallback of token in case of exchange failing instead of aggressive retries)

auth/tokenprovider/exchange.go
madhav-db: We only need to inspect the issuer claim, not validate the signature. Token validation is done during the exchange process. This is done just to determine whether the exchange is needed

auth/tokenprovider/exchange.go
madhav-db: Token exchange is a separate concern from the SQL query operations, I think it is fine.

Issue #288, opened 2025-10-24
Bug: Driver discards context, breaking context-based authentication


dgiagio:
Hello, we've identified a critical issue where the driver discards the `context.Context` during polling, making it impossible to use authentication mechanisms (like Azure OBO) that rely on passing credentials via the context. This also appears to be an issue during connection `Close()`, which leads to intermittent failures.

### Bug 1: Polling context loss
This bug causes all long-running queries to fail when using context-based authentication.

When a query is executed (e.g., via QueryContext), it correctly receives a context containing authentication credentials. However, if the query is long-running, the driver uses a polling sentinel to check for results. This sentinel is explicitly created with a new, empty context using `context.Background()`, discarding the credentials.

**Code reference:** https://github.com/databricks/databricks-sql-go/blob/main/connection.go#L365-L370

When our authenticator is invoked to sign the poll request, it receives this `newCtx` empty context, cannot find the required credentials (e.g., OBO token), and authentication fails.

**Suggested fix:** Create the `newCtx` based on the passed in `ctx`, and not `context.Background()`.

### Bug 2: Connection `Close()` context loss
The standard Go `database/sql` package manages the connection pool. When it needs to refresh a connection (e.g., due to lifetime expiration), it calls the `Close()` method on the driver connection.

The `driver.Conn` interface for `Close()` does not accept a `Context.context`. If the driver's `Close()` implementation needs to make an authenticated HTTP call to Databricks (e.g., to gracefully close a session), it has no context to pass to the authenticator. This leads to the same authentication failure as Bug 1.

**Suggested fix:** This is a more complex issue due to `database/sql` interface limitations. We would request that the driver team investigate if it's possible to avoid making authenticated HTTP calls within the `Close()` method.

PR #295, opened 2025-11-04
Fix context loss in polling and connection close operations

dgiagio:
This PR fixes an issue where the driver discards the `context.Context` during polling, making it impossible to use authentication mechanisms (like Azure OBO) that rely on passing credentials via the context.

Reference:
- https://github.com/databricks/databricks-sql-go/issues/288
- https://github.com/grafana/grafana/issues/112955

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -16,6 +16,7 @@ import (
 	dbsqlerr "github.com/databricks/databricks-sql-go/errors"
 	"github.com/databricks/databricks-sql-go/internal/cli_service"
 	"github.com/databricks/databricks-sql-go/internal/client"
+	context2 "github.com/databricks/databricks-sql-go/internal/compat/context"
 	"github.com/databricks/databricks-sql-go/internal/config"
 	dbsqlerrint "github.com/databricks/databricks-sql-go/internal/errors"
 	"github.com/databricks/databricks-sql-go/internal/rows"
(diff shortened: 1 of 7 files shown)


connection.go
```diff
@@ -367,7 +365,7 @@ func (c *conn) pollOperation(ctx context.Context, opHandle *cli_service.TOperati
 	log := logger.WithContext(c.id, corrId, client.SprintGuid(opHandle.OperationId.GUID))
 	var statusResp *cli_service.TGetOperationStatusResp
 	ctx = driverctx.NewContextWithConnId(ctx, c.id)
-	newCtx := driverctx.NewContextWithCorrelationId(driverctx.NewContextWithConnId(context.Background(), c.id), corrId)
+	newCtx := context.WithoutCancel(ctx)
```
gopalldb: if context is cancelled, polling will not have cancellation information. Won't this affect cancellation of long running queries?

internal/rows/rows.go
```diff
@@ -67,22 +67,22 @@ var _ driver.RowsColumnTypeLength = (*rows)(nil)
...
 
 func NewRows(
-	connId string,
-	correlationId string,
+	ctx context.Context,
 	opHandle *cli_service.TOperationHandle,
 	client cli_service.TCLIService,
 	config *config.Config,
 	directResults *cli_service.TSparkDirectResults,
 ) (driver.Rows, dbsqlerr.DBError) {
```
gopalldb: do we need a check for context being nil?

connection.go
dgiagio: I see your concern, but in this specific implementation, cancellation is handled by `ctx` rather than through the new context.

Even though `newCtx` doesn't carry the cancellation signal, the `pollSentinel` is created using the original `ctx`. The `pollSentinel.Watch` function explicitly monitors that original context. As soon as it detects a timeout or cancel signal, it invokes `OnCancelFn` to terminate the operation.

So, effectively, `newCtx` doesn't need the cancellation info because pollSentinel is already handling that responsibility via `ctx`.

PR #306, opened 2025-11-24
oauth/m2m: make "token fetched successfully" a debug log

mdibaiee:
The log is quite repetitive, and as we use this sql driver in our customer-facing product where customers see logs, the extra log seems like spam to customers. I think it would be nice to have it as debug log instead of info.


vikrantpuppala (approved):

vikrantpuppala:
thank you!

Issue #307, opened 2025-11-24
CloudFetch transport layer not configurable


Multimo:
We are using the driver with large queries and when there is a large enough query that comes through, the driver swaps to cloudfetch. 

Unfortunately the instance is behind a private network, smaller queries work but larger queries fail with a networkiong error: `host is unavailable`. 

The httpClient the driver uses for most queries has a configurable transport however the [httpClient that cloudfetch uses](https://github.com/databricks/databricks-sql-go/blob/main/internal/rows/arrowbased/batchloader.go#L321) is not configurable unless you modify the global `http.DefaultClient`

We would like to ensure that all the drivers httpclients have a configurable transport layer that they all functionality can be used when a private network is involved.

PR #308, opened 2025-11-24
Cloudfetch: Allow configuration of httpclient for cloudfetch

Multimo:
[issues](https://github.com/databricks/databricks-sql-go/issues/307)

Hello, as per issue looking for ways to modify the transport layer of the httpclient that cloudfetch uses. Happy to go another way to solving this, this just seamed like the simplest. Thanks for your work on the driver, its been very useful 👍

### DIFF
--- a/internal/rows/arrowbased/batchloader_test.go
+++ b/internal/rows/arrowbased/batchloader_test.go
@@ -257,7 +257,10 @@ func TestCloudFetchIterator(t *testing.T) {
 	t.Run("should use custom HTTPClient when provided", func(t *testing.T) {
 		handler = func(w http.ResponseWriter, r *http.Request) {
 			w.WriteHeader(http.StatusOK)
-			w.Write(generateMockArrowBytes(generateArrowRecord()))
+			_, err := w.Write(generateMockArrowBytes(generateArrowRecord()))
+			if err != nil {
+				panic(err)
+			}
 		}
 
 		startRowOffset := int64(100)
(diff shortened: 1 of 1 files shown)


samikshya-db:
Thanks for making the clean fix, @Multimo ! 
In the long term we would ideally want to have a single httpClient throughout, with configurable options made available to the user. But as a short term fix, this looks good. 

One minor comment on renaming.

connector.go
```diff
@@ -269,6 +269,13 @@ func WithCloudFetch(useCloudFetch bool) ConnOption {
 	}
 }
 
+// WithCloudFetchHTTPClient allows a custom http client to be used for cloud fetch. Default is http.DefaultClient.
+func WithCloudFetchHTTPClient(httpClient *http.Client) ConnOption {
```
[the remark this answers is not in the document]
Multimo: An alternative, until you merge the two http clients, could be to use have the WithTransport add to the CloudFetchConfig also. I can update to do this if you prefer.

internal/rows/arrowbased/batchloader.go
```diff
@@ -317,9 +322,15 @@ func fetchBatchBytes(
 		}
 	}
 
+	httpClient := http.DefaultClient
+	if transport != nil {
+		httpClient = &http.Client{
+			Transport: transport,
```
samikshya-db: Does this mean a new HttpClient is created on every fetch request when a custom transport is provided? If yes, can we make that more optimal?

Multimo: ya good catch! That would not be ideal. Storing the httpClient in the CloudFetchConfig now so its lifetime is much longer.

samikshya-db:
Thanks @Multimo !! Merged 🚢

Multimo:
Thanks @samikshya-db !

mohammedkhu:
@Multimo @samikshya-db Just following through some of these PR's, could the fix be also needed for python connector as well for the issue https://github.com/databricks/databricks-sql-go/issues/307 ?
I realize this is only for go based connector. 
Context: We are observing similar issue on our python side of things, so wondering if this has to do with this.
Thanks

PR #311, opened 2025-12-15
Add circuit breaker for telemetry + amend design plan to be similar to include more details on circuit breaker and config

samikshya-db:
### Changes : 

The changes in this PR : 
1. Add config params for telemetry similar to oss jdbc 
2. Add circuit breaker logic code (Note : this will be integrated at a later point in time)
3. Keep the lowest level of logging for telemetry (to keep telemetry as silent as possible)

The original design doc was changed to : 
1. Keep config names similar to the ones used in JDBC. 
2. Add more details to circuit breaker logic. (This will be useful from a spec driven development perspective )

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -260,6 +260,12 @@ func WithAuthenticator(authr auth.Authenticator) ConnOption {
 func WithTransport(t http.RoundTripper) ConnOption {
 	return func(c *config.Config) {
 		c.Transport = t
+
+		if c.CloudFetchConfig.HTTPClient == nil {
+			c.CloudFetchConfig.HTTPClient = &http.Client{
+				Transport: t,
+			}
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


telemetry/config.go
```diff
@@ -47,12 +57,25 @@ func DefaultConfig() *Config {
...
+		if v == "true" || v == "1" {
+			cfg.ForceEnableTelemetry = true
+			cfg.Enabled = true // Also set Enabled for backward compatibility
+		}
+	}
+
+	// Check for enableTelemetry flag (respects server feature flags)
+	if v, ok := params["enableTelemetry"]; ok {
+		if v == "true" || v == "1" {
+			cfg.EnableTelemetry = true
```
sreekanth-db: does `Enabled` ultimately store whether telemetry is active or not (final source of truth to check whether telemetry is active - determined based on `ForceEnableTelemetry` or `EnableTelemetry` in combination with server flag) ? if so, then shouldn't we update that variable as well here based on the server flag ?

samikshya-db: Thanks for the review, Sreekanth. 

The `Enabled ` is intended to be final source of truth, But the actual closure for this is yet to come in phase 5 of impl. (I.e., not part of this PR)

PR #313, opened 2026-01-12
Fix int64/uint64 type inference to use BIGINT instead of INTEGER

vikrantpuppala:
## Summary
- Fixes a bug where int64 and uint64 values were incorrectly mapped to SqlInteger instead of SqlBigInt
- Fixes data loss issue where int64 was converted using `strconv.Itoa(int(value))` which truncates large values
- Adds safe type assertion in `convertNamedValuesToSparkParams` to prevent panic when `Parameter.Value` is not a string

## Problem
When inserting int64/uint64 values into BIGINT columns, the driver was sending them with type `INTEGER` instead of `BIGINT`, causing the server to reject large values with error:
```
[INVALID_PARAMETER_MARKER_VALUE.INVALID_VALUE_FOR_DATA_TYPE] An invalid parameter mapping was provided: 
the value '1311768467463790320' for parameter 'null' cannot be cast to INT because it is malformed.
```

## Test plan
- [x] Added unit tests for int64/uint64 type inference (`TestParameter_BigInt`)
- [x] Verified large int64 values are correctly inserted and retrieved from BIGINT columns
- [x] All existing parameter tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/parameter_test.go
+++ b/parameter_test.go
@@ -37,6 +37,61 @@ func TestParameter_Inference(t *testing.T) {
 	})
 }
 
+func TestParameter_BigInt(t *testing.T) {
+	t.Run("Should infer int64 as BIGINT", func(t *testing.T) {
+		maxInt64 := int64(9223372036854775807)
+		values := []driver.NamedValue{
+			{Value: maxInt64},
+		}
+		parameters, err := convertNamedValuesToSparkParams(values)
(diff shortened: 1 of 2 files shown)


vikrantpuppala:
Closing in favor of #316 which includes this fix along with the float64 fix

PR #344, opened 2026-04-06
Add enforceEmbeddedSchemaCorrectness to TExecuteStatementReq

gopalldb:
## Summary
- Adds a new optional bool field `enforceEmbeddedSchemaCorrectness` (field ID `0xD19` / 3353) to `TExecuteStatementReq` in the thrift contract
- Exposes it as an opt-in configuration parameter (`EnforceEmbeddedSchemaCorrectness`, default `false`) via connector option, DSN parameter, and config struct
- When enabled (`true`), the field is set in the thrift request so the server enforces embedded schema correctness during query execution

## Changes
- `internal/cli_service/cli_service.go` — Added field with full Read/Write/IsSet/Equals support following existing `*bool` pointer pattern
- `internal/config/config.go` — Added `EnforceEmbeddedSchemaCorrectness` to `UserConfig` + DSN parsing
- `connector.go` — Added `WithEnforceEmbeddedSchemaCorrectness()` connector option
- `connection.go` — Wired config to `TExecuteStatementReq` in `executeStatement()`

## Test plan
- [ ] Verify compilation passes
- [ ] Test with `enforceEmbeddedSchemaCorrectness=true` in DSN
- [ ] Test with `WithEnforceEmbeddedSchemaCorrectness(true)` connector option
- [ ] Verify field is serialized in thrift request when enabled
- [ ] Verify field is not sent when disabled (default)

Resolves: ES-1804970


internal/cli_service/cli_service.go
```diff
@@ -11894,6 +11897,13 @@ func (p *TExecuteStatementReq) GetStatementConf() *TStatementConf {
   }
 return p.StatementConf
 }
+var TExecuteStatementReq_EnforceEmbeddedSchemaCorrectness_DEFAULT bool
+func (p *TExecuteStatementReq) GetEnforceEmbeddedSchemaCorrectness() bool {
```
samikshya-db, line 11901: Was this manually added and not by thrift command?

vikrantpuppala (approved):

internal/cli_service/cli_service.go
vikrantpuppala: +1, let's make sure this was added by the thrift cli

gopalldb: generated by thrift command now

PR #346, opened 2026-04-10
[ES-1804970] Fix CloudFetch returning stale column names from cached results

sreekanth-db:
## Summary

Fixes a bug where `arrow.Record.Schema()` returns stale column aliases when CloudFetch serves cached Arrow IPC files from a structurally identical prior query with different `AS` aliases.

- **Root cause:** `NewCloudBatchIterator` was not receiving the authoritative schema bytes from `GetResultSetMetadata`, unlike the local batch path which already had this. CloudFetch Arrow IPC files have column names baked in from the original query, and the driver was reading them as-is.
- **Fix:** Pass `arrowSchemaBytes` (the authoritative schema from `GetResultSetMetadata`) into `NewCloudBatchIterator`. After records are deserialized from the IPC stream, replace the stale schema with the authoritative one using `array.NewRecord()` (zero-copy — shares underlying column data, only swaps metadata).

## Changes

- **`arrowRecordIterator.go`** — Pass `ri.arrowSchemaBytes` to `NewCloudBatchIterator` in `newBatchIterator()`
- **`arrowRows.go`** — Pass `schemaBytes` to `NewCloudBatchIterator` in `NewArrowRowScanner()`
- **`batchloader.go`** — Core fix:
  - `NewCloudBatchIterator` accepts `arrowSchemaBytes`, parses into `*arrow.Schema`, stores on `batchIterator`
  - `batchIterator.Next()` applies override schema to CloudFetch records only (local path is untouched, `overrideSchema` is `nil`)
  - Added `schemaFromIPCBytes()` helper
  - Field count validation guard to prevent panics on schema mismatch
  - Schema parse failure logged at `Warn` level
- **`batchloader_test.go`** — Added `TestCloudFetchSchemaOverride` with two subtests:
  - Verifies stale column names `["id","name"]` are overridden to `["x","y"]`
  - Verifies `nil` schema bytes pass through original names unchanged

## Who is affected

Go driver users with CloudFetch enabled (`WithCloudFetch(true)`) who read `arrow.Record.Schema()` directly. Python, ODBC, and JDBC drivers are not affected.

## Test plan

- [x] All existing unit tests pass (37 tests in `internal/rows/arrowbased/`)
- [x] New unit test `TestCloudFetchSchemaOverride` covers the override and no-override paths
- [x] Verified end-to-end against a real Databricks warehouse using `samples.tpch.lineitem` (~30M rows) with two queries differing only in column aliases — confirmed `arrow.Record.Schema()` now returns correct aliases


gopalldb (approved):
Can we test this?

sreekanth-db:
> Can we test this?

yes, tested on real warehouse

PR #347, opened 2026-04-13
Add SPOG (Custom URL) routing support via x-databricks-org-id header

msrathore-db:
## Summary

On SPOG (Custom URL / account-level) workspaces, `httpPath` has the form `/sql/1.0/warehouses/<id>?o=<workspaceId>`. The `?o=` parameter routes Thrift calls correctly via the URL, but other endpoints (telemetry push, feature-flag check) run on separate hosts and need `x-databricks-org-id` as an HTTP header to route to the right workspace. Without it, those requests 404 or get misrouted on SPOG hosts.

## Change

All contained in `connector.go`:

1. `extractSpogHeaders(httpPath string) map[string]string` — parses the `?o=` query param using `url.ParseQuery` (stdlib, not regex). Returns `{"x-databricks-org-id": "<workspaceId>"}` or `nil`. Three DEBUG log paths cover: malformed query, missing `?o=`, and successful extraction.

2. `headerInjectingTransport` — a lightweight `http.RoundTripper` wrapper that clones the request per the contract and sets the provided headers if not already set by the caller.

3. `withSpogHeaders(base *http.Client, headers map[string]string) *http.Client` — returns a new client with the same settings but a wrapped transport.

4. In `Connect()`, when `extractSpogHeaders` returns non-nil, the driver passes a wrapped client into `TelemetryInitOptions.HTTPClient`. The wrapped client is used for both the feature-flag check and the telemetry push. The driver's own `c.client` is left alone, so Thrift routing (which uses `?o=` in the URL) is unaffected.

## Why a transport wrapper instead of threading a parameter

An earlier revision of this PR threaded an `extraHeaders` parameter through `telemetry.TelemetryInitOptions` → `isTelemetryEnabled` → `featureFlagCache.isTelemetryEnabled` → `fetchFeatureFlag`. That approach:

- Required API-surface changes in 3 telemetry files (`config.go`, `featureflag.go`, `driver_integration.go`).
- Only covered the feature-flag check; `telemetry/exporter.go` (telemetry push) still sent `Content-Type` as the only header — SPOG routing would 404 at push time.

The RoundTripper wrapper:

- Keeps `telemetry/*` identical to `origin/main`. Zero API churn.
- Automatically applies to every outbound request using the wrapped client — feature-flag check, telemetry push, and any future HTTP paths that reuse it.
- Respects caller-set headers (if a request already has `x-databricks-org-id` for some reason, the wrapper does not override).

## Endpoints covered

| Endpoint | Uses wrapped client? | Gets `x-databricks-org-id`? |
| :---- | :----: | :----: |
| Feature flags `/api/2.0/connector-service/feature-flags/GOLANG/{v}` | Yes | ✅ |
| Telemetry push | Yes | ✅ |
| Thrift | No (uses `c.client` directly; already routes via URL) | — (not needed — URL-routed) |
| OAuth token exchange | No (separate client, talks to `login.microsoftonline.com`) | — (not needed) |
| CloudFetch / Volume operations | No (presigned URLs) | — (not needed) |

## Verification

- `go build ./...` clean.
- Debug log output confirms extraction on SPOG URLs:
  ```
  SPOG header extraction: injecting x-databricks-org-id=<id> (extracted from ?o= in httpPath)
  ```

## Note on the earlier `auth/oauth/u2m/u2m.go` change

The two earliest commits on this branch (`23697e5`, `0ec7e06`) modified `auth/oauth/u2m/u2m.go` to avoid sending an empty `client_secret` on the PKCE public-app flow, documented as a fix for the server's `"Public app should not use a client secret"` rejection. **That change was empirically verified to not be needed** (see commit `3576c92` which reverts it):

- **Prod Legacy** (`adb-6436897454825492.12.azuredatabricks.net`): Unpatched `u2m.go` PASSES end-to-end. Server accepts the empty `client_secret`.
- **Stg Legacy** (`adb-7064161269814046.2.staging.azuredatabricks.net`): Unpatched `u2m.go` FAILS with `unexpected HTTP status 400 Bad Request` during token exchange.

Since production tolerates the current behavior, customers aren't impacted. Keeping this PR minimal; if staging-level strictness later rolls out to prod, we can re-add the u2m fix then.

## Test plan

- [ ] Unit tests pass (`go test ./...`)
- [ ] Manually verified SPOG header injection appears in `x-databricks-org-id` on feature-flag check and telemetry push against a SPOG workspace (not gated by this PR, but prerequisite for SPOG telemetry to work at all)
- [ ] No regressions on non-SPOG workspaces (wrapper is a no-op when `extractSpogHeaders` returns nil)


samikshya-db (approved):
Thanks for this, please add in nodejs too 🙏

PR #348, opened 2026-04-13
Move CI to Databricks protected runners with JFrog OIDC

vikrantpuppala:
## Summary
- Add `.github/actions/setup-jfrog` composite action for OIDC-based JFrog authentication (configures GOPROXY and `.netrc` for Go module proxy)
- Switch all workflow jobs (`lint`, `build-and-test`, `dco-check`) from `ubuntu-latest` to `databricks-protected-runner-group`
- Add `id-token: write` permission for JFrog OIDC token exchange

## Test plan
- [ ] DCO check workflow passes on this PR
- [ ] Lint job passes with Go modules resolved through JFrog proxy
- [ ] Build and test job passes with Go modules resolved through JFrog proxy
- [ ] Verify JFrog OIDC token exchange works on protected runners

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -8,15 +8,21 @@ on:
 
 permissions:
   contents: read
+  id-token: write
 
 jobs:
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on:
+      group: databricks-protected-runner-group
+      labels: linux-ubuntu-latest
 
     steps:
       - name: Check out code into the Go module directory
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -33,13 +39,17 @@ jobs:
     strategy:
       matrix:
         go-version: [1.20.x]
```
shivam2680, line 41: llm comment on EOL version 1.20.x. 
switch to 1.21 or 1.22 + later versions

shivam2680 (approved):

.github/workflows/go.yml
vikrantpuppala: cannot update go version, certain customer apps require 1.20.x compatibility

PR #355, opened 2026-05-07
[ES-1892645] Retry transient S3 errors in CloudFetch downloads

jayantsing-db:
fetchBatchBytes had no retry on HTTP failures, so any single 5xx from S3 cancelled the entire query. With thousands of concurrent GETs on large result sets, even a sub-percent per-request failure rate makes at least one failure near-certain. Adds exponential backoff with equal jitter for 408/429/500/502/503/504 plus connection errors, honoring the existing RetryMax / RetryWaitMin / RetryWaitMax config and parseable integer Retry-After response headers. Link expiry is re-checked after each backoff so retries don't outlive the presigned URL.

Co-authored-by: Isaac

### DIFF
--- a/internal/rows/arrowbased/batchloader.go
+++ b/internal/rows/arrowbased/batchloader.go
@@ -304,6 +304,9 @@ func (cft *cloudFetchDownloadTask) Run() {
 			cft.link.RowCount,
 		)
 		downloadStart := time.Now()
+		// fetchBatchBytes now buffers the body in memory and retries on
+		// mid-stream failures, so the returned reader is always a
+		// *bytes.Reader and needs no Close.
 		data, err := fetchBatchBytes(
 			cft.ctx,
 			cft.link,
@@ -319,9 +322,10 @@ func (cft *cloudFetchDownloadTask) Run() {
 			return
 		}
 
-		// Read all data into memory before closing
+		// Decompression sits outside the retry loop on purpose: a malformed
+		// LZ4 frame is data corruption, not a transient network condition,
+		// and won't recover on retry.
 		buf, err := io.ReadAll(getReader(data, cft.useLz4Compression))
-		data.Close() //nolint:errcheck,gosec // G104: close after reading data
 		downloadMs := time.Since(downloadStart).Milliseconds()
 		if err != nil {
 			cft.sendResult(cloudFetchDownloadTaskResult{data: nil, err: err})
@@ -369,11 +373,18 @@ func logCloudFetchSpeed(fullURL string, contentLength int64, duration time.Durat
 }
 
 // fetchBatchBytes downloads a single Cloud Fetch result link from object
-// storage. Transient failures — connection errors and HTTP 408/429/500/502/503/504
-// from S3-style endpoints — are retried up to retryMax times with exponential
-// backoff and equal jitter. Link expiry is rechecked before every attempt: a
-// long retry chain can outlive a presigned URL, and continuing past expiry is
-// guaranteed to fail.
+// storage and returns the raw (still-compressed, if any) response body
+// buffered in memory. Both connection-time failures and mid-stream body-read
+// failures are retried up to retryMax times with exponential backoff and
+// equal jitter, alongside HTTP 408/429/500/502/503/504. The body read
+// happens inside the retry loop on purpose: with multi-MB S3 objects, a TCP
+// RST or truncated response surfaces as an io.ReadAll error *after* the
+// 200 OK headers have already arrived, and that's exactly the failure mode
+// the customer hits at scale. Decompression and IPC parsing are left to the
+// caller — those errors aren't transient, so retrying them is wasted work.
+//
+// Link expiry is rechecked after each backoff: a long retry chain may outlive
+// a presigned URL, and continuing past expiry is guaranteed to fail.
 func fetchBatchBytes(
 	ctx context.Context,
 	link *cli_service.TSparkArrowResultLink,
@@ -383,7 +394,7 @@ func fetchBatchBytes(
 	retryMax int,
 	retryWaitMin time.Duration,
 	retryWaitMax time.Duration,
-) (io.ReadCloser, error) {
+) (io.Reader, error) {
 	if retryMax < 0 {
 		retryMax = 0
 	}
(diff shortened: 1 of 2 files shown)


internal/rows/arrowbased/batchloader.go
```diff
@@ -339,43 +360,174 @@ func logCloudFetchSpeed(fullURL string, contentLength int64, duration time.Durat
...
+	}
+	base := time.Duration(expo)
+	if base <= 0 {
+		return 0
+	}
+	half := base / 2
+	if half <= 0 {
+		return base
+	}
+	return half + time.Duration(rand.Int63n(int64(half))) //nolint:gosec // G404: jitter only, non-cryptographic
```
gopalldb, line 544: This uses the global rand. Concurrent calls across MaxDownloadThreads goroutines will contend on the global rand mutex. Probably      
  immaterial at typical MaxDownloadThreads values, but for many-thousands-concurrent-downloads workloads (which is the stated motivation), worth swapping in a per-goroutine rand.Rand or math/rand/v2's Uint64N.

internal/rows/arrowbased/batchloader.go
```diff
@@ -339,43 +360,174 @@ func logCloudFetchSpeed(fullURL string, contentLength int64, duration time.Durat
...
+			return nil, dbsqlerrint.NewDriverError(ctx, msg, nil)
+		}
 	}
-	if res.StatusCode != http.StatusOK {
-		msg := fmt.Sprintf("%s: %s %d", errArrowRowsCloudFetchDownloadFailure, "HTTP error", res.StatusCode)
-		return nil, dbsqlerrint.NewDriverError(ctx, msg, err)
+
+	if lastStatus != 0 {
+		msg := fmt.Sprintf("%s: %s %d (after %d retries)", errArrowRowsCloudFetchDownloadFailure, "HTTP error", lastStatus, retryMax)
+		return nil, dbsqlerrint.NewDriverError(ctx, msg, nil)
```
gopalldb: shall we include lastErr here as well instead of nil?

gopalldb (approved):
Added some minor comments

vikrantpuppala:
Force-pushed `b511e86` to rebase onto current main and address review findings. Summary of changes from `4b72cac`:

**Rebase / conflict resolution**
- Rebased onto current `main` (was behind by #357, #358, #360).
- Resolved the #357 conflict in `Run()`: all three result-send sites in the retry-aware `Run()` now go through `cft.sendResult(...)` (the goroutine-leak guard from #357) instead of raw `cft.resultChan <- ...`. Without this, #357's fix for issue #356 silently regressed — and the regression is *worse* than before #357 because the new code holds a fully-buffered `[]byte` in the result instead of a small `io.ReadCloser`.

**Fixes**
- `time.After(wait)` in the backoff `select` → `t := time.NewTimer(wait)` with `t.Stop()` + drain on `ctx.Done()`. Prevents up to `MaxDownloadThreads` lingering timers (each up to `RetryWaitMax`=30s) when a query is cancelled mid-retry.
- Dropped the dead `if res.Header != nil` guard around `res.Header.Get("Retry-After")` — `http.Response.Header` is always non-nil after `httpClient.Do`.
- `writeTruncatedOK` test helper: replaced `t.Fatal` with `t.Errorf + return` — `FailNow` is undefined behavior from a non-test goroutine (it's called from inside an httptest handler).
- Added a code comment at the exhaust-after-retries error path explaining why `lastErr` is `nil` here by construction (resolves @gopalldb's open thread — see below).

**Regression test**
- Added `TestCloudFetchIterator_CloseReleasesAfterRetry`. Companion to `TestCloudFetchIterator_CloseReleasesInFlightDownloads` but exercises the retry path (server flaps 503 → 200) so the goroutine produces its result *after* having gone through `cloudFetchBackoff`. Verified locally:
  - With `sendResult` (this PR): test passes, all goroutines exit after `Close()`.
  - With raw `cft.resultChan <-` (simulating a naive #357 conflict resolution): test fails with 18 leaked `cloudFetchDownloadTask.Run` goroutines.

**Open review threads — responses**
- *Global `rand` mutex contention*: won't fix. At `MaxDownloadThreads=10` default, `rand.Int63n` is sub-microsecond under lock and called only once per backoff (not per request). Per-goroutine `rand.Rand` is non-trivial code for no measurable gain.
- *Pass `lastErr` instead of `nil` on HTTP-status exhaust*: won't fix. The HTTP-status branch explicitly sets `lastErr = nil` on every iteration, so when the loop exits with `lastStatus != 0`, `lastErr` is `nil` by construction. The status code is captured in the error message. Now documented as a code comment.

**Known tradeoffs (worth flagging, not addressed in this PR)**
- LZ4 path now holds `rawBody` (compressed) + `buf` (decompressed) simultaneously during decompression. Necessary for retry-on-truncated-body to work (can't replay a consumed stream). ~200-700 MB extra peak heap on the FactSet workload at default concurrency.
- `logCloudFetchSpeed` now reports actual `len(buf)` instead of `res.ContentLength` — strictly more accurate, but emits speed log lines for chunked responses that previously produced none.
- `downloadMs` telemetry now silently includes backoff time when retries fire. Future follow-up: separate `cloudFetchRetryCount` dimension.

**Deferred follow-ups** (not blocking)
- `Retry-After: 0` jitter floor (hot-loop guard if a misbehaving server returns 0).
- `Retry-After` integer overflow guard.

Co-authored-by: Isaac

internal/rows/arrowbased/batchloader.go
vikrantpuppala: Won't fix — at `MaxDownloadThreads=10` (default) the global `rand` mutex is held for sub-microsecond `Int63n` draws, and only once per backoff event (not per request). Per-goroutine `rand.Rand` is non-trivial code for no measurable gain. Happy to revisit if a future config tunes `MaxDownloadThreads` significantly higher.

internal/rows/arrowbased/batchloader.go
vikrantpuppala: Won't fix — `lastErr` is `nil` by construction at this point. The HTTP-status branch (line ~475) sets `lastErr = nil` on every iteration whenever a response is received, so when the loop exits with `lastStatus != 0`, there's no underlying error to wrap. The status code is already captured in the message. Added a code comment near this branch documenting the invariant so a future reader doesn't "fix" it by passing `lastErr`.

gopalldb (approved):

PR #364, opened 2026-05-25
Telemetry: normalize host key for per-host client + breaker registries

samikshya-db:
## Summary
- `clientManager` and `circuitBreakerManager` keyed their maps by the raw host string, so DSN variants like `example.com`, `example.com/`, and `https://example.com` created separate telemetry clients and separate circuit breakers for the same logical host. That fragments trip state and defeats the per-host consolidation the managers are designed to do.
- Add `normalizeHostKey` (lowercase, trim whitespace, strip `http(s)://`, trim trailing slashes) and use it for all registry lookups in `clientManager.getOrCreateClient`, `clientManager.releaseClient`, and `circuitBreakerManager.getCircuitBreaker`.
- The host string flowing into URL construction is unchanged — first-caller-wins, matching the existing user-agent semantics documented on `getOrCreateClient`.

Addresses Gopal's review comment on #354: https://github.com/databricks/databricks-sql-go/pull/354#discussion_r2546... (host normalization on `telemetry/exporter.go:62`).

## Test plan
- [x] `TestNormalizeHostKey` — covers scheme stripping, lowercase, trim trailing slash, whitespace, empty string
- [x] `TestClientManager_HostVariantsShareClient` — four DSN variants of the same host share a single telemetry client; release via a different variant still cleans up
- [x] `TestCircuitBreakerManager_HostVariantsShareBreaker` — same coverage for the breaker registry
- [x] Full `go test ./telemetry/...` passes

### DIFF
--- a/telemetry/circuitbreaker.go
+++ b/telemetry/circuitbreaker.go
@@ -332,9 +332,15 @@ func getCircuitBreakerManager() *circuitBreakerManager {
 
 // getCircuitBreaker gets or creates a circuit breaker for the host.
 // Thread-safe for concurrent access.
+//
+// The host is normalized (scheme stripped, lowercased, trailing slash trimmed)
+// so DSN variants ("example.com", "example.com/", "https://example.com")
+// share a single breaker per logical host instead of fragmenting trip state.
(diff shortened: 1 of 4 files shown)


jprakash-db (approved):
🚢

PR #370, opened 2026-06-02
fix: lossless decimal decode (fixes float64 precision loss) + reachable native-decimal opt-in

vikrantpuppala:
## What this fixes

The driver has a **live precision bug on the default path**, plus a stranded, unreachable native-decimal code path. This PR fixes both.

**The bug (reachable today, no opt-in required):** complex types are native by default (`UseArrowNativeComplexTypes = true`), so any `DECIMAL` nested inside an `ARRAY`/`STRUCT`/`MAP` is decoded by `decimal128Container`, which used `Decimal128.ToFloat64`. `float64` cannot represent high-precision/high-scale decimals, so values are silently corrupted. Reproduced directly against current `main`:

| Exact value from server | `main` decodes as | |
|---|---|---|
| `12345678901234567890.123456789012345678` | `1.234567890123457e+19` | corrupted |
| `99999999999999999999.999999999999999999` | `1e+20` | collapsed to a round number |
| `3.30` | `3.3000000000000003` | wrong even for a trivial value |
| SQL `NULL` (top-level scan) | `float64(0)` | null becomes a real zero |

**The stranded code:** `UseArrowNativeDecimal` exists as a config field with full wire plumbing (`connection.go` sends `DecimalAsArrow`) and a decode container — but **no DSN key or connector option ever sets it**, so no user can reach the top-level native-decimal path at all. It is dead code.

This is also a corrected take on #282, which tried to enable native decimal by default but kept the lossy `float64` decode (and offered no opt-out). The review of #282 flagged exactly the issues addressed here.

## What this PR does

- **Lossless decode** (`internal/rows/arrowbased/columnValues.go`): `decimal128Container.Value` returns `Decimal128.ToString(scale)` — a scale-normalized, exact string — instead of the lossy `ToFloat64`. It is also null-safe (returns `nil` for null cells, not `0`).
- **Makes the path reachable** (`connector.go`): new public `WithUseArrowNativeDecimal(enable bool) ConnOption`. Previously `ArrowConfig.UseArrowNativeDecimal` lived in `internal/config` with no entry point.
- **Unblocks top-level decimal in `ScanRow`** (`internal/rows/arrowbased/arrowRows.go`): decimals are no longer rejected as "unsupported native type"; interval types remain blocked.
- **Consistent Go type**: returning a string matches both the non-native default path and the column-based (non-Arrow) transport, so a `DECIMAL` column comes back as the same Go type regardless of which transport the server picks. (Previously the native path returned `float64` while the column path returned `string` for the same column.)
- **Default unchanged, non-breaking**: `ArrowConfig.WithDefaults()` still leaves `UseArrowNativeDecimal = false`. For top-level decimal columns the observable Go type is `string` either way; opting in only changes the wire transport. The user-visible behavior fix is for the nested-in-complex-type case, which now decodes losslessly.
- **Docs** (`doc.go`): `DECIMAL(p,s) --> string` (corrects the inaccurate `sql.RawBytes`).
- **CHANGELOG**: entry under `Unreleased`.

## Why string, not a numeric type

`database/sql`'s `driver.Value` has no decimal type — the allowed kinds are `int64`, `float64`, `bool`, `[]byte`, `string`, `time.Time`. Returning `string` is the standard idiom other Go SQL drivers use for `DECIMAL`/`NUMERIC` (e.g. `lib/pq`, `pgx`), because it is exact and lets the caller decode into `shopspring/decimal`, `*big.Rat`, etc. without precision already lost. It is also what this driver already returned for decimals, so it is the non-breaking choice.

## Tests

- **`TestDecimal128ContainerValue`** — asserts exact lossless decode of a `DECIMAL(38,18)` high-precision value, negatives, scale normalization, and `NULL -> nil`. Verified to **fail** against `main`'s `ToFloat64` decode and **pass** with the fix.
- **`TestDecimal128NestedInListPreservesPrecision`** — guards the default path (decimal nested in `ARRAY`). Verified to **fail** against `main` (renders `[12345678901234570000,null,3.3000000000000003]`) and **pass** with the fix.
- Existing `ScanRow` boundary test updated to reflect decimal now succeeding while intervals still error.

`go build ./...`, `go vet ./...`, and `go test ./...` are all clean.

## Behavior change to review deliberately

Nested decimals inside native complex types previously serialized as (lossy) JSON numbers, e.g. `[5.15,123.45,null]`. They now serialize as lossless JSON strings, e.g. `["5.15","123.45",null]`. This is a representation change for direct consumers of native complex-type values, in exchange for exact precision. The relevant existing test expectation is updated accordingly.

## Resolution of the #282 review findings

| Finding on #282 | Resolution here |
|---|---|
| Lossy `float64` (default-path bug) | Lossless `string` via `ToString(scale)` |
| SQL `NULL` decoded as `0` | Null-safe `Value()` returns `nil` |
| No value-level test for the native path | `TestDecimal128ContainerValue` + nested-list test |
| `doc.go` stale (`sql.RawBytes`) | Updated to `string` |
| No public opt-out / path unreachable | `WithUseArrowNativeDecimal` connector option |
| Type confusion (Arrow `float64` vs column `string`) | Both transports return `string` |
| No changelog entry | Added under `Unreleased` |

Refs #282.

### DIFF
--- a/internal/rows/arrowbased/columnValues.go
+++ b/internal/rows/arrowbased/columnValues.go
@@ -521,9 +521,17 @@ type decimal128Container struct {
 var _ columnValues = (*decimal128Container)(nil)
 
 func (tvc *decimal128Container) Value(i int) (any, error) {
+	if tvc.decimalArray.IsNull(i) {
+		return nil, nil
+	}
 	dv := tvc.decimalArray.Value(i)
-	fv := dv.ToFloat64(tvc.scale)
-	return fv, nil
+	// Return the decimal as a lossless string. float64 cannot exactly
+	// represent high-precision/high-scale decimals, so converting here would
+	// silently lose precision for the very type users reach for to avoid that.
+	// Returning a string also matches the non-native default path and the
+	// column-based transport, so the Go type is consistent regardless of how
+	// the server chooses to send the result.
+	return dv.ToString(tvc.scale), nil
 }
 
 func (tvc *decimal128Container) IsNull(i int) bool {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,5 @@
 # Release History
 
-## Unreleased
-- Add native Arrow decimal support, opt-in via `WithUseArrowNativeDecimal(true)`. DECIMAL columns are decoded from native Arrow decimal128 values as a lossless decimal string, preserving full precision and scale. The default remains unchanged (decimals are still returned as strings), so this is a non-breaking, opt-in change.
-
 ## v1.12.0 (2026-05-25)
 - Retry transient S3 errors in CloudFetch downloads and staging PUT/GET/REMOVE operations (databricks/databricks-sql-go#355, #361)
 - Telemetry: normalize host key for per-host client + breaker registries; stop retrying into 429s, honour Retry-After, fix userAgent (databricks/databricks-sql-go#354, #364)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release History
 
+## Unreleased
+- Add native Arrow decimal support, opt-in via `WithUseArrowNativeDecimal(true)`. DECIMAL columns are decoded from native Arrow decimal128 values as a lossless decimal string, preserving full precision and scale. The default remains unchanged (decimals are still returned as strings), so this is a non-breaking, opt-in change.
+
 ## v1.12.0 (2026-05-25)
 - Retry transient S3 errors in CloudFetch downloads and staging PUT/GET/REMOVE operations (databricks/databricks-sql-go#355, #361)
 - Telemetry: normalize host key for per-host client + breaker registries; stop retrying into 429s, honour Retry-After, fix userAgent (databricks/databricks-sql-go#354, #364)
--- a/connector.go
+++ b/connector.go
@@ -409,6 +409,18 @@ func WithCloudFetch(useCloudFetch bool) ConnOption {
 	}
 }
 
+// WithUseArrowNativeDecimal enables native Arrow decimal support. Default is false.
+// When enabled, DECIMAL columns retrieved over Arrow batches are decoded as
+// native Arrow decimal128 values and returned as a lossless decimal string
+// (e.g. "123.45"), preserving full precision and scale. When disabled (the
+// default), decimals are returned as strings via the non-native path, so the
+// observable Go type is the same either way.
+func WithUseArrowNativeDecimal(enable bool) ConnOption {
+	return func(c *config.Config) {
+		c.UseArrowNativeDecimal = enable
+	}
+}
+
 // WithMaxDownloadThreads sets up maximum download threads for cloud fetch. Default is 10.
 func WithMaxDownloadThreads(numThreads int) ConnOption {
 	return func(c *config.Config) {
--- a/doc.go
+++ b/doc.go
@@ -363,7 +363,7 @@ DATE --> time.Time
 
 TIMESTAMP --> time.Time
 
-DECIMAL(p,s) --> sql.RawBytes
+DECIMAL(p,s) --> string
 
 BINARY --> sql.RawBytes
 
--- a/internal/rows/arrowbased/arrowRows.go
+++ b/internal/rows/arrowbased/arrowRows.go
@@ -218,10 +218,11 @@ func (ars *arrowRowScanner) ScanRow(
 			col := ars.colInfo[i]
 			dbType := col.dbType
 
-			if (dbType == cli_service.TTypeId_DECIMAL_TYPE && ars.UseArrowNativeDecimal) ||
-				(isIntervalType(dbType) && ars.UseArrowNativeIntervalTypes) {
-				//	not yet fully supported
-				ars.Error().Msgf(errArrowRowsUnsupportedNativeType(dbType.String()))
+			// Decimal types are supported natively (returned as a lossless string
+			// via decimal128Container). Only interval types remain unsupported.
+			if isIntervalType(dbType) && ars.UseArrowNativeIntervalTypes {
+				//	interval types are not yet fully supported
+				ars.Error().Msg(errArrowRowsUnsupportedNativeType(dbType.String()))
 				return dbsqlerrint.NewDriverError(ars.ctx, errArrowRowsUnsupportedNativeType(dbType.String()), nil)
 			}
 
--- a/internal/rows/arrowbased/arrowRows_test.go
+++ b/internal/rows/arrowbased/arrowRows_test.go
@@ -887,7 +887,10 @@ func TestArrowRowScanner(t *testing.T) {
 
 			err := ars.ScanRow(dest, 0)
 
-			if i < 3 {
+			// Columns are ordered: 0=array, 1=map, 2=struct, 3=decimal,
+			// 4=interval_ym, 5=interval_dt. Complex types (0-2) and decimal (3)
+			// are supported natively; only the interval types (4-5) still error.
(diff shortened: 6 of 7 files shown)


internal/rows/arrowbased/columnValues.go
```diff
@@ -521,9 +521,17 @@ type decimal128Container struct {
...
 	dv := tvc.decimalArray.Value(i)
-	fv := dv.ToFloat64(tvc.scale)
-	return fv, nil
+	// Return the decimal as a lossless string. float64 cannot exactly
+	// represent high-precision/high-scale decimals, so converting here would
+	// silently lose precision for the very type users reach for to avoid that.
+	// Returning a string also matches the non-native default path and the
+	// column-based transport, so the Go type is consistent regardless of how
+	// the server chooses to send the result.
+	return dv.ToString(tvc.scale), nil
```
gopalldb, line 534: but this is changing the return type

vikrantpuppala: Good catch — yes, `decimal128Container.Value` changes its return from `float64` to `string`. Worth being precise about who can observe that:

- **Top-level decimal columns**: the native path that produces this value was **unreachable** before this PR — `UseArrowNativeDecimal` had no DSN key or connector option, so no user could turn it on. So the old `float64` return was never observable; there's no behavior to break. This PR makes the path reachable (via `WithUseArrowNativeDecimal`) and lands it on `string`, which is what the existing non-native default path and the column-based path already return for decimals — so the type is now *consistent* across transports rather than newly divergent.

- **Decimals nested in complex types** (array/struct): this container *is* reached today, because complex types are native by default. Here the type does change (JSON number → JSON string in the marshaled value). But the old `float64` value was **corrupt** — e.g. `12345678901234567890.123456789012345678` decoded as `1.234567890123457e+19`, and SQL `NULL` decoded as `0`. So this is fixing silently-wrong output, not regressing correct output. `TestDecimal128NestedInListPreservesPrecision` pins this and was verified to fail against the old `ToFloat64` decode.

On *why* `string` rather than a numeric type: `database/sql`'s `driver.Value` has no decimal type (only `int64`/`float64`/`bool`/`[]byte`/`string`/`time.Time`), so a lossless decimal has to come back as a string for the caller to decode into `shopspring/decimal`, `*big.Rat`, etc. It's the same choice `lib/pq`/`pgx` make for `NUMERIC`.

Happy to add a typed wrapper instead if you'd prefer that direction — but it'd be a new public type and a larger surface, so I kept this PR to the lossless-string fix. Let me know.

gopalldb:
Risk level: HIGH · Recommendation: 🔴 REQUEST CHANGES

  This switches Arrow decimal decoding from dv.ToFloat64(scale) to dv.ToString(scale) to fix float64 precision loss, adds a WithUseArrowNativeDecimal opt-in, and fixes a null=0 bug. The intent is right — but
  the chosen fix is itself lossy at high scale, which I verified by running the actual pinned library.

  🔴 P0 (Critical) — verified against the real library, must fix

  decimal128.ToString in the pinned arrow v12.0.1 silently corrupts decimals at scale ≳ 25 — the fix defeats its own "lossless" premise.
  internal/rows/arrowbased/columnValues.go:534 (return dv.ToString(tvc.scale), nil)

  │           Input           │     Column     │               ToString output               │
  ├───────────────────────────┼────────────────┼─────────────────────────────────────────────┤
  │ 1                         │ DECIMAL(38,38) │ 0.99999999999999997748809823456034029568 ❌ │
  ├───────────────────────────┼────────────────┼─────────────────────────────────────────────┤
  │ 5                         │ DECIMAL(38,35) │ 4.99999999999999984316830539589877759 ❌    │
  ├───────────────────────────┼────────────────┼─────────────────────────────────────────────┤
  │ 0.1                       │ DECIMAL(38,25) │ 0.1000000000000000090596966 ❌              │
  ├───────────────────────────┼────────────────┼─────────────────────────────────────────────┤
  │ 3567933                   │ DECIMAL(38,30) │ 3567933.000000000070947009154460418048 ❌   │
  ├───────────────────────────┼────────────────┼─────────────────────────────────────────────┤
  │ high-precision @ scale 18 │ DECIMAL(38,18) │ exact ✓                                     │
  └───────────────────────────┴────────────────┴─────────────────────────────────────────────┘

  Root cause (confirmed in v12.0.1 source): ToString does f := (&big.Float{}).SetInt(n.BigInt()); f.Quo(f, ...10^scale); f.Text('f', scale) — no SetPrec. big.Float.SetInt sets precision to just the mantissa's
  bit-width, so dividing by 10^scale injects binary rounding error that surfaces at high scale. (Arrow v18 fixed this by adding .SetPrec(128); v12 has no such fix.)

  Why this is worse than the bug it replaces: the old float64 path was visibly approximate; this produces a string that looks exact but is wrong. Databricks DECIMAL supports scale up to 38, and this hits both
  the new opt-in top-level path and the un-gated default nested-decimal path (see P1) — so even users who never set the flag get corrupted nested decimals. The PR's unit tests only use scale 18, the one scale
  that happens to round-trip correctly, so CI stays green.
  → Fix: don't use v12's ToString. Either format from decimal128.Num.BigInt() with manual decimal-point placement (pure big.Int/string, no float), or bump arrow to a version whose ToString uses SetPrec(128).
  Add a DECIMAL(38,38) / DECIMAL(38,30) regression test — the current scale-18-only tests would not catch this.

  🟠 P1 (Important)

  1. Default-path nested-decimal behavior change is NOT gated by the opt-in flag (compat break) — columnValues.go:523-535. UseArrowNativeComplexTypes defaults true, and makeColumnValueContainer decodes a
  decimal nested in ARRAY/MAP/STRUCT via decimal128Container regardless of UseArrowNativeDecimal. So this PR changes nested-decimal JSON from 5.15 (number) to "5.15" (string) for all existing users — exactly
  the arrowRows_test.go expectation change [5.15,...] → ["5.15",...]. This contradicts WithUseArrowNativeDecimal's own doc ("When disabled (the default)... non-native path"). → document it as an intentional
  breaking change in release notes and fix the misleading flag doc. (At high scale this changed output is also wrong per P0.)
  2. Nested-decimal tests hardcode scale, weakening the safety net — columnValues_test.go:170,214. Production wiring is currently correct (makeColumnValueContainer sources scale: t.Scale recursively, and
  SetValueArray does not set scale), so no scale-0 corruption today — but a future regression dropping scale: t.Scale would mis-place the decimal point (data corruption) and these tests would still pass. → add
  one test driving a nested decimal through the real makeColumnValueContainer path.

  🟡 P2 (Minor — all correct)

  - Msgf → Msg (arrowRows.go:223) — correct; errArrowRowsUnsupportedNativeType returns an already-Sprintf'd string, so the old Msgf would treat % in a type name as a format verb.
  - IsNull → (nil,nil) (columnValues.go:524-526) — correct; resolves the old null→float64(0) bug on the scalar native path; nil any scans cleanly into *string/*[]byte/*sql.RawBytes/*interface{}.
  - doc.go sql.RawBytes → string — a correction, not a new change; top-level default decimals already map to arrow.BinaryTypes.String and return string. No top-level type break.

  ---
  Bottom line: REQUEST CHANGES. Two suspected bugs cleared (top-level type unchanged; nested scale wiring correct), and the null fix + doc correction are right. But the headline ToString "lossless" claim is
  false on the pinned arrow v12.0.1 — I reproduced silent corruption at scale ≥ ~25 against the real library, affecting both the opt-in and the un-gated default nested path. Do not merge until decimal
  formatting is precision-safe at scale 38 (manual big.Int formatting or an arrow bump with SetPrec(128)) with a high-scale regression test, and the nested-decimal default-path change is documented.

vikrantpuppala:
fixed in #383

PR #381, opened 2026-07-02
ci: trigger Go integration tests in databricks-driver-test

mani-mathur-arch:
This adds a workflow that runs the driver against the proxy-based Go integration
test suite hosted in `databricks/databricks-driver-test`, and reports the outcome
back onto the pull request as the `Go Integration Tests` check.

## Why

The integration suite exercises the driver end to end through a recording/replay
proxy. It lives in a separate repository because it needs shared test fixtures and
a warehouse-backed environment that do not belong in the driver repo. This workflow
is the bridge: it forwards a pull request (or merge-queue commit) to that repository
via `repository_dispatch`, and the suite posts its result back as a status check.

The suite validates the Thrift backend that the driver uses for query execution, so
almost any change in this repository can affect it. For that reason the workflow does
not path-filter — it applies to every change.

## How it runs

There are two execution modes:

- **Replay** — the suite serves pre-recorded proxy responses. It is deterministic,
  needs no warehouse or credentials, and runs in seconds. This is the mode used for
  the pre-merge gate.
- **Passthrough** — the suite runs the full set of tests, including ones that cannot
  be recorded (cloud fetch, concurrency, retry/error injection, authentication),
  against a live warehouse. This is slower and is run on demand.

The workflow wires those modes to pull-request events as follows:

- **On an ordinary pull request**, it posts a passing `Go Integration Tests` check
  immediately without running the suite. This keeps the required check satisfied on
  the pull request itself; the real run happens in the merge queue (below). It avoids
  re-running the suite on every push while a pull request is still in progress.
- **When a maintainer applies the `integration-test` label**, it runs the replay
  suite against the pull request commit as a preview. Applying `integration-test-full`
  instead runs the full passthrough suite. Both labels are removed automatically when
  new commits are pushed, so a maintainer must re-review before re-running.
- **In the merge queue** (`merge_group` event), it runs the replay suite as the real
  required gate, once, immediately before the commit merges.

Only a maintainer with write access can apply labels, and the merge-queue run only
sees code that has already been approved for merging. The workflow itself never checks
out or executes pull-request code — it only reads pull-request metadata and calls the
dispatch API — so it uses the ordinary `pull_request` event.

## Required setup

This workflow is inactive until the following are in place, and has no effect on the
repository before then:

1. A GitHub App for cross-repository dispatch, with its `INTEGRATION_TEST_APP_ID` and
   `INTEGRATION_TEST_PRIVATE_KEY` stored as repository secrets here.
2. That App granted access to `databricks-driver-test` (to send `repository_dispatch`)
   and to this repository (`checks:write`, so the suite can report the result back).
3. The `integration-test` and `integration-test-full` labels created in this repository.
4. The merge queue enabled on this repository, with `Go Integration Tests` listed as a
   required status check. Until the merge queue is enabled, the merge-queue job does not
   run; the pull-request check and the label-triggered previews work regardless.

### DIFF
--- a/.github/workflows/trigger-integration-tests.yml
+++ b/.github/workflows/trigger-integration-tests.yml
@@ -36,7 +36,7 @@ jobs:
       group: databricks-protected-runner-group
       labels: linux-ubuntu-latest
     permissions:
-      pull-requests: write
+      # Labels + comments are the issues API; no pulls.* calls here.
       issues: write
     steps:
       - name: Remove integration-test labels
@@ -125,8 +125,8 @@ jobs:
       group: databricks-protected-runner-group
       labels: linux-ubuntu-latest
     permissions:
+      # Comments are the issues API; the dispatch uses the App token, not GITHUB_TOKEN.
       issues: write
-      pull-requests: write
       checks: write
     steps:
       - name: Resolve proxy mode from label
@@ -199,7 +199,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.issue.number,
-              body: `Go integration tests triggered (\`${process.env.PROXY_MODE}\`). [View workflow run](https://github.com/databricks/databricks-driver-test/actions/workflows/databricks-go-integration-tests.yml).`,
+              body: `Go integration tests triggered (\`${process.env.PROXY_MODE}\`). [View workflow runs](https://github.com/databricks/databricks-driver-test/actions).`,
             });
 
   # Merge queue: the required gate. Runs replay once before merge.
--- a/.github/workflows/trigger-integration-tests.yml
+++ b/.github/workflows/trigger-integration-tests.yml
@@ -1,0 +1,283 @@
+name: Trigger Integration Tests
+
+# Dispatches the proxy-based Go integration suite in
+# databricks/databricks-driver-test to run against this PR's commit, and reports
+# the result back as the "Go Integration Tests" check.
+#
+# Matches the label-gated / merge-queue pattern used by the Node.js and Python
+# connectors: normal PR events get an immediate green check, maintainers can
+# preview with the `integration-test` label (replay) or `integration-test-full`
+# (full passthrough), and merge queue runs the real required gate.
+#
+# Required external setup:
+#
+#   1. `integration-test` and `integration-test-full` labels exist in this repo.
+#   2. `INTEGRATION_TEST_APP_ID` / `INTEGRATION_TEST_PRIVATE_KEY` repo secrets
+#      are installed in this repo for the dispatcher GitHub App.
+#   3. The app is installed/granted on `databricks-driver-test` so this workflow
+#      can send `repository_dispatch`.
+#   4. The same app is installed/granted on `databricks-sql-go` with checks:write
+#      so driver-test can report the final `Go Integration Tests` check back to
+#      this PR/merge-queue commit.
+#   5. Merge queue branch protection lists `Go Integration Tests` as a required
+#      status check.
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


gopalldb:
💡 [MINOR] Null dereference of pr.head.repo when source fork is deleted
Line 65 | bug | logical_claude_agent

Null dereference on deleted-fork PRs. Line 65 computes pr.head.repo.full_name, but GitHub populates head.repo as null when the PR's source fork has been deleted. This code runs in remove-label-on-new-commit on a synchronize event whenever a trigger label was present (removed.length > 0, line 63). For a fork PR whose fork was deleted while the label was still applied, pr.head.repo is null and the .full_name access throws a TypeError, failing the label-reset step (and skipping the reset comment).

---
⚠️ [MAJOR] Merge-queue PR-number regex takes leftmost pr-<n> match, can extract wrong PR
Line 218 | bug | logical_claude_agent

Unanchored pr-([0-9]+) regex extracts the wrong PR number from the merge-queue ref. The merge_group.head_ref has the form refs/heads/gh-readonly-queue/<base_branch>/pr-<n>-<sha>. Bash [[ =~ ]] returns the leftmost match, so pr-([0-9]+) (line 218) captures the first pr-<digits> in the whole ref. If the base branch name itself contains a pr-<digits> segment (e.g. feature/pr-123-fix), the extraction yields that segment's number instead of the real PR number — verified: matching against refs/heads/gh-readonly-queue/feature/pr-123-fix/pr-456-abc captures 123, not 456. The wrong number then flows into the driver-test dispatch (pr_number/pr_url, lines 248 & 251), causing the required merge-queue gate to be reported against an unrelated PR.

mani-mathur-arch:
Thanks @gopalldb, Both the fixes are now in place.

**[MAJOR] Merge-queue PR-number regex takes leftmost pr- match, can extract wrong PR.** 
You're right — the unanchored `pr-([0-9]+)` would capture the wrong number if a base branch name contained a
  `pr-<digits>` segment (e.g. `.../feature/pr-123-fix/pr-456-<sha>` → `123`). Anchored it to the trailing queue segment:
  
```bash
  if [[ "$MERGE_QUEUE_REF" =~ /pr-([0-9]+)-[0-9a-f]+$ ]]; then
```
The leading / binds pr- to a path segment and -[0-9a-f]+$ pins to the final pr-<n>-<sha>, so any interior pr-N is followed by a / and can't reach the anchor — that case now correctly yields 456. Verified under bash against real gh-readonly-queue/<base>/pr-<n>-<sha> refs; fails closed (exit 1 → red check) on a non-matching ref. Worth noting the sibling workflows carry the same unanchored form — databricks-sql-nodejs trigger-integration-tests.yml:242 and databricks-sql-python:378 — so they may want the same fix.

**[MINOR] Null dereference of pr.head.repo on deleted-fork PRs (.github/workflows/trigger-integration-tests.yml:67).** Fixed with optional chaining:
 
```js
  const isFromFork = pr.head.repo?.full_name !== pr.base.repo.full_name;
```

When the source fork is deleted pr.head.repo is null, so this now evaluates undefined !== <base> → treated as a fork, no TypeError on the synchronize label-reset path. Same note here — databricks-sql-nodejs:75 and databricks-sql-python:100 still use the unguarded pr.head.repo.full_name.
  
On the overall shape: the green-stub-on-PR + real-gate-in-merge_group model here mirrors the databricks-sql-nodejs and databricks-sql-python trigger-integration-tests.yml workflows — same label-gating (integration-test / integration-test-full), same merge-queue-scoped required check, same pull_request event. This PR just hardens the two spots above relative to those siblings.

msrathore-db (approved):

gopalldb (approved):

PR #382, opened 2026-07-05
refactor: extract Backend/Operation seam for SEA integration

mani-mathur-arch:
## Summary

Introduce a pluggable execution-backend abstraction (`internal/backend`) and fold today's Thrift/HiveServer2 code into it as the sole implementation (`internal/backend/thrift`). `conn` no longer talks to the generated Thrift client directly — it holds one `backend.Backend` and drives session lifecycle and statement execution through the neutral `Backend` / `Operation` interfaces. **No behavior change, no cgo, no new user-facing API.**

This is Layer 1 of adding SEA (Statement Execution API) support: it draws the seam so a second, SEA-via-kernel backend can slot in behind a connect-time flag later, without touching `conn`, `stmt`, or any public API. Row decoding is **not** part of this neutral seam: `internal/rows` is coupled to Thrift result types today, so each backend supplies `driver.Rows` through its own construction path. A follow-up PR will decouple the rows layer before or with Layer 2.

## Why

The driver is migrating to the SEA protocol, which we plan to add via the Rust kernel. Wiring a second protocol into `conn` as it stands would mean threading Thrift and SEA types through every `database/sql` method. Drawing the seam first — with Thrift as the only implementation — keeps that future change to a single factory branch and makes *this* PR a pure, verifiable refactor. It is the regression firewall the kernel PRs build on.

## What changes

- **`internal/backend`** — neutral `Backend` / `Operation` interfaces + `ExecRequest` / `Param` request types. Only a backend implementation imports its protocol's client (the Thrift backend imports `internal/cli_service`); nothing else in the driver does for the execute/poll/close/session path.
- **`internal/backend/thrift`** — the Thrift execute/poll/close/session logic moved here behind the interfaces; `thriftOperation` wraps the execute + status responses and exposes only neutral accessors. The existing Thrift-mechanism tests moved with it.
- **`conn` / `connector`** — drop the `client` + `session` fields for a single `backend` field; the 9 `database/sql` methods keep their signatures and delegate. The connector factory builds the Thrift backend (a future flag will select the kernel backend at that same point).
- **`parameters.go`** — split the parameter conversion at a neutral `backend.Param` boundary (type inference stays in `dbsql`, wire-mapping moves to the backend) to avoid a `dbsql`↔`backend` import cycle. `internal/querytags` extracted for the same reason; the public `dbsql.SerializeQueryTags` forwards to it, unchanged.
- **`internal/debuglog`** — a step-level debug tracer for the upcoming FFI work. It is a thin helper over the driver's existing logger, not a second logger: every step is an ordinary `level=debug` line on the same sink, in the same JSON/pretty format as every other driver log, so Go and (later) kernel-side stderr logs read as one interleaved stream in execution order. Turn it on the way you turn on any driver logging — `SetLogLevel("debug")` or `DATABRICKS_LOG_LEVEL=debug` (off by default; near-zero cost when off). Each step carries its function name, an `enter`/`done` phase (the `enter` is emitted *before* the step runs, so a step that hangs at the FFI edge is still visible), the elapsed time on `done`, and the conn/correlation/query ids from the context.

The data plane (`internal/rows` / `rowscanner`) was already backend-agnostic and is intentionally untouched.

## Test plan

- [x] `go test ./... -count=1` passes — unit tests + the moved Thrift-mechanism tests
- [x] Real-warehouse e2e (`TestE2EArrowBatchesSurviveQueryContextCancellation`, `TestE2ECloudFetchExactRowCount`) passes against a staging warehouse
- [x] `CGO_ENABLED=0 go build ./...` still works — the pure-Go default is preserved
- [x] `go vet ./...` and `gofmt -s` clean; `golangci-lint` clean; `internal/debuglog` and `internal/backend/thrift` race-clean under `-race`
- [x] Reviewed with `isaac review` and the code-review squad; all findings fixed (see below)

## Review notes

Reviewed with `isaac review` (clean) and a 9-reviewer squad. All findings that touched behavior or coverage were fixed in this PR:

- Cached the session/statement id — the refactor was recomputing `SprintGuid` per call vs. the old cached `conn.id`.
- Added coverage for the operation-close skip branches (a subtest dropped in the test move).
- Collapsed a redundant duplicate result method (`Results`/`StagingResults` were byte-identical).
- Threaded the connect context into the client-init error (`thrift.New`) so the correlation id isn't lost.
- De-duplicated a doubly-wrapped staging-metadata error.
- **Restored the `queryId` enrichment semantics** (commit 2): the refactor had changed the execute path from *fill-if-empty* to *overwrite-always* and stopped firing the `QueryIdCallback` on the no-handle path. Both are now preserved and pinned by `TestEnrichQueryId`.
- **`internal/debuglog` emits through the existing logger, not a second one.** An earlier revision added a trace-level sibling logger gated by its own `DBSQL_DEBUG_LOG` env var; @gopalldb flagged a data race on that sibling (`SetLogOutput` swapped it while query goroutines read it). Rather than patch it with an `atomic.Pointer`, the step tracer now writes through the single `logger.Logger` at debug level, gated by the existing `SetLogLevel` / `DATABRICKS_LOG_LEVEL`. This removes the second logger and its separate knob entirely, so the race is gone by construction — `logger.go` reverts to its pre-refactor shape — and there is one logger for maintainers to reason about, one format on the wire, and one place to redirect output. Net **−230 lines**.
- **Code-review squad follow-up** (later commit): documented the exec-path vs query-path close-ownership split on `Operation` (query path hands teardown to `Rows.Close`), made `thriftOperation.Close` truly idempotent via a `closed` memo, restored the end-to-end composed-error assertion the move dropped, pinned the "Execute returns a non-nil Operation on error" contract, added `StatementID` / `AffectedRows` accessor tests, collapsed a byte-identical duplicate test constructor, and rewrote a handful of comments that narrated change history instead of stating the contract.

**Deferred to a follow-up (no behavior change — internal only):** the parameter conversion makes an extra intermediate pass (`[]backend.Param`) that the old single-pass build didn't. The output is byte-identical (proven by the shared param tests); it's a few extra allocations per parameterized query, inherent to the neutral seam, and fusing it would re-entangle the `dbsql`/`backend` split. Can be optimized later if it ever matters.

Also unrelated to this PR: the test named \`"ExecContext currently does not support query parameters"\` has been stale since parameters landed in 2023 (worth a cleanup ticket).


gopalldb:
⚠️ [MAJOR] · bug · logger/logger.go:98

Data race on debugLogger between SetLogOutput and debuglog step tracing

Non-atomic reassignment of the global debugLogger races with concurrent step tracing. The newly added debugLogger (logger.go:50) is a plain zerolog.Logger var, and DebugLogger() (logger.go:59) returns &debugLogger — a pointer to the mutable global. The public SetLogOutput overwrites the entire struct with a plain store at logger.go:98 (debugLogger = newDebugLogger(w)). Meanwhile, whenever debug tracing is enabled (DBSQL_DEBUG_LOG / SetEnabled), debuglog.event() reads that same struct through the pointer on every step (debuglog.go:143, logger.DebugLogger().Log()). A caller invoking the public SetLogOutput while queries are emitting debug traces is an unsynchronized read/write of a multi-field struct — a data race that -race will flag and that can surface a torn logger value in a long-running process. Notably, debuglog's own sibling globals (enabled, seq, clockOverride) are all atomic precisely because they are touched concurrently with logging; debugLogger was left unprotected.

mani-mathur-arch:
Thanks @gopalldb — fixed in e24e903.

  **[MAJOR] Data race on `debugLogger` between `SetLogOutput` and step tracing.** You're right — the sibling gates (`enabled`, `seq`, `clockOverride`) were made atomic precisely because they're touched concurrently with logging, and `debugLogger` was left as a plain `zerolog.Logger` var that `SetLogOutput` overwrote wholesale while `debuglog.event()` read it through the `DebugLogger()` pointer on every step. `-race` flags it and a torn multi-field read is possible in a long-running process. Now held in an `atomic.Pointer[zerolog.Logger]`: the three assignment sites go through a small `storeDebugLogger` helper that `Store`s a fresh logger, and `DebugLogger()` returns `debugLogger.Load()` — a stable snapshot, so a concurrent `SetLogOutput` swaps in a new logger without mutating the one already handed out. This matches the atomic treatment the sibling globals already get.
  
Added `TestConcurrentSetLogOutputIsRaceFree`, which drives concurrent `SetLogOutput` against active tracing; I confirmed it reports `DATA RACE` under `-race` on the plain-var version and passes with the fix.
  
  `gofmt` / `go vet` clean; `go build ./...` and the full `internal/debuglog` suite pass under `-race`.

internal/backend/backend.go
```diff
@@ -0,0 +1,109 @@
...
+	// supplies whichever subset it needs (a full query wires all three, the
+	// staging path wires only chunk timing).
+	Results(ctx context.Context, callbacks *dbsqlrows.TelemetryCallbacks) (driver.Rows, error)
+
+	// IsStaging reports whether this is a staging (PUT/GET/REMOVE) operation. It
+	// may issue a metadata lookup if the answer is not already available from
+	// direct results.
+	IsStaging(ctx context.Context) (bool, error)
+
+	// Close best-effort closes the server-side operation, if one is still open.
```
mani-mathur-arch, line 92: The `Operation` doc documents `Close` ("closes the server-side operation") and `Results`, but never states **who closes the server operation on the row-returning path**. In practice there are two path-dependent closers, and the driver relies on both being idempotent:

- **Exec / staging path** (`ExecContext`): closes explicitly via `op.Close(newCtx)` — `connection.go:175`, the *only* `op.Close` call site in the file.
- **Query path** (`QueryContext`): `op.Close` is **never** called; the server operation is closed solely as a side effect of the returned `driver.Rows.Close()` → `ResultPageIterator.Close()` → `CloseOperation`.

An engineer/agent implementing the forthcoming SEA `Operation` from the interface + comments alone (the PR's stated goal) reasonably puts server-statement teardown in `Close`, and implements `Results` to return a streaming `Rows` that does *not* self-close — assuming the framework calls `Operation.Close`. Correctness tests pass, but because `QueryContext` never calls `op.Close`, **every SELECT leaks a server-side statement**. Invisible to correctness tests; surfaces as server-side resource exhaustion under load.

The Thrift impl is correct today — this is a doc-clarity gap that becomes a real bug in Layer 2. Suggested fix: document on `Operation` / `Results` / `Close` that on the query path the caller does not call `Operation.Close`; the `driver.Rows` returned by `Results` owns closing the server operation on its `Close()`, and that `Operation.Close` and `Rows.Close` must each be independently idempotent.

---

internal/backend/thrift/operation.go
```diff
@@ -0,0 +1,137 @@
...
+// status response.
+func (o *thriftOperation) AffectedRows() int64 {
+	return o.opStatusResp.GetNumModifiedRows()
+}
+
+// Results builds the driver.Rows for the operation's result set by delegating to
+// the rows layer with the Thrift handle, client, and direct results. The caller
+// supplies the telemetry callbacks (full for a query, chunk-timing-only for a
+// staging read).
+func (o *thriftOperation) Results(ctx context.Context, callbacks *dbsqlrows.TelemetryCallbacks) (driver.Rows, error) {
```
mani-mathur-arch, line 67: `Operation.Results` returns a neutral `driver.Rows` (the right altitude), but the only way to produce one — `dbsqlrows.NewRows` (`internal/rows/rows.go:110`) — is hardwired to Thrift end-to-end: it requires `*cli_service.TOperationHandle`, `cli_service.TCLIService`, and `*cli_service.TSparkDirectResults`, and the whole decode subsystem (arrow, CloudFetch, pagination, scanners) consumes `cli_service` types.

The PR description says a SEA backend can slot in "without touching … **the rows layer**." That holds for `conn`, `stmt`, and the public API (all cleanly decoupled — verified), but **not** the rows layer: a SEA `Operation.Results` has no `TOperationHandle`/`TCLIService`/`TSparkDirectResults` to pass, so it must either reimplement `driver.Rows` (duplicating arrow/CloudFetch/scanner/telemetry code) or re-cut `internal/rows`. Either contradicts the stated contract.

This isn't a bug in the shipped code — it's a scoping honesty issue for the follow-up. Suggested: either introduce a narrow neutral `ResultSource` seam now (fetch-next-page / metadata / close over Arrow-IPC bytes or a neutral chunk struct) that `internal/rows` consumes, with the Thrift `TCLIService`-backed source as its first impl; or explicitly state that the rows layer will be re-cut in Layer 2 and that `Results` returning `driver.Rows` is a boundary, not a reuse boundary.

---

internal/backend/thrift/operation.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return true, err
+	}
+	return true, nil
+}
+
+// ExecutionError wraps cause as the driver's execution error, attaching this
+// operation's terminal sqlstate from the status response. Returns nil when cause
+// is nil.
+func (o *thriftOperation) ExecutionError(ctx context.Context, cause error) error {
```
mani-mathur-arch, line 141: The deleted `connection_test.go` subtest drove the full `ExecContext` across all five terminal operation states and asserted the exact user-facing string, e.g. `assert.EqualError(t, err, "databricks: execution error: failed to execute query: unexpected operation state ERROR_STATE: error state")`. The move split this into `TestBackend_runQuery` (state machine — asserts only `assert.Error`, not the message) and `TestOperationClose` (close matrix — asserts the `closed` bool). The composition that produces the string the user actually sees — `op.ExecutionError(ctx, err)` wrapping `unexpectedOperationState(status)` with `ErrQueryExecution` + sqlstate — is now asserted by **nothing** (grep for `EqualError`/`ErrorContains`/`unexpected operation state` across the thrift `*_test.go` at head returns nothing).

A future change to `ExecutionError`, `unexpectedOperationState`, or the error format would ship green while altering the exact error string returned from a failed `ExecContext`/`QueryContext`, silently breaking downstream tooling that matches on it. Suggested fix: add one subtest asserting `op.ExecutionError(ctx, unexpectedOperationState(opStatus))` produces the exact composed string and carries the sqlstate — or restore one end-to-end `ExecContext` case driving a terminal ERROR/CANCELED state with `assert.EqualError`.

---

internal/backend/backend.go
```diff
@@ -0,0 +1,109 @@
...
+	SessionValid() bool
+
+	// SessionID is the formatted server session id (conn.id). Valid only after a
+	// successful OpenSession; empty otherwise.
+	SessionID() string
+
+	// Execute runs a statement to a terminal state and returns an Operation
+	// describing the outcome.
+	//
+	// An implementation MUST return a non-nil Operation, even when err is non-nil:
```
mani-mathur-arch, line 44: The interface states `Execute` "MUST return a non-nil Operation, even when err is non-nil" — and `connection.go` depends on it: after `op, err := c.runQuery(...)`, the `op == nil` branch is treated as a *pre-backend* param failure (skips close/telemetry), while a non-nil-op-with-error path calls `op.StatementID()`, `op.Close()`, `op.ExecutionError()`. If `Execute` ever returned `(nil, err)` on the post-backend path, either close/telemetry are silently skipped or a later `op.` call nil-panics.

This invariant has no test. The only `be.Execute(...)` call in the thrift test files is the success case in `debuglog_integration_test.go` (discards the op); `TestBackend_runQuery` tests the unexported `runQuery`, not the `Execute` wrapper that builds the non-nil `thriftOperation`. A future `if err != nil { return nil, err }` in `Execute` would pass every existing test. Suggested fix: `TestBackend_Execute` with a failing `FnExecuteStatement` asserting `require.NotNil(op)` + `require.Error`, plus `op.StatementID() == ""` and `op.Close(ctx) == (false, nil)`.

---

internal/backend/thrift/operation.go
```diff
@@ -0,0 +1,137 @@
...
+		return false, err
+	}
+	return resp.IsStagingOperation != nil && *resp.IsStagingOperation, nil
+}
+
+// Close best-effort closes the server operation. It skips the CloseOperation RPC
+// when there is no handle, when direct results already closed it, or when the
+// operation is already in the CLOSED state; closed reports whether the RPC was
+// actually sent, so the caller records CLOSE_STATEMENT telemetry only in that case.
+func (o *thriftOperation) Close(ctx context.Context) (bool, error) {
```
mani-mathur-arch, line 110: The interface contract (`backend.go:76`) says `Close` "is idempotent and safe to call regardless of whether execution succeeded." The Thrift impl decides whether to skip the RPC purely from immutable fields (`hasHandle()`, `DirectResults.CloseOperation`, `opStatusResp` state); it never records that it already issued a `CloseOperation` and never mutates `opStatusResp`. So a second `Close()` on a live FINISHED/ERROR/CANCELED handle takes the same branch as the first: it issues `CloseOperation` **again** and returns `closed=true` again (re-recording `CLOSE_STATEMENT` telemetry).

A caller relying on the documented idempotency (e.g. a `defer op.Close()` plus an explicit error-path `op.Close()`) sends a duplicate RPC and double-fires telemetry. `TestOperationClose` only ever calls `Close` once, so this is untested. Fix: either drop the "idempotent" claim, or set a `closed bool` after issuing the RPC and short-circuit subsequent calls to `(false, nil)`.

---

connection.go
```diff
@@ -124,25 +119,38 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
-	log, ctx = client.LoggerAndContext(ctx, exStmtResp)
+	op, err := c.runQuery(ctx, query, args)
+	// A nil op means execution never reached the backend (parameter conversion
+	// failed): no server statement to close or measure. Return the already-wrapped
+	// error directly.
+	if op == nil {
+		log.Err(err).Msgf("databricks: failed to execute query: query %s", query)
+		return nil, err
+	}
+	// Enrich the logger/context with the statement id, matching the removed
```
mani-mathur-arch: Three comments describe `enrichQueryId` as matching the "removed" (lines 136, 269) / "pre-refactor" (line 406) `client.LoggerAndContext` semantics, and `enrich_queryid_test.go:12` says the semantics are "preserved from the removed client.LoggerAndContext." The function was **not** removed — it's defined at `internal/client/client.go:421` and still called at the top of these very functions (`ExecContext` at connection.go:119, `QueryContext` at connection.go:249). What was removed is one *call* inside the old `runQuery`, not the symbol.

A future reader who greps `LoggerAndContext`, finds it alive and invoked three lines up, will distrust the comment and the surrounding subtle query-id logic. Suggested fix: state the contract directly, e.g. `// enrichQueryId fills the context queryId from the statement id only when the caller hasn't set one, and always calls NewContextWithQueryId on the empty branch so a registered QueryIdCallback still fires.` The "this used to live in LoggerAndContext" history belongs in the commit message.

---

internal/backend/thrift/backend_test.go
```diff
@@ -0,0 +1,1237 @@
...
+	"github.com/databricks/databricks-sql-go/internal/cli_service"
+	"github.com/databricks/databricks-sql-go/internal/client"
+	"github.com/databricks/databricks-sql-go/internal/config"
+	"github.com/databricks/databricks-sql-go/internal/thrift_protocol"
+	"github.com/stretchr/testify/assert"
+)
+
+// newTestBackend builds a Backend around a mock client and canned session,
+// bypassing New/OpenSession.
+func newTestBackend(cli cli_service.TCLIService, session *cli_service.TOpenSessionResp, cfg *config.Config) *Backend {
```
mani-mathur-arch: `newTestBackend` (unexported, `backend_test.go:24`) and `NewForTest` (exported, `export_test_helper.go:25`) have identical six-line bodies and both live in package `thrift`. Because `export_test_helper.go` lacks the `_test.go` suffix, `NewForTest` is compiled into the package proper and is directly callable from `backend_test.go` — there's no visibility barrier forcing a second copy. The two must be kept in lockstep; when the `Backend` struct gains a required field or `sessionID` derivation changes (likely as the SEA backend lands), editing one and not the other yields test-only divergence. Fix: collapse to one — have `newTestBackend` delegate to `NewForTest`, or call `NewForTest` directly at the ~30 sites.

---

internal/backend/thrift/operation.go
mani-mathur-arch, line 67: Of the six `Operation` accessors, `Close` and `IsStaging` are well-covered directly in `operation_test.go`, but `AffectedRows`, `StatementID` (handle-less → `""`), and `Results` have no direct tests — only indirect happy-path coverage via conn-level tests. The handle-less zero-value contracts stated in the interface doc are the ones a second (SEA) implementer will read as the spec, so they should have direct tests. Low impact today (indirect coverage exercises the happy path; the Thrift getters are nil-tolerant). Fix: extend `operation_test.go` with a small table asserting `StatementID()` returns `""` for a handle-less op and the GUID otherwise, and `AffectedRows()` returns `NumModifiedRows` (and `0` when `opStatusResp` is nil).

---

mani-mathur-arch:

**Score**: 77/100 — MODERATE RISK

A clean, genuinely behavior-preserving refactor on the control plane — security, language, ops, and performance reviewers found nothing at bar after reading the moved code against the pre-PR original, and no correctness or security regression ships here. The deductions are one architectural scoping gap (the rows/data plane stays Thrift-bound), two test-coverage regressions the move introduced, and interface-doc gaps that would bite the *next* PR (the SEA backend) more than this one. One devil's-advocate finding (QueryIdCallback allegedly stops firing on the param-error path) was dropped as a verified false positive — `client.LoggerAndContext(ctx, nil)` still fires it at the top of ExecContext/QueryContext.

---

internal/backend/backend.go
mani-mathur-arch: Fixed in 276f409. Documented the exec-path (caller uses Close) vs query-path (Rows.Close owns close) split on the Operation type doc, and added a corresponding note on Results and Close. Both closers are now specified as independently idempotent

internal/backend/thrift/operation.go
mani-mathur-arch: Doc fixed in 276f409: tightened the package-doc neutrality claim to carve out row decoding. The behavior half (neutral ResultSource seam or re-cutting internal/rows) is deferred to a follow-up PR .

internal/backend/thrift/operation.go
mani-mathur-arch: Fixed in 276f409. Added TestOperationExecutionError in internal/backend/thrift/operation_test.go: nil-cause → nil; the exact composed string "databricks: execution error: failed to execute query: unexpected operation state <STATE>:  <displayMsg>" for ERROR/CANCELED/TIMEDOUT/CLOSED; and sqlstate propagation via errors.As(err, &DBExecutionError).

internal/backend/backend.go
mani-mathur-arch: Fixed in 276f409. Added TestBackend_Execute with two subtests, a pre-handle failure (op != nil, StatementID() == "", Close() returns (false, nil) and issues no RPC) and post-handle terminal-error (op != nil, non-empty StatementID, Close()  returns (true, nil) and issues the RPC, ExecutionError builds off the non-nil op).

internal/backend/thrift/operation.go
mani-mathur-arch: Fixed in 276f409. Added a closed bool memo to thriftOperation; Close short-circuits with (false, nil) on a second call (whether the first succeeded or errored). Added TestOperationClose_Idempotency pinning both paths.

connection.go
mani-mathur-arch: Fixed in 276f409. Rewrote the four locations (connection.go:136/:269/:406 and the enrich_queryid_test.go header + inline comment) to state the contract directly instead of narrating history.

internal/backend/thrift/backend_test.go
mani-mathur-arch: Fixed in 276f409. Deleted newTestBackend and renamed all call sites across the three test files to use the exported NewForTest in export_test_helper.go

internal/backend/thrift/operation.go
mani-mathur-arch: Fixed in 276f409. Added TestOperationStatementID (empty when no handle, correct formatted GUID when present, memo check) and TestOperationAffectedRows (NumModifiedRows on success path, 0 when unset).

vikrantpuppala (approved):
lgtm - almost perfect pr, thank you!

vikrantpuppala:
let's name stuff as kernel rather than sea (in future prs)

PR #389, opened 2026-07-08
Stop using html/template in U2M OAuth callback to restore linker DCE (fixes #343)

vikrantpuppala:
## Summary

Reachable use of the stdlib `html/template` in package `auth/oauth/u2m` disabled the Go linker's public dead-code elimination (DCE) for the **entire binary** of any application that imports `databricks-sql-go`. `html/template` builds on `text/template`, whose reflection-based field evaluation forces the linker to retain code it would otherwise strip. This bloats consumer binaries — up to ~20% (40 MB) in multi-library binaries like [`xo/usql`](https://github.com/xo/usql), as reported in #343.

This PR rewrites the mostly-static OAuth callback page as plain string building with explicit `html.EscapeString` on the dynamic fields (the same approach [`xo/tblfmt`](https://github.com/xo/tblfmt/commit/c8da2821c2c334f51dac261563158550bbb6a6ac) used for a similar issue), removing `html/template`/`text/template` from the reachable graph. The now-unused embedded `templates/simple.html` is deleted.

The dynamic fields remain HTML-escaped, so the escaping behavior of the previous `html/template` implementation is preserved.

## Impact

Measured with a minimal `main.go` that imports the driver (the repro from the issue):

| | Before | After |
|---|---|---|
| `text/template`/`html/template` symbols (`go tool nm`) | 615 | **0** |
| Binary size | 23.7 MB | **14.4 MB (~39% smaller)** |

The size delta is larger than the driver's own footprint because re-enabling DCE benefits the whole binary, including other libraries.

## Test plan

- Added `auth/oauth/u2m/html_template_test.go`:
  - `TestRenderHTML` (table-driven): basic field rendering + escaping of `<`, `>`, `&`, `"`; the conditional action-link block; the conditional code block; and that each page is a complete, well-formed document.
  - `TestInfoHTML`: success page title/heading/content, no code block.
  - `TestErrorHTML`: error page title and that the error message is HTML-escaped (no raw `<script>` leaks through).
- `go build ./...`, `go vet ./auth/oauth/u2m/`, and `gofmt -l` are all clean.
- `go test ./auth/oauth/u2m/` passes.
- Verified via `go tool nm` on a consumer binary that the template symbols are gone (615 → 0) and the binary shrinks as shown above.

## Notes

- No other `html/template` or `text/template` usages exist in the codebase — this was the sole occurrence of the pattern.
- CHANGELOG updated under an `Unreleased` section.

Closes #343


vikrantpuppala:
Superseded by #390, which is branched cleanly off main (this branch accidentally included an unrelated README commit from docs/196-dsn-parameters).

PR #393, opened 2026-07-09
feat(kernel): SEA-via-kernel backend (opt-in, behind build tag)

mani-mathur-arch:
## Summary

Add a second execution backend that runs statements over the Statement Execution API via the Rust `databricks-sql-kernel`, reached through a cgo C ABI.

It is opt-in behind `WithUseKernel(true)` + the `databricks_kernel` build tag. The default build stays pure-Go (`CGO_ENABLED=0`) and continues to use the existing Thrift backend unchanged.

The user-facing API is otherwise the same: the kernel backend reads the same `config.Config` as Thrift and either forwards supported options or rejects unsupported ones loudly with typed errors.

## What’s Implemented

- **Auth:** PAT, including PAT supplied through `WithAuthenticator(&pat.PATAuth{...})`
- **Execution:** SQL read/write, Direct Results, Cloud Fetch, affected rows, query tags, server statement timeout via session conf
- **Cancellation:** execute-path context cancellation through the kernel statement canceller
- **Data types:** primitives, binary, date/timestamp in session timezone, decimal, array, map, struct, variant, geometry/WKT
- **Result parity:** kernel rendering is byte-identical to Thrift for covered scalar and nested Arrow values
- **Connectivity:** TLS skip-verify, HTTP proxy resolution, SPOG org routing via HTTP path
- **Errors:** kernel errors map into the driver error surface with sqlstate/query id where available
- **Safety:** statement-path errors never return `driver.ErrBadConn`, avoiding silent database/sql replays of possibly committed DML
- **Build/CI:** tagged kernel backend now builds and runs unit tests in CI via `make test-kernel`

## Unsupported or Deferred

Unsupported kernel options/features fail loudly instead of being silently ignored. Rejections wrap `errors.ErrNotSupportedByKernel` so callers can detect them with `errors.Is`.

Rejected at connect time:

- Initial namespace (`catalog` / `schema`)
- Metric-view metadata
- Non-PAT authenticators, including OAuth
- `WithTimeout`
- `WithRetries(-1)` retry-disable mode
- Non-default `WithPort`
- Non-https protocol
- Custom `WithTransport`

Rejected at execute time:

- Bound query parameters
- Staging operations (`PUT` / `GET` / `REMOVE`)

Deferred kernel/C-ABI follow-ups:

- Success-path statement/query id accessor for telemetry
- Mid-fetch cancellation
- Session lifecycle deadlines for `OpenSession` / `CloseSession`
- Published per-platform kernel artifacts instead of source-building the `.a`
- Full single-sourcing of the Thrift and kernel nested renderers

## Structure

- `internal/backend/kernel/`: cgo kernel backend implementation
- `internal/arrowscan/`: pure-Go Arrow cell and nested JSON renderer used by the kernel path
- `internal/decimalfmt/`: exact fixed-point decimal formatter shared with Thrift
- `kernel_config.go`: pure-Go config validation and PAT resolution
- `kernel_stub.go`: default-build stub returning `ErrKernelNotCompiled`
- `build/kernel-lib.sh` / `Makefile`: pinned kernel static-lib build recipe
- `.github/workflows/go.yml`: tagged kernel CI job

## Test Plan

- [x] `CGO_ENABLED=0 go test ./...`
- [x] Default pure-Go CI remains green
- [x] `Test (kernel backend)` CI job builds the pinned Rust kernel static library and runs `go test -tags databricks_kernel ./...`
- [x] Untagged tests cover:
  - config validation and unsupported-option rejections
  - kernel-not-compiled stub
  - error classification / no statement replay
  - proxy resolution
  - staging detection
  - decimal formatting
  - Arrow scalar and nested rendering
  - cross-backend Thrift/kernel rendering parity
- [x] Tagged tests cover:
  - kernel backend wiring
  - execute-time parameter/staging rejections
  - execution-error contract
- [x] Live staging verification performed for select, data types, Cloud Fetch, cancellation, query tags, timezone, TLS skip-verify, and Thrift parity
- [x] `golangci-lint` clean


internal/arrowscan/arrowscan.go
```diff
@@ -0,0 +1,329 @@
...
+	case *array.Null:
+		return nil, nil
+	case *array.Boolean:
+		return c.Value(row), nil
+	case *array.Int8:
+		return int64(c.Value(row)), nil
+	case *array.Int16:
+		return int64(c.Value(row)), nil
+	case *array.Int32:
+		return int64(c.Value(row)), nil
```
mani-mathur-arch: **Top-level TINYINT/SMALLINT/INT return different Go types across backends; the new scalar-parity test omits exactly these widths.**

The kernel's `ScanCell` widens every narrow integer to `int64` (`*array.Int8`/`Int16`/`Int32` → `int64(c.Value(row))`). The Thrift path returns the **native** width: `makeColumnValueContainer` maps `*arrow.Int8Type → columnValuesTyped[*array.Int8, int8]` (arrowRows.go:620-627), whose `Value` returns raw `int8`/`int16`/`int32`. So a top-level `TINYINT`/`SMALLINT`/`INT` yields `int64` on the kernel but `int8`/`int16`/`int32` on Thrift.

`TestArrowbasedKernelTopLevelScalarParity` (added round 5) claims to assert "the same Go value," but its cases are only int64/float32/float64/string/timestamp/date32 — the integer widths where the divergence lives are omitted, so the test passes while its stated contract is violated.

**Failure scenario**: `db.QueryRow("SELECT CAST(1 AS TINYINT)").Scan(&x)` with `var x any` → `int64(1)` on kernel, `int8(1)` on Thrift. A downstream `v.(int64)` works on one backend and panics on the other. (Scanning into a typed `*int64` is smoothed by `convertAssign`; the raw `driver.Value` type genuinely differs.) `TestKernelE2EDataTypes` expects int64 and runs only on the kernel, so it can't surface the mismatch.

<details>

either widen the Thrift narrow-int containers to `int64` for parity, or return native `int8`/`int16`/`int32` from `arrowscan`; then add the three integer-width cases to `TestArrowbasedKernelTopLevelScalarParity` so the invariant is actually guarded.

</details>
---

kernel_config_test.go
```diff
@@ -0,0 +1,212 @@
...
+var kernelConfigFieldDisposition = map[string]string{
+	// Forwarded to kernel.Config (see newKernelBackend).
+	"Host":          "forwarded",
+	"HTTPPath":      "forwarded",
+	"WarehouseID":   "forwarded",
+	"AccessToken":   "forwarded", // as the resolved PAT (kc.Token)
+	"Authenticator": "forwarded", // PAT authenticator resolved to the token
+	"Location":      "forwarded",
+	"SessionParams": "forwarded",
+	"Port":          "forwarded", // part of the endpoint URL (host:port)
```
mani-mathur-arch: **`Port` and `Protocol` are classified "forwarded" but `kernel.Config` has no such fields — the drop-guard passes on a false label.**

`kernelConfigFieldDisposition` records `"Port": "forwarded"` and `"Protocol": "forwarded"`, but `kernel.Config` (backend.go) has no `Port` or `Protocol` field, and `newKernelBackend` forwards only Host/HTTPPath/WarehouseID/Token/Location/SessionConf + TLS/proxy. `cfg.Port`/`cfg.Protocol` are never passed to the kernel.

`TestKernelConfigFieldsClassified` still passes because the names *are* in the map — so the exhaustiveness guard, whose entire purpose is to force forward/reject/inert and prevent silent drops, is satisfied by a factually incorrect "forwarded" label.

**Failure scenario**: `WithPort(8443)` (or a DSN non-443 port, or `http` protocol for a private-link/test endpoint) routes to that port on Thrift (`ToEndpointURL` uses host:port) but is silently ignored on the kernel path — the exact silent-divergence the disposition contract and doc.go's "None of these is silently ignored" exist to prevent.

<details>

reclassify honestly. If a non-default `Port`/`Protocol` is unsupported on the kernel path, reject it in `validateKernelConfig` (loud); otherwise mark it `"inert"` and document it alongside the other inert knobs — not `"forwarded"`.

</details>
---

internal/backend/kernel/kernel_test.go
```diff
@@ -0,0 +1,160 @@
+//go:build cgo && databricks_kernel
```
mani-mathur-arch, line 1: **The H1/L1 safety-critical error-classifier and its tests run in NO CI job.**

`isBadConnection`, `isSessionFatal`, `toStatementError`, `toConnError` — the functions that decide which errors are allowed to trigger `database/sql`'s transparent statement replay — and their tests (`TestToStatementErrorNeverBadConn`, `TestEvictIfSessionFatal`, `TestIsBadConnection`, `TestIsSessionFatal`) all carry `//go:build cgo && databricks_kernel`. CI runs only `make test` with `CGO_ENABLED=0` and no `-tags databricks_kernel` job (the go.yml comment confirms it is deferred). These guards run nowhere in CI.

**Failure scenario**: a future edit — folding `Unauthenticated` back into `isBadConnection`, or making `toStatementError` return `ErrBadConn` for `Unavailable` — reintroduces the exact silent-duplicate-write bug (a retried, possibly-already-executed INSERT/UPDATE/MERGE) or auth-retry storm this PR spent rounds 4-5 closing, and `make test` stays green. The status consts are `int(C.KernelStatusCode_...)`, binding them to cgo despite a comment claiming non-cgo code can reference them.

This is the single most safety-critical logic in the backend and the most cleanly untaggable.

<details>

mirror what was already done for `arrowscan`/`decimalfmt`/`kernel_config`/`kernel_proxy` — extract `KernelError`, the `status*` consts (as plain Go consts with a lockstep comment / compile-time assertion against the C enum), and the four classifier functions into an untagged `.go` file; move their tests to an untagged `_test.go` so they run under `CGO_ENABLED=0`.

</details>
---

internal/rows/arrowbased/columnValues.go
```diff
@@ -364,18 +363,27 @@ func (mvc *mapValueContainer) SetValueArray(colData arrow.ArrayData) error {
 type structValueContainer struct {
 	structArray     *array.Struct
 	fieldNames      []string
```
mani-mathur-arch: **`structValueContainer.fieldNames` is dead — allocated and populated but never read.**

The round-5 fix added `fieldKeys` (the precomputed escaped `"name":` prefix) and rewrote `Value` to read it. `fieldNames` is still declared here and populated (arrowRows.go:690, 695) but a full-repo grep shows zero readers.

**Impact**: wasted alloc + populate loop per struct column, plus a latent trap — `fieldNames` holds the raw *unescaped* names, so a maintainer reaching for it thinking it mirrors `fieldKeys` would reintroduce the exact JSON-escaping bug this PR fixed.

<details>

remove the `fieldNames` field and its two write sites (arrowRows.go:690 and 695). `fieldKeys` fully replaces it.

</details>
---

internal/rows/arrowbased/arrowscan_parity_test.go
```diff
@@ -0,0 +1,318 @@
...
+			}
+			thrift := renderViaArrowbased(t, arr, 0, loc)
+			if kernel != thrift {
+				t.Errorf("top-level %s divergence:\n  kernel = %#v (%T)\n  thrift = %#v (%T)", tc.name, kernel, kernel, thrift, thrift)
+			}
+		})
+	}
+}
+
+// TestTopLevelDecimalRendering documents the top-level DECIMAL story (review r5
```
mani-mathur-arch: **Review-round narration survived the round-5 comment cleanup.**

`TestTopLevelDecimalRendering`'s doc comment opens with `documents the top-level DECIMAL story (review r5 M1-r5)`. `M1-r5` is a review-thread finding ID that rots on merge — a future reader cannot resolve it. The round-5 cleanup removed this class of narration from `kernel_config_test.go` but missed this instance.

<details>

drop the `(review r5 M1-r5)` parenthetical. The rest of the comment (the three-bullet explanation of why the paths differ) is legitimate WHY-rationale and should stay.

</details>
---

internal/rows/arrowbased/arrowscan_parity_test.go
```diff
@@ -0,0 +1,318 @@
...
+	// decimal + nested read); DecimalStringValue is the exact top-level scan path.
+	// Assert both so a change to either is caught.
+	holder := &decimal128Container{scale: dt.Scale}
+	if err := holder.SetValueArray(arr.Data()); err != nil {
+		t.Fatal(err)
+	}
+	if s := holder.ValueString(0); s != "1234567890123456.7890" {
+		t.Errorf("Thrift DecimalStringValue = %q, want exact string", s)
+	}
+	if v, _ := holder.Value(0); v == "1234567890123456.7890" {
```
mani-mathur-arch: **`TestTopLevelDecimalRendering`'s lossy-path assertion is near-dead.**

The DECIMAL reconciliation itself is correctly pinned (both exact strings agree via an independent literal — no false equality). But this auxiliary assertion, `if v, _ := holder.Value(0); v == "1234567890123456.7890"`, compares an `any` holding a `float64` against a `string` literal. In Go that interface comparison is **always false** (different dynamic types), so the error branch can only fire if `Value()`'s return *type* changes — it never verifies the lossy float64 *value*. The comment "documented lossy-float64 path" overstates what this line catches.

Minor — the load-bearing parity assertions are solid, so no real divergence can hide; this just gives false confidence in the auxiliary check.

<details>

assert the actual float value, e.g. `if got, ok := v.(float64); !ok || got != 1234567890123456.7890 { t.Errorf(...) }`, so the check pins the lossy float64 rather than only the return type.

</details>
---

mani-mathur-arch:

**Score**: 87/100 — MODERATE RISK

Round 6 of review. Highest score yet and the first round with no High and no Critical — 3 points under the LOW RISK line. All five round-5 findings verified closed against source, including the two that needed careful checks: the dead-session eviction is race-free and preserves the no-replay guarantee, and the struct-key memoization is bounded per-`kernelRows` (not the earlier global-cache leak). Five of nine reviewer dimensions found nothing at reportable confidence; the architecture assessment was "the design has converged." What remains are three Medium test/contract-completeness gaps and four Low cleanups — no correctness or safety defect in shipping behavior. The kernel backend is opt-in + build-tag-gated with zero production callers. Trajectory across six rounds: 28 → 66 → 75 → 69 → 79 → 87.

---

Could not post inline comments for: L1 — see body below.

## Low Findings

<details>
<summary>Low findings — click to expand</summary>

**L1** (Low) — `doc.go`:L33 — DSN params `useKernel`/`warehouseId` and `WithWarehouseID` are absent from doc.go's parameter/option lists.

**DSN params `useKernel`/`warehouseId` and `WithWarehouseID` are absent from doc.go's parameter/option lists.** (Low severity.)

`ParseDSN` parses `useKernel` and `warehouseId` (config.go:332-340, covered by a passing test), and `WithWarehouseID` is a public option — but doc.go's "Supported optional connection parameters" list, "Supported functional options" list, and the SEA section (which documents only the `NewConnector`/`WithUseKernel` object path) omit all three.

**Impact**: an agent or user told to select the kernel backend via a DSN string scans the supported-parameters list, finds no `useKernel`, and wrongly concludes the DSN path can't select the kernel backend. `warehouseId` is undocumented everywhere in doc.go, so the bare-warehouse-id addressing (which the backend actually prefers over HTTPPath) is undiscoverable.

- **Suggested fix**: add `useKernel` and `warehouseId` to doc.go's DSN parameter list, and add `WithUseKernel`/`WithWarehouseID` to the functional-options list (or one sentence in the SEA section noting the DSN equivalents). If DSN-level kernel selection is intentionally undocumented for an experimental feature, note that scoping explicitly.



</details>

internal/arrowscan/arrowscan.go
mani-mathur-arch: Fixed. ScanCell now returns the native int8/int16/int32 instead of widening to int64, so both backends return the same type for TINYINT/SMALLINT/INT. Added tinyint/smallint/int cases to the parity test, and updated the e2e expectations to the native widths. Went the "match Thrift" way rather than widening Thrift, since Thrift has real callers and the kernel doesn't. Filed a separate follow-up to move both backends to int64 (the spec-correct end state),that one needs a maintainer call since it changes Thrift's observable behavior. Although the database/sql only allows int64 but we match the thrift path here as that is already in prod

kernel_config_test.go
mani-mathur-arch: Fixed, the label was lying. The kernel takes a bare host and connects over https:443, so it genuinely can't honor a custom port/scheme. Now validateKernelConfig rejects a non-default port or non-https protocol at connect (defaults still pass), and I reclassified both as "rejected" in the disposition map. Added tests for the reject + the default-accepted case.

internal/backend/kernel/kernel_test.go
mani-mathur-arch: Fixed exactly as suggested. Moved KernelError, the status* consts, and the four classifiers into an untagged errors_classify.go, with the consts as plain ints plus compile-time assertions in cgo.go that keep them in lockstep with the C enum. Their tests moved to an untagged errors_classify_test.go, so the H1/L1 guards now run under CGO_ENABLED=0 in the default job, a regression here fails CI instead of shipping green.

internal/rows/arrowbased/columnValues.go
mani-mathur-arch: Removed the field and both write sites, fieldKeys fully covers it. Agreed on the latent trap too (it held the unescaped names), so good to have it gone.

internal/rows/arrowbased/arrowscan_parity_test.go
mani-mathur-arch: Dropped the (review r5 M1-r5) bit; kept the three-bullet explanation.

internal/rows/arrowbased/arrowscan_parity_test.go
mani-mathur-arch: Fixed, it now asserts v.(float64) == 1234567890123456.7890 instead of the always-false cross-type compare, so it actually verifies the lossy value, not just the type.

vikrantpuppala:

**Score**: 40/100 — HIGH RISK

Well-isolated, opt-in dual-backend feature: the default pure-Go (CGO_ENABLED=0) build is genuinely unchanged and security/performance found nothing. The score is driven by four High-severity gaps on the new kernel backend — silent no-ops, a telemetry blind spot, no CI coverage, and a one-directional drift guard — the kind that ship green and surface as production incidents. None block the default release; all matter before running WithUseKernel(true) in production.

---

internal/backend/kernel/operation.go
```diff
@@ -0,0 +1,251 @@
...
+		o.backend.evictIfSessionFatal(err)
+		o.close()
+		return nil, fmt.Errorf("kernel: get_result_stream: %w", toStatementError(err))
+	}
+	return newKernelRows(ctx, o, stream, callbacks)
+}
+
+// IsStaging reports whether this is a staging (PUT/GET/REMOVE) operation. The
+// kernel backend does not support staging, so this is always false.
+func (o *kernelOp) IsStaging(ctx context.Context) (bool, error) { return false, nil }
```
vikrantpuppala, line 234: `kernelOp.IsStaging` unconditionally returns `false, nil`, so `conn.execStagingOperation` returns early and the driver **never performs the local file transfer** for a volume-ingest statement — the call returns success while no upload/download happened. Every *other* unsupported feature returns a loud error; staging violates the PR's own "nothing silently ignored" contract and is **absent from the doc.go caveat list**. A user migrating a volume workload to the kernel backend sees green with silently missing data.

- **Suggested fix:** detect staging statements on the kernel path and return an explicit execute-time "not yet supported" error (mirroring the bound-params rejection); at minimum document it.

---

internal/backend/kernel/operation.go
```diff
@@ -0,0 +1,251 @@
...
+var _ backend.Operation = (*kernelOp)(nil)
+
+// StatementID returns "": the kernel C ABI exposes no success-path statement/query
+// id accessor (only the error path carries a query_id, on KernelError). Because
+// conn.ExecContext/QueryContext gate per-statement telemetry on
+// StatementID() != "", kernel queries currently emit no EXECUTE_STATEMENT metric
+// and their query-id log field is empty — there is no session-id fallback on that
+// gate. Surfacing an id needs a kernel accessor (e.g.
+// kernel_executed_statement_query_id); tracked as a follow-up.
+func (o *kernelOp) StatementID() string { return "" }
```
vikrantpuppala, line 201: `kernelOp.StatementID()` always returns `""`, and `conn.ExecContext`/`QueryContext` gate **all** per-statement telemetry (EXECUTE_STATEMENT, execution latency, failure recording, CLOSE_STATEMENT) on `StatementID() != ""`. Additionally `kernelRows.Close()` (rows.go:88) never fires the `OnClose` callback. Net effect: kernel-path queries emit **zero per-query telemetry, including failures** — dashboards/alerts that fire on the Thrift path show nothing for kernel traffic; detection depends on the customer reporting. Documented in doc.go for EXECUTE_STATEMENT, but the CLOSE_STATEMENT drop is undocumented.

- **Suggested fix:** expose a kernel query-id via the C ABI so the existing telemetry gate engages; until then emit at least a session-scoped success/failure/latency metric so kernel traffic isn't a blind spot.

---

.github/workflows/go.yml
```diff
@@ -90,6 +90,12 @@ jobs:
           fi
           go get -v -t -d ./...
 
+      # This matrix builds the default pure-Go driver (CGO_ENABLED=0), which does
```
vikrantpuppala, line 93: The workflow comment itself states a `CGO_ENABLED=1 -tags databricks_kernel` job "needs" to be added — confirming no such job exists. All tagged files (`backend.go`, `cgo.go`, `operation.go`, `rows.go`) compile/run only under the tag, and the only tests touching them are e2e/parity tests that both carry the tag **and** `t.Skip` without live creds. So the cancel-watcher goroutine, per-batch release/EOF handling, and `close()` idempotency **never execute in any pipeline**. A deadlock/leak/double-free ships green.

- **Suggested fix:** add a `CGO_ENABLED=1 -tags databricks_kernel` job (build-only + a stubbed C ABI for the watcher/close logic) so lifecycle code is exercised without a live warehouse. Track the acknowledged follow-up explicitly.

---

internal/backend/kernel/cgo.go
```diff
@@ -0,0 +1,180 @@
...
+	return ke
+}
+
+// The plain-int status constants used by the untagged classifier logic
+// (errors_classify.go) must stay in lockstep with the C enum in
+// databricks_kernel.h. These compile-time assertions make a drift a build error
+// under -tags databricks_kernel: the array length is 0 when the values match and
+// negative (illegal) otherwise, so the file won't compile if the header changes.
+var (
+	_ [statusInvalidArgument - int(C.KernelStatusCode_InvalidArgument)]struct{}
```
vikrantpuppala: The guard `_ [statusInvalidArgument - int(C.KernelStatusCode_InvalidArgument)]struct{}` fails to compile only when the C value *exceeds* the Go constant (negative array size). If the Rust header renumbers a code *downward* (e.g. 1→0), the difference is positive, the array is legal, and the drift is **undetected** — then `isBadConnection`/`isSessionFatal` silently misclassify: a transient failure stops producing `driver.ErrBadConn` (no retry) or a benign SQL error starts evicting the pooled conn. Not caught by CI (tagged tests need a live warehouse). Verified: `var _ [1-0]struct{}` compiles; `[1-2]struct{}` errors.

- **Suggested fix:** make the assertion bidirectional, e.g. `const _ = uint(a-b) | uint(b-a)` so either direction underflows.

---

kernel_config.go
```diff
@@ -0,0 +1,94 @@
...
+// resolves the PAT the kernel authenticates with. On success it returns the token
+// to use. Options it does NOT reject are either forwarded by newKernelBackend or
+// intentionally accepted-but-inert (documented in doc.go and asserted by
+// TestKernelConfigFieldsClassified).
+func validateKernelConfig(cfg *config.Config) (token string, err error) {
+	// Initial namespace (WithInitialNamespace): no kernel C-ABI setter yet, so the
+	// session would run in the default namespace and unqualified names would
+	// resolve differently than Thrift.
+	if cfg.Catalog != "" || cfg.Schema != "" {
+		return "", errors.New("databricks: WithInitialNamespace (catalog/schema) is not yet supported by the kernel backend; " +
```
vikrantpuppala: Every kernel rejection is a bare `errors.New("databricks: ...")` (here and at backend.go:242) while the repo ships `errors.DriverError`/`ExecutionError` sentinels and `DBError` interfaces. A caller writing fallback logic ("if kernel rejects params, retry on Thrift") can only fragile-substring-match, which breaks on any wording edit.

- **Suggested fix:** wrap kernel rejections with the existing `internal/errors` constructors or export `ErrKernelUnsupported` sentinels.

---

internal/rows/arrowbased/arrowscan_parity_test.go
```diff
@@ -0,0 +1,343 @@
+package arrowbased
```
vikrantpuppala, line 1: Every map parity case has exactly one entry and every case renders row 0 only. Comma placement, multi-entry key ordering, and cross-row `StructKeyCache` reuse are never compared between backends — the exact divergences the "byte-identical" contract exists to catch.

- **Suggested fix:** add cases with a 2+ entry map and a 2+ row struct/list array asserting `kernel == thrift` per row.

---

internal/backend/kernel/backend.go
```diff
@@ -0,0 +1,246 @@
...
+	// The C ABI exposes no formatted session-id accessor; use the handle pointer
+	// as a stable per-conn id for logging / telemetry correlation.
+	k.sessionID = fmt.Sprintf("kernel-%p", sess)
+	klog("OpenSession OK session=%s", k.sessionID)
+	return nil
+}
+
+// CloseSession tears down the server-side session. Best-effort: the kernel's
+// close is async (see the C header), so an error is logged, not hard-failed.
+func (k *KernelBackend) CloseSession(ctx context.Context) error {
```
vikrantpuppala, line 227: `conn.Close()` → `CloseSession(ctx)` → blocking `C.kernel_session_close` with no deadline or ctx propagation. If the close RPC stalls, `database/sql`'s pool-cleanup goroutine blocks, idle-conn reaping stalls, and graceful shutdown blows its budget.

- **Suggested fix:** bound the close with a watchdog (run detached, return after a fixed timeout) and emit a close-timeout metric so leaked server sessions are visible.

---

internal/arrowscan/arrowscan.go
```diff
@@ -0,0 +1,338 @@
...
+		}
+		if err := writeJSON(b, values, i, loc, keys); err != nil {
+			return err
+		}
+	}
+	b.WriteByte(']')
+	return nil
+}
+
+func writeMapJSON(b *strings.Builder, m *array.Map, row int, loc *time.Location, keys *StructKeyCache) error {
```
vikrantpuppala, line 275: `writeMapJSON` writes every entry key unconditionally in Arrow order, so an Arrow-legal `MAP` with two equal keys renders `{"1":"a","1":"b"}`. This *mirrors* the Thrift path (parity holds), but the shared grammar both backends now commit to is not valid single-value JSON — a strict consumer silently drops the first value or errors. No test covers a repeated map key.

- **Suggested fix:** decide the contract (dedup-last vs. error) and enforce it in *both* renderers, with a parity test.

---

internal/backend/kernel/backend.go
```diff
@@ -0,0 +1,246 @@
...
+// Per the Backend contract it returns a non-nil Operation even on error so the
+// caller can read StatementID / wrap the error / Close uniformly.
+func (k *KernelBackend) Execute(ctx context.Context, req backend.ExecRequest) (backend.Operation, error) {
+	// Bound parameters are not yet wired for the kernel backend. Reject them with
+	// a clear error rather than silently shipping the query with unbound
+	// placeholders (which would behave differently than Thrift). Parameters arrive
+	// per-query, so this is an execute-time error, not a connect-time one. Return a
+	// non-nil Operation per the Backend contract.
+	if len(req.Params) > 0 {
+		return &kernelOp{}, errors.New("databricks: query parameters are not yet supported by the kernel backend; " +
```
vikrantpuppala: The bound-params rejection hardcodes a string that diverges from the existing `errors.ErrParametersNotSupported` constant; an agent grepping the canonical constant misses the kernel path.

**Fix:** reference/extend the shared constant.

---

internal/backend/kernel/backend.go
```diff
@@ -0,0 +1,246 @@
...
+// never transparently re-run (no duplicate write). The backend
+// is single-owner per conn (only the cancel watcher shares state, and it touches
+// only the canceller's inflight slot), so this write is race-free.
+func (k *KernelBackend) markSessionDead() { k.valid = false }
+
+// evictIfSessionFatal marks the session dead when err is a session-fatal
+// KernelError, so a conn whose server session died mid-life is evicted from the
+// pool. A no-op for nil, non-KernelError, or non-fatal errors.
+func (k *KernelBackend) evictIfSessionFatal(err error) {
+	if ke, ok := err.(*KernelError); ok && isSessionFatal(ke.Code) {
```
vikrantpuppala: `evictIfSessionFatal` uses a bare `err.(*KernelError)` type assertion; a future wrapped-error call site would silently fail to evict a dead session.

**Fix:** `var ke *KernelError; if errors.As(err, &ke) && isSessionFatal(ke.Code) { ... }`.

---

internal/backend/kernel/backend.go
```diff
@@ -0,0 +1,246 @@
...
+	err := call(func() C.KernelStatusCode { return C.kernel_session_open(cfg, &sess) })
+	consumed = true
+	if err != nil {
+		return fmt.Errorf("kernel: session_open: %w", toConnError(err))
+	}
+	k.session = sess
+	k.valid = true
+	// The C ABI exposes no formatted session-id accessor; use the handle pointer
+	// as a stable per-conn id for logging / telemetry correlation.
+	k.sessionID = fmt.Sprintf("kernel-%p", sess)
```
vikrantpuppala: `sessionID = fmt.Sprintf("kernel-%p", sess)` derives the id from a malloc'd address that can be reused after free, colliding telemetry/log correlation across sequential connections.

**Fix:** generate a process-unique id (UUID/atomic counter) at OpenSession, independent of the freed pointer.

---

internal/backend/kernel/cgo.go
```diff
@@ -0,0 +1,180 @@
...
+func klog(format string, args ...any) {
+	if !kdebug {
+		return
+	}
+	fmt.Fprintf(os.Stderr, "[kernel] "+format+"\n", args...)
+}
+
+// KernelDebugEnabled reports whether binding-level debug logging is on. Tests
+// assert the flag wiring; benchmarks assert it is false before measuring.
+func KernelDebugEnabled() bool { return kdebug }
```
vikrantpuppala, line 74: `KernelDebugEnabled()` is exported dead code (0 callers) with a doc comment claiming tests/benchmarks assert it — none exist in the PR.

**Fix:** drop it until the benchmark that needs it lands, or add the asserting test now.

---

internal/backend/kernel/errors_classify.go
```diff
@@ -0,0 +1,133 @@
...
+// This file is intentionally NOT behind the `cgo && databricks_kernel` build tag.
+// It holds the kernel error type + the connection-classification logic that
+// enforces two safety-critical guarantees — a statement-path error must never
+// become driver.ErrBadConn, so database/sql cannot silently re-run a possibly-
+// committed statement, and a permanent auth failure is not retried. Keeping
+// it pure Go lets its tests run in the default CGO_ENABLED=0 build; a future edit
+// that reintroduced the duplicate-write bug (or the auth-retry storm) would then
+// fail CI instead of shipping green. The cgo file (cgo.go) populates KernelError
+// from the C struct and asserts these status constants stay in lockstep with the
+// C enum at compile time (see cStatusCodeAssertions).
```
vikrantpuppala: The comment says drift is asserted "at compile time (see `cStatusCodeAssertions`)", but no such symbol exists — the actual construct is the anonymous assertion block in `cgo.go` (~line 161).

**Fix:** replace with "see the compile-time assertion block in cgo.go".

---

internal/backend/kernel/cgo.go
```diff
@@ -0,0 +1,180 @@
+//go:build cgo && databricks_kernel
```
vikrantpuppala, line 1: `klog` and the installed kernel Rust subscriber emit unstructured lines to `os.Stderr`, not through `logger.Logger`, so they carry no `connId`/`corrId`/`queryId` and can't be correlated in a multi-conn process.

**Fix:** route klog through `logger.Logger.Debug()` or include `k.sessionID` in each line.

---

internal/arrowscan/arrowscan_test.go
```diff
@@ -0,0 +1,311 @@
+package arrowscan
```
vikrantpuppala, line 1: The test asserts `cached == uncached`, but both paths call the same `keyPrefixes`, so it proves memoization consistency, not escaping correctness; a wrong-but-consistent escaping rule passes.

**Fix:** add one literal assertion pinning the escaped output to an independent expected value.

---

connector.go
```diff
@@ -301,6 +318,26 @@ func WithHTTPPath(path string) ConnOption {
...
+// `-tags databricks_kernel` and CGO_ENABLED=1; in the default pure-Go build a
+// connection made with this option set returns a clear error at connect time
+// (the kernel backend is not linked in).
+func WithUseKernel(useKernel bool) ConnOption {
+	return func(c *config.Config) {
+		c.UseKernel = useKernel
+	}
+}
+
+// WithWarehouseID sets the bare SQL warehouse id. The kernel backend addresses a
```
vikrantpuppala: **[Low] WithWarehouseID godoc omits that it is inert without WithUseKernel**

The godoc does not state that `WithWarehouseID` has no effect at all unless `WithUseKernel(true)` is also set; a caller setting only it sees it silently ignored on the Thrift path with no error.

**Fix:** add "Has no effect unless WithUseKernel(true) is also set; the default Thrift backend ignores it entirely."

<sub>— Code Review Squad (F15)</sub>

mani-mathur-arch:
Thanks for the thorough review, @vikrantpuppala. I've worked through all of it. Grouping by disposition:
  
  ###  Fixed on this branch (pushed) 

  - **Staging silent no-op** (`operation.go` `IsStaging`) — the kernel path can't perform the local file transfer and the C ABI exposes no `IsStagingOperation` signal, so returning `false` was a silent success with no data moved. Now rejected
  at execute time via a new `isStagingStatement` detector (strips leading whitespace + `--`/`/* */` comments so a comment-prefixed command can't slip through), with a doc.go caveat. Untagged + tested under `CGO_ENABLED=0`.
  - **One-directional drift guard** (`cgo.go`) — the `[a-b]struct{}` assertion only failed when the C value exceeded the Go const; a downward enum renumber went undetected. Switched to a bidirectional `uint(a-b) | uint(b-a)` assertion that
  fails on drift in either direction.
  - **Untyped rejections** (`kernel_config.go`, bound-params in `backend.go`) — added `errors.ErrNotSupportedByKernel` and wrapped every "unsupported option/feature" rejection with it, so callers can `errors.Is` the case instead of
  substring-matching. Added an `errors.Is` test. 
  - **`evictIfSessionFatal` bare type assertion** — switched to `errors.As` so it still fires on a wrapped `KernelError`.
  - **`sessionID` from handle pointer** (`kernel-%p`) — replaced with a process-unique `atomic.Uint64` counter; a freed pointer address can be reused and collide log/telemetry correlation.
  - **Stale/incorrect comments** — corrected the `cStatusCodeAssertions` reference (no such symbol; it's the assertion block in `cgo.go`) and the `KernelDebugEnabled` doc that claimed a test which doesn't exist yet. 
  
  These changes passed a clean Isaac Review pass (0 findings after validation).
  
  ### Already addressed in stacked follow-up branches
  
  - **Telemetry blind spot** (`StatementID()` returns `""` → no per-query metrics) — a stacked branch surfaces the server query id via `kernel_executed_statement_query_id`, which populates `StatementID()` and engages the existing telemetry
  gate.
  - **No `CGO_ENABLED=1 -tags databricks_kernel` CI job** — this is the acknowledged H5 follow-up (blocked on the kernel-lib distribution work); the 2-mode build recipe + tagged CI job land in a stacked branch alongside that.
  
  ### Deferred / tracked (with rationale)
  
  - **Blocking `CloseSession` with no deadline** — tracked as the fire-and-forget-close C-ABI gap; the clean fix needs a `kernel_session_close_blocking` (or watchdog) and is grouped with the other kernel C-ABI asks.
  - **`klog` not routed through `logger.Logger`** (no connId/corrId correlation) — tracked follow-up to unify kernel logging onto the driver's existing `DATABRICKS_LOG_LEVEL` knob rather than the separate `DBSQL_KERNEL_DEBUG` flag.
  - **Repeated-map-key JSON** (`writeMapJSON`) — this mirrors the existing Thrift renderer (parity holds); changing the emitted grammar is a cross-backend contract decision, so I'd rather do it deliberately in its own change than diverge the
  two paths here.
  - **Parity/escaping test strengthening** (multi-entry maps, multi-row, independent escaped-literal assertion) — fair; queued as test-hardening follow-ups. The parity contract is guarded today, these widen the coverage.
  - **`WithWarehouseID` godoc** (inert without `WithUseKernel`) — minor doc clarification, will fold into the doc pass.

internal/backend/kernel/operation.go
```diff
@@ -0,0 +1,253 @@
...
+		o.stmt = nil
+	}
+	klog("kernelOp closed (didClose=%v)", didClose)
+	return didClose
+}
+
+// ExecutionError wraps cause as the driver's execution error. The kernel error
+// already carries the sqlstate (see KernelError), so this returns cause as-is
+// (nil when cause is nil), matching the neutral contract.
+func (o *kernelOp) ExecutionError(ctx context.Context, cause error) error {
```
vikrantpuppala, line 279: `kernelOp.ExecutionError` returns `toStatementError(cause)` — the bare `*KernelError` — while the Thrift path wraps the cause via `NewExecutionError`, whose `Is` matches `dbsqlerr.ExecutionError`/`DatabricksError` and which implements `DBExecutionError` (`SqlState()`, `QueryId()`). `KernelError` has no `Is` method and exposes SQLState/QueryID only as struct fields.

**Impact:** an agent (or any caller) writing backend-agnostic error handling — `errors.Is(err, dbsqlerr.ExecutionError)`, `errors.As(err, &dbExecErr)` to read `SqlState()` — sees it work on Thrift and silently fail on the kernel backend, despite comments claiming "matching the Thrift error surface." SQLState/QueryID become unreachable via the public API even though `KernelError` holds them.

- **Suggested fix:** have `kernelOp.ExecutionError` wrap the cause in the same `NewExecutionError` (or an equivalent carrying `KernelError.SQLState`/`QueryID`) so kernel errors match the same sentinels and expose the same `DBExecutionError` methods as Thrift.

---

internal/arrowscan/arrowscan.go
```diff
@@ -0,0 +1,338 @@
...
+
+// writeJSON writes the JSON form of col[row] into b, recursing for nested types.
+func writeJSON(b *strings.Builder, col arrow.Array, row int, loc *time.Location, keys *StructKeyCache) error {
+	if col.IsNull(row) {
+		b.WriteString("null")
+		return nil
+	}
+	switch c := col.(type) {
+	case *array.List:
+		return writeListJSON(b, c.ListValues(), int(c.Offsets()[row]), int(c.Offsets()[row+1]), loc, keys)
```
vikrantpuppala: `writeJSON` computes child bounds as `c.Offsets()[row]`/`[row+1]` for List/LargeList and `m.Offsets()[row]` for Map (line 239). arrow-go's `List.Offsets()`/`Map.Offsets()` return the full, **un-sliced** offsets buffer, so this indexes `offsets[row]` — whereas correctness (and the Thrift path in `columnValues.go`) require `offsets[row + data.offset]`, exactly what `ValueOffsets(row)` returns.

For any List/Map array with a non-zero logical offset (a sliced array, or a List/Map field of a struct whose parent carries an offset — `Struct.Field` re-slices preserving `data.offset`), the kernel path reads the wrong element range → wrong or panicking nested-value JSON, breaking the stated byte-for-byte parity contract. Leaf scalar accessors are offset-relative and stay correct, which makes the mismatch subtle. The parity tests build only fresh offset-0 arrays, so they cannot catch it.

- **Suggested fix:** use the offset-aware accessor, mirroring Thrift:
  ```go
  case *array.List:
      s, e := c.ValueOffsets(row)
      return writeListJSON(b, c.ListValues(), int(s), int(e), loc, keys)
  // Map: s, e := m.ValueOffsets(row)
  // FixedSizeList: base := (row + c.Data().Offset()) * n
  ```
  (LargeList: arrow-go's own `LargeList.ValueOffsets` omits `data.offset` — verify separately; the Thrift path doesn't handle LargeList.)

---

internal/backend/kernel/cgo.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	klog("kernel error: code=%d sqlstate=%q vendor=%d http=%d retryable=%v msg=%q",
+		ke.Code, ke.SQLState, ke.VendorCode, ke.HTTPStatus, ke.Retryable, ke.Message)
+	// Also emit at the driver's default (Warn) level — no SQL text or PII, just
+	// the status/sqlstate/http fields plus the server query id (a correlation
+	// token, not PII) so on-call can pivot to server-side query history — so a
+	// kernel-path failure is visible without DBSQL_KERNEL_DEBUG. This is the error
+	// path only (never the hot per-row/per-batch path), so it does not perturb
+	// benchmarks.
+	logger.Logger.Warn().Msgf("databricks: kernel call failed: code=%d sqlstate=%q vendor=%d http=%d retryable=%v queryId=%q",
```
vikrantpuppala: `lastError` emits `logger.Logger.Warn()` for every non-Success kernel status, including `SqlError` (syntax/semantic errors) and `InvalidArgument` — routine user faults. The Thrift path treats user SQL errors as user faults excluded from operational noise; this asymmetry floods logs with WARN for a single fat-fingered query.

**Production scenario:** a dashboard/log alert keyed on driver WARN rate fires during normal usage where users submit invalid SQL; on-call chases a non-incident, and genuine session-fatal warnings drown in user-error noise.

- **Suggested fix:** gate the WARN on a fault-vs-error classification — log user faults (`SqlError`, `InvalidArgument`) at Debug/Info and reserve Warn for infra-side codes (`Unavailable`, `NetworkError`, `Timeout`, `Unauthenticated`), reusing the existing `isSessionFatal`/`isBadConnection` predicates.

---

internal/backend/kernel/backend.go
```diff
@@ -0,0 +1,267 @@
...
+			return fmt.Errorf("kernel: set_session_conf[%s]: %w", key, toConnError(errSet))
+		}
+	}
+
+	// kernel_session_open takes ownership of cfg here, on both success and
+	// failure — mark consumed before checking the error so the deferred free
+	// never runs on the already-consumed config.
+	var sess *C.kernel_session_t
+	err := call(func() C.KernelStatusCode { return C.kernel_session_open(cfg, &sess) })
+	consumed = true
```
vikrantpuppala, line 204: The double-free avoidance sets `consumed = true` unconditionally before checking `err`, resting entirely on the comment's claim that `kernel_session_open` reclaims `cfg` on both success and failure. That invariant lives in the Rust C ABI (not visible in this repo), not in Go.

If the kernel ever returns an error **without** taking ownership (e.g. an early argument-validation reject before the box is reclaimed), `cfg` leaks permanently; conversely the pre-`consumed` deferred-free is the only thing between correct and a double-free. A future kernel revision that validates args before consuming the config turns every failed `session_open` into a leak.

- **Suggested fix:** confirm and pin the C ABI ownership contract (a comment referencing the specific kernel API guarantee, or a test), or restructure so the Go side's free/no-free decision is driven by the kernel's return rather than an assumed invariant.

---

internal/backend/kernel/kernel_test.go
```diff
@@ -0,0 +1,79 @@
...
+		t.Error("the statement-path error must not be driver.ErrBadConn (no replay)")
+	}
+}
+
+// Bound parameters are rejected up front by Execute with a clear error, before
+// any session/C work — so this runs on a zero-value backend. The returned
+// Operation must be non-nil (Backend contract) and its Close must report
+// closed=false, since no server statement was ever created (a phantom
+// CLOSE_STATEMENT would otherwise be recorded for it).
+func TestExecuteRejectsParams(t *testing.T) {
```
vikrantpuppala, line 52: `isStagingStatement` is unit-tested in isolation and `TestExecuteRejectsParams` (here) covers the parallel params-rejection arm, but no test drives `Execute` with a staging statement to confirm it returns an `ErrNotSupportedByKernel`-wrapped error plus a non-nil handle-less op. The wiring between the detector and Execute is untested.

A refactor could drop or reorder the staging check in Execute (e.g. move it below `k.execute`, which needs a live session) and the isolated detector test would still pass — reintroducing the silent no-op / data-loss path the check was added to guard against. (Mitigated: this tagged test would only run once a `-tags databricks_kernel` CI job exists.)

- **Suggested fix:** add a tagged `TestExecuteRejectsStaging` mirroring `TestExecuteRejectsParams`: `Execute(ctx, ExecRequest{Query: "PUT '/x' INTO '/Volumes/y'"})`, assert `errors.Is(err, dbsqlerr.ErrNotSupportedByKernel)`, `op != nil`, and `op.Close()` reports closed=false.

---

doc.go
```diff
@@ -155,6 +159,72 @@ The result log may look like this:
...
+	# kernel logs plus its HTTP stack (hyper/reqwest):
+	DBSQL_KERNEL_DEBUG=1 RUST_LOG=debug ./your_app 2>&1
+
+The kernel backend currently supports PAT authentication; reading scalar, nested,
+and complex-typed results (CloudFetch is handled transparently); context
+cancellation during execute (a cancelled ctx fires a real server-side cancel; on
+the read path cancellation is honored at result-batch boundaries, not mid-fetch);
+and the TLS, proxy, and session-conf (query tags, statement timeout, time zone)
+connection options. OAuth (M2M/U2M), initial catalog/schema, and
+WithEnableMetricViewMetadata are not yet supported and return a clear error at
```
vikrantpuppala: `validateKernelConfig` rejects a non-https protocol and a non-443 port at connect time (wrapping `ErrNotSupportedByKernel`), but this caveat block enumerates the connect-time rejections (OAuth, catalog/schema, metric-view, WithTimeout, WithRetries-disable) without listing `WithPort`/non-https.

An agent or user reading doc.go to decide whether a config is kernel-compatible would not learn that a custom `WithPort` fails at connect, and would be surprised by the rejection.

- **Suggested fix:** add `WithPort` (non-443) and non-https protocol to the doc.go list of options rejected at connect time on the kernel backend.

---

internal/backend/kernel/operation.go
```diff
@@ -0,0 +1,253 @@
...
+	// fires would only hammer the server. Falls back to firing until execute
+	// returns (done) if the RPC never dispatches.
+	done := make(chan struct{})
+	var watcherWg sync.WaitGroup
+	if canceller != nil && ctx.Done() != nil {
+		watcherWg.Add(1)
+		go func() {
+			defer watcherWg.Done()
+			select {
+			case <-ctx.Done():
```
vikrantpuppala, line 88: After `kernel_statement_execute` returns, the main goroutine has not yet run `close(done)`. If `ctx` is cancelled in that window, the watcher's first `select` can pick `<-ctx.Done()` and call `fireCancel` on a statement that already reached a terminal (success) state. The canceller is freed only after `watcherWg.Wait()`, so this is memory-safe but semantically fires a cancel RPC against a completed statement.

**What goes wrong:** a query that completes at almost exactly the ctx deadline dispatches a spurious server-side cancel — a server-side no-op, but a wasted RPC and a confusing "cancelled" entry in server-side query-history/audit for a query that actually returned rows.

- **Suggested fix:** after execute returns, re-check `ctx.Err()`/the done signal before the watcher commits to firing, or have `fireCancel` no-op when the statement is already terminal.

---

internal/backend/kernel/cgo.go
```diff
@@ -0,0 +1,187 @@
...
+var initLoggingOnce sync.Once
+
+// initKernelLogging turns on the kernel's own Rust (tracing) logs, gated on the
+// same DBSQL_KERNEL_DEBUG flag as klog so both are OFF by default and, in
+// particular, OFF during benchmarks — the subscriber is never installed in a
+// benchmark run. level=NULL lets the kernel honor RUST_LOG (default warn);
+// file_path=NULL sends kernel logs to stderr so they interleave with klog on one
+// stream. Best-effort: Internal (e.g. the host already installed a global
+// subscriber) is a documented, benign outcome — logged, never fatal to connect.
+func initKernelLogging() {
```
vikrantpuppala, line 100: `kdebug` is captured once at package init from `DBSQL_KERNEL_DEBUG`, and `initKernelLogging` installs the kernel's global Rust log subscriber via `sync.Once` on the first `OpenSession` where `kdebug` is true. The subscriber is process-wide and first-call-wins, never uninstalled — there is no way to scope debug logging to one connection or turn it off after the first kernel session opens.

**What goes wrong:** a long-lived multi-tenant process that opens one debug connection permanently installs the kernel log subscriber (and stderr spew) for all subsequent kernel sessions in the process; the "off during benchmarks" guarantee depends entirely on the env var being unset before package init.

- **Suggested fix:** document the process-wide first-call-wins semantics explicitly, and/or gate the subscriber install behind a per-process config rather than an env var read at package init so it is controllable.

---

vikrantpuppala:
## [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) — Re-review (round 2)

**Score**: 53/100 — HIGH RISK (was 40/100)

The "address review" commits genuinely closed 5 prior findings including 2 Highs (staging now rejected up front, drift guard made bidirectional, deferred rejections typed). This fresh pass at `cbcd20a7` surfaced 8 new findings — including one High parity break the first round missed (kernel exec errors don't satisfy the public `errors.Is`/`As` taxonomy). Three prior Highs/Mediums remain open on their existing threads (telemetry blind spot, multi-entry map parity, CloseSession timeout). None block the default `CGO_ENABLED=0` release; all matter before running `WithUseKernel(true)` in production.

---

internal/backend/kernel/operation.go
vikrantpuppala: Re-review round 2 @ `cbcd20a7`: still open, and there's an additional half to this. `kernelRows.Close()` stores `callbacks` but only ever invokes `OnChunkFetched` — it never calls `OnClose`. So even the empty-StatementID fallback path in `QueryContext` that would `RecordOperation(...CloseStatement...)` never fires either. Net: kernel queries emit *zero* per-statement telemetry (EXECUTE_STATEMENT and CLOSE_STATEMENT), not just the EXECUTE_STATEMENT gap documented in doc.go. Wiring `callbacks.OnClose(...)` from `Close()` would at least restore the close-operation metric.

<sub>— Code Review Squad (round 2 re-review)</sub>

internal/rows/arrowbased/arrowscan_parity_test.go
vikrantpuppala: Re-review round 2 @ `cbcd20a7`: still open, re-confirmed. Worth noting *why* single-entry cases can't catch it: the two backends implement the element separator with structurally different logic — arrowscan uses leading-comma (`if i > start`), the Thrift container uses trailing-comma (`if i < len-1`). For a 1-entry map both coincidentally emit no comma, so they only agree by accident. A 2+ entry map (and a multi-row batch, to exercise per-row offset indexing on rows 1+) is exactly what the parity contract needs to guard.

<sub>— Code Review Squad (round 2 re-review)</sub>

internal/backend/kernel/backend.go
vikrantpuppala: Re-review round 2 @ `cbcd20a7`: still open, re-confirmed. `CloseSession` still ignores its `ctx` and blocks in the synchronous `call()`→`kernel_session_close`; the header comment says close is async but the cgo call still waits for the C function to return, with no deadline. `conn.Close()` builds a `context.Background()` with no timeout, so a stalled kernel-side close (network partition during shutdown) blocks database/sql pool cleanup and delays graceful shutdown past the deployment budget, with no timeout log to distinguish hung from slow.

<sub>— Code Review Squad (round 2 re-review)</sub>

mani-mathur-arch:
Thanks again @vikrantpuppala. Here's a consolidated status across round 1 and round 2. Everything below is pushed to this branch; 
  
  ## Fixed on this branch

  **Correctness**
  - **Nested-value offset bug (round 2, High)** — `writeJSON` indexed List/Map via the raw un-sliced `Offsets()[row]` instead of the offset-aware `ValueOffsets`, so a sliced array (or a List/Map field of a struct carrying a logical offset) read
  the wrong element range or panicked. Now uses `ValueOffsets` for List/Map, manual `+data.offset` for LargeList (arrow-go omits it there), and `(row+offset)*n` for FixedSizeList. Verified against the arrow-go v12.0.1 source.
  - **Staging silent no-op (round 1, High)** — `IsStaging` returning `false` meant a volume `PUT/GET/REMOVE` reported success with no file moved. Now rejected at execute time via a SQL detector (strips leading whitespace + `--`/`/* */` comments
  so a comment-prefixed command can't slip through).
  - **One-directional drift guard (round 1, High)** — the `[a-b]struct{}` assertion only failed when the C value exceeded the Go const; a downward enum renumber went undetected. Switched to a bidirectional `uint(a-b) | uint(b-a)` assertion.

  **Error/observability**
  - **Untyped rejections** — added `errors.ErrNotSupportedByKernel`; every kernel "unsupported option/feature" rejection now wraps it so callers can `errors.Is` instead of substring-matching.
  - **`evictIfSessionFatal`** — `errors.As` instead of a bare type assertion (survives wrapped errors).
  - **`sessionID` from handle pointer** — replaced `kernel-%p` with a process-unique `atomic.Uint64` counter (a freed address can be reused and collide correlation).
  - **WARN log noise (round 2)** — `lastError` now logs user faults (`SqlError`/`InvalidArgument`) at Debug and reserves Warn for infra codes, via a new `isUserFault` predicate.
  - **Spurious cancel (round 2)** — the cancel watcher re-checks the done signal after `ctx.Done()`, so a query completing exactly at the deadline no longer fires a cancel RPC against an already-terminal statement.
  
  **Tests / docs**
  - Parity test now renders **every row** and adds multi-entry map, multi-row list, and **sliced** List/Map/struct cases — confirmed to fail on the pre-fix indexing and pass after (a real regression guard for the offset fix).
  - Escaping test gained an independent literal assertion (was only `cached == uncached`).
  - Tagged `TestExecuteRejectsStaging` drives `Execute` end-to-end (not just the detector in isolation).
  - `isUserFault` unit test.
  - doc.go: added `WithPort`/non-https to the connect-time reject list; `WithWarehouseID` godoc now states it's inert without `WithUseKernel`.
  - Corrected the stale `cStatusCodeAssertions` reference and the `KernelDebugEnabled` doc comment; pinned the `kernel_session_open` ownership contract to the header's documented "CONSUMES config on both success and failure" guarantee.
  
  All of the above passed a clean Isaac Review pass (0 findings after validation), and the full default + kernel-tagged suites pass locally (incl. `-race` on the kernel package).
  
  ##  Planned for follow-up work (not in this PR)
  
  These are on the roadmap for the follow-up work that builds on this backend; calling them out here so they're not lost, but they aren't part of this PR:
  
  - **`ExecutionError` typing (round 2, High)** — wrapping kernel exec errors so `errors.Is(dbsqlerr.ExecutionError)` / `errors.As(&DBExecutionError)` and `SqlState()`/`QueryId()` behave like the Thrift path.
  - **Telemetry** (`StatementID()` empty → no per-statement metrics) — surfacing the server query id to engage the telemetry gate, plus wiring `OnClose` for the CLOSE_STATEMENT metric.
  - **Tagged CI job** (`CGO_ENABLED=1 -tags databricks_kernel`) — depends on the kernel-lib distribution work (the acknowledged H5 follow-up).
  
  If it'd help, I can link these once the follow-up changes are up for review.
  
  ## Deferred / tracked (now marked inline in the source)
  
  Because the second pass re-flagged items whose rationale lived only in planning notes, each deferred site now carries a `Deferred (tracked)` comment in the code itself:
  
  - **`CloseSession` no-deadline blocking close** — needs a kernel `close_blocking` (with deadline) or a Go-side watchdog; grouped with the kernel C-ABI follow-ups.
  - **`klog` not routed through `logger.Logger`** (no connId/corrId; own `DBSQL_KERNEL_DEBUG` knob) — logging-unification follow-up; also covers the process-wide first-call-wins subscriber note.
  - **Duplicate map keys render non-unique JSON** (`writeMapJSON`) — mirrors the Thrift path exactly (parity holds); dedup-last-vs-error is a cross-backend contract decision for its own PR.
  - **Multi-entry/multi-row map parity** — the parity test now covers this (see above), so this thread is addressed.

internal/backend/kernel/operation.go
vikrantpuppala: Round 3 re-review @ `b8957934`: still open — `kernelOp.ExecutionError` still returns bare `toStatementError(cause)`, so `errors.Is(err, dbsqlerr.ExecutionError)` / `errors.As(err, &dbExecErr)` continue to fail on the kernel path while working on Thrift.

This is the most tractable of the remaining Highs — a one-function wrap mirroring `thriftOperation.ExecutionError` (which does `dbsqlerrint.NewExecutionError(ctx, dbsqlerr.ErrQueryExecution, cause, o.opStatusResp)`). The kernel op has no Thrift `opStatusResp`, so pass `nil`:

```go
func (o *kernelOp) ExecutionError(ctx context.Context, cause error) error {
	return dbsqlerrint.NewExecutionError(ctx, dbsqlerr.ErrQueryExecution, toStatementError(cause), nil)
}
```

That restores sentinel matching (`ExecutionError`/`DatabricksError`) and the `DBExecutionError` methods (`SqlState()`/`QueryId()`). If `NewExecutionError` can't accept a nil status resp cleanly, an equivalent wrapper that carries `KernelError.SQLState`/`QueryID` works too — the key is that the returned error satisfies the same public interface as Thrift.

<sub>— Code Review Squad (round 3 re-review)</sub>

internal/backend/kernel/operation.go
vikrantpuppala: Round 3 re-review @ `b8957934`: verified the round-2 change — it added ctx-error *reporting* after `kernel_statement_execute` returns (`if ctx.Err() != nil { ... return execute cancelled }`), which is a good improvement. But it doesn't fully close the finding: the watcher's `select { case <-ctx.Done(): ... }` can still land in the window *after* execute returns successfully and *before* `close(done)` runs, so `fireCancel` can still dispatch a server-side cancel against an already-completed statement.

Memory-safe (canceller freed after `watcherWg.Wait()`), so this is Low severity — but if you want it fully closed, have `fireCancel` no-op when the statement is already terminal, or re-check the done signal before committing to fire. Flagging for your call; fine to leave deferred.

<sub>— Code Review Squad (round 3 re-review)</sub>

vikrantpuppala:
## [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad) — Round 3 status @ `b8957934`

Strong iteration — **7 of 8 round-2 findings are resolved and verified genuine**, including an excellent nested-offset fix (N2) that also handled the LargeList/FixedSizeList edge cases beyond what the finding named, and the WARN-log-noise fix (N3). Duplicate map keys (F8) is consciously deferred to its own PR with a clear code comment — a defensible call.

**Three Highs remain before `WithUseKernel(true)` is production-ready:**

| # | Finding | Status |
|---|---------|--------|
| **N1** | Kernel exec errors don't satisfy `errors.Is(ExecutionError)` / `errors.As` (Thrift parity break) | Open — one-function wrap; concrete fix posted on the thread |
| **F2** | Kernel queries emit zero per-statement telemetry (`StatementID()==""` + `OnClose` never fired) | Open — needs a kernel query-id via the C ABI |
| **F3** | No `CGO_ENABLED=1 -tags databricks_kernel` CI job — the entire cgo backend runs in no pipeline | Open — acknowledged in the workflow comment |

N1 is the most tractable and worth grabbing now. F2/F3 need C-ABI/infra work, so they're understandably longer-tail — but they're the ones that make the kernel backend safe to actually turn on. None block the default `CGO_ENABLED=0` release.

<sub>— Code Review Squad (round 3 re-review)</sub>

mani-mathur-arch:
Thanks for the thorough multi-round reviews, @vikrantpuppala. Consolidating where things landed. Everything below is pushed to this branch.
  
  ## Fixed on this PR
  
  **Correctness**
  - **Nested-value offset bug**: List/Map were indexed via the un-sliced `Offsets()[row]` instead of the offset-aware `ValueOffsets`; wrong/panicking for sliced arrays (or a List/Map field of a struct carrying a logical offset). Fixed for
  List/Map/LargeList/FixedSizeList, with sliced-array parity cases as a regression guard.
  - **Staging silent no-op**: volume `PUT`/`GET`/`REMOVE` reported success with no file transferred. Now detected from the SQL and rejected at execute time.
  - **One-directional enum-drift guard**: only caught the C value exceeding the Go const; a downward renumber slipped through. Now bidirectional. 
  - **No `ErrBadConn` on the execute/read path**: prevents `database/sql` silently replaying a committed non-idempotent DML, while still evicting a dead session.
  - **Execution-error typing (N1/F1)**: `kernelOp.ExecutionError` now wraps the cause so `errors.Is(err, dbsqlerr.ExecutionError)` / `errors.As(&DBExecutionError)` and `SqlState()`/`QueryId()` behave like the Thrift path (both sqlState and the
  server query id are carried through from `KernelError`).
  - **Cancellation semantics**: a completed statement is never reported as cancelled (matches Thrift; avoids a double-write-on-retry hazard); cancel is honored pre-completion via the watcher, with no spurious cancel against a terminal
  statement.
  
  **Contract / observability**
  - Every unsupported-option rejection wraps `errors.ErrNotSupportedByKernel` (programmatically detectable, not message-matched); the "not-compiled-in" stub wraps `errors.ErrKernelNotCompiled`.
  - `WithTransport`, non-default `WithPort`, and non-https protocol are now rejected at connect (were silently dropped).
  - Kernel-call failures log user faults (bad SQL / bad argument) at Debug and reserve Warn for infra codes, so a fat-fingered query doesn't inflate the WARN rate.
  - Proxy resolves correctly in warehouse-id addressing mode; session id is a process-unique counter (not a reused pointer); `evictIfSessionFatal` uses `errors.As`.
  
  **Tests / CI**
  - **Tagged CI job (F2/F3)**: a `build-and-test-kernel` job now builds the kernel static lib (2-mode recipe, pinned `KERNEL_REV`) and runs the `databricks_kernel`-tagged unit tests on every PR/push. The `cgo && databricks_kernel` path is no
  longer dead in CI. (Unit-only, matching how the Thrift path is treated in this workflow; live e2e runs via the integration-test dispatch for both backends.)
  - Cross-backend parity test hardened (all rows, multi-entry maps, sliced arrays); arrowscan-only tests added for the LargeList/FixedSizeList offset arithmetic, empty-but-non-null collections, and the DECIMAL scale boundary.
  
  ## Deferred to follow-ups (documented, not silently dropped)
  
  These need kernel C-ABI / cross-backend / infra work and are tracked with inline `Deferred (tracked)` markers:
  
  - **Per-statement telemetry**: `StatementID()` is `""` and `OnClose` doesn't fire, so kernel queries emit no EXECUTE_STATEMENT/CLOSE_STATEMENT metrics. Needs a kernel C-ABI success-path query-id accessor.
  - **Session-lifecycle / mid-fetch context deadlines**: `OpenSession`/`CloseSession`/`nextBatch` block in an uninterruptible cgo call and can't honor a ctx deadline mid-call. Needs a kernel-side deadline/cancel.
  - **Duplicate map keys** rendering non-unique JSON: mirrors the Thrift path exactly; changing the emitted grammar is a cross-backend contract decision for its own PR.
  - **Single-sourcing the two Arrow→JSON renderers**: justified duplication (pure-Go / no-cgo constraint), guarded by the parity test.
  - **Integer widths → `int64`**: both backends return native Go widths today; unifying on int64 changes the production Thrift path's return type and needs maintainer sign-off.
  
  ## Scope
  
  This PR has grown well beyond the original opt-in backend as we've worked through the review rounds (correctness, error contract, CI). To keep it reviewable and landable, the fixes above are the scope of this PR; any further changes including the deferred items will be handled in dedicated follow-up PRs. None of the deferred items affect the default pure-Go (`CGO_ENABLED=0`) build, which remains unchanged; they matter before
  running `WithUseKernel(true)` in production.

vikrantpuppala (approved):

.github/workflows/go.yml
mani-mathur-arch: Addressed on the branch. The `build-and-test-kernel` job (commit `5a3c09a`) builds the pinned Rust kernel static lib and runs the `databricks_kernel`-tagged unit tests under `CGO_ENABLED=1` on every PR/push, so the cgo backend (cancel-watcher, per-batch release/EOF, `close()` idempotency) is exercised in CI rather than shipping untested. The `Test (kernel backend)` check is green on the latest commit. Resolving as done.

internal/backend/kernel/operation.go
mani-mathur-arch: Deferred and tracked in PECOBLR-3665 (both halves: the EXECUTE_STATEMENT gate needs a kernel C-ABI success-path query-id accessor to populate `StatementID()`, and the CLOSE_STATEMENT metric needs `callbacks.OnClose` wired from `kernelRows.Close()`). Marked inline with a `Deferred (tracked)` comment at `StatementID()`. No effect on the default `CGO_ENABLED=0` build; matters before running `WithUseKernel(true)` in production. Resolving as tracked.

internal/backend/kernel/backend.go
mani-mathur-arch: Deferred and tracked in PECOBLR-3647 (blocking `kernel_session_close_blocking` with a deadline) and PECOBLR-3646 (session-lifecycle ctx deadlines on Open/Close). The clean fix needs a kernel C-ABI `close_blocking` or a Go-side watchdog; marked inline with a `Deferred (tracked)` comment at `CloseSession`. No effect on the default build. Resolving as tracked.

internal/arrowscan/arrowscan.go
mani-mathur-arch: Deferred and tracked in PECOBLR-3661. This mirrors the Thrift renderer exactly (parity holds and it matches what ships today), so dedup-last-vs-error is a cross-backend grammar contract decision best made deliberately in its own change rather than diverging the two paths here. Marked inline with a `Deferred (tracked)` comment at `writeMapJSON`. Resolving as tracked.

internal/backend/kernel/cgo.go
mani-mathur-arch: The misleading half is fixed on the branch — the doc comment no longer claims a test/benchmark asserts it (there is none yet). Removing the still-unused export is tracked in PECOBLR-3664 (delete it and reintroduce it with the benchmark that asserts `DBSQL_KERNEL_DEBUG` is off before measuring). Resolving with the comment fixed + removal tracked.

internal/backend/kernel/cgo.go
mani-mathur-arch: Deferred and tracked in PECOBLR-3650 (route kernel debug logging through the driver's `logger.Logger`/`DATABRICKS_LOG_LEVEL` knob so lines carry connId/corrId/queryId; also covers the process-wide first-call-wins subscriber note). Marked inline with a `Deferred (tracked)` comment in `cgo.go`. Resolving as tracked.

Issue #162, opened 2023-09-20
Tools.go causes package conflict


andyleap:
When trying to vendor this package into a bazel environment, tools.go package name "tools" is conflicting with the rest of the go files.  This is due to the fact that gazelle generation largely ignores build tags in it's initial phase.  Is there any way that file could either get the package matched or moved into it's own directory?  Happy to make a PR for that either way...

rcypher-databricks:
Thanks for raising this issue.  We will look into it.

cetteup:
We just ran into this issue with a project that happened to use a `tools` build tag internally. Because we have golangci-lint configured to also cover code with that build tag, it runs into a compile error with the two packages in the root of this library (`dbsql`, `tools`).

Unless you specifically need the `tools` build tag, the easiest solution would be to change the tag to the officially documented `ignore`.

PR #396, opened 2026-07-12
feat(kernel): close deferred feature rows — OAuth M2M/U2M, metric-view, initial namespace

mani-mathur-arch:
## What

Closes the four deferred feature rows the SEA-via-kernel backend previously rejected at connect time. All are Go-side only — **no new kernel dependency**: the OAuth setters are on merged kernel #162, metric-view is an existing server session conf, and the initial namespace uses plain SQL.

- **Metric view** — `config.EffectiveSessionParams()` derives the server conf (`spark.sql.thriftserver.metadata.metricview.enabled`) once, backend-neutrally, so both backends send the identical conf. The Thrift `OpenSession` special-case is removed (behaviour-preserving); the kernel forwards it via `SessionConf`. Reject dropped, reclassified forwarded.
- **Initial namespace** — applied post-connect via `USE CATALOG` / `USE SCHEMA` (the OSS ODBC workaround) since the kernel C ABI has no catalog/schema setter. `quoteIdent` (untagged) backtick-quotes identifiers; a `USE` failure fails connect and closes the session (routed through `call()` so the close is logged). Reject dropped, reclassified forwarded.
- **OAuth M2M / U2M** — the kernel drives its own OAuth flow from raw credentials (mirroring pyo3/napi and the Node/Python kernel bindings), read off `cfg.Authenticator` — the single source of truth (last-writer-wins, matching Thrift). The m2m/u2m authenticators expose credential-provider interfaces defined in `internal/backend/kernel` (kept off the public API so a secret-read capability is never exported); `resolveKernelAuth` type-switches them and returns `kernel.Auth` directly. `KernelBackend.setAuth` branches to `set_auth_pat` / `set_auth_m2m` / `set_auth_u2m`. U2M uses Go's cloud-inferred client id (scopes/redirect-port default to the kernel's; no Go option feeds them yet). **No new `config.UserConfig` fields.**

Stacked on #395 (`mani/sea-kernel-build-recipe`); additive only, so #393 and #395 are untouched.

## Commits

1. **feat** — the four features above.
2. **fix** — code-review remediation: moved the OAuth credential-provider interfaces into `internal/` (off the public API); collapsed the duplicate auth descriptor so `validateKernelConfig`/`resolveKernelAuth` return `kernel.Auth` directly; routed the initial-namespace failure-path close through `call()` so it logs; added a live M2M e2e + a last-writer-wins auth regression test; documented U2M interactivity, the dormant U2M scopes/port wiring, and the metric-view smoke test.

## Known limitations (documented, not blocking)

- **OAuth U2M is interactive** — on a token-cache miss, opening a connection launches the system browser and blocks up to the kernel's ~120s callback timeout; the connect context deadline is not honored during that window (the C ABI can't interrupt session open, and exposes no timeout override). Use PAT or M2M for headless/deadline-bound connects.
- **Custom M2M scopes** aren't forwardable over the C ABI (`set_auth_m2m` takes no scopes arg) — a kernel gap shared with ODBC; no authorization impact (`all-apis` is always requested).

## Testing

- Default `CGO_ENABLED=0` suite + `golangci-lint` v2.12.2 — clean repo-wide.
- Tagged `databricks_kernel` unit tests — auth-mode → C-setter mapping (PAT/M2M/U2M), `quoteIdent`, config assembly/classification, error classifier.
- Live staging e2e — PAT connect, per-type data types, CloudFetch, ctx cancellation, query tags / statement timeout / timezone, TLS skip-verify, initial namespace (`current_catalog()`/`current_schema()` match config), metric-view (connect+query smoke — the conf isn't `SET`-introspectable on either backend), Thrift parity.
- M2M/U2M auth resolution covered by unit tests; live M2M skips without a staging service principal, U2M is interactive (not headless-runnable) — auth wiring is unit-verified.
- Reviewed with the multi-reviewer code review + Isaac Review; all findings addressed or documented, final Isaac Review clean (0 findings).

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -9,28 +9,6 @@ type Authenticator interface {
 	Authenticate(*http.Request) error
 }
 
-// M2MCredentialsProvider is implemented by the OAuth M2M authenticator to expose
-// its raw client-credentials. The SEA-via-kernel backend reads these to drive the
-// kernel's own M2M flow (the kernel owns the token exchange), rather than using the
-// authenticator's Authenticate method. This keeps cfg.Authenticator the single
-// source of truth for auth on both backends — the kernel selects M2M by asserting
-// this interface, so the last WithX option applied wins, exactly as on Thrift.
-type M2MCredentialsProvider interface {
-	// M2MCredentials returns the client id and client secret. (The kernel's C-ABI
-	// M2M setter takes no scopes — it applies its own default scope set, matching
-	// the Go authenticator's default — so scopes are not exposed here.)
-	M2MCredentials() (clientID, clientSecret string)
-}
-
-// U2MCredentialsProvider is implemented by the OAuth U2M authenticator to expose
-// the cloud-inferred client id the kernel should use for its browser/PKCE flow (so
-// the kernel path uses the same client id the Thrift path would). See
-// M2MCredentialsProvider for why this rather than a parallel config carrier.
-type U2MCredentialsProvider interface {
-	// U2MClientID returns the OAuth client id for the U2M browser flow.
-	U2MClientID() string
-}
-
 type AuthType int
 
 const (
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -9,6 +9,28 @@ type Authenticator interface {
 	Authenticate(*http.Request) error
 }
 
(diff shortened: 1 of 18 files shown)


mani-mathur-arch:
superseded by #399

PR #397, opened 2026-07-12
feat(kernel): SEA-via-kernel data types, params, errors & query-id telemetry

mani-mathur-arch:
Stacked on #396 (DAIS gap-closure). The PuPr feature set for the SEA-via-kernel backend, as a feature-commit stack. Depends on the kernel PuPr statement surface (databricks-sql-kernel#165) for the two tagged commits.

## Commits
Untagged / no kernel dependency (run in the default `CGO_ENABLED=0` CI):
- `feat(arrowscan): render INTERVAL day-time & year-month` — Duration / MonthInterval → Thrift's server-string form, in the shared untagged scanner.
- `test(kernel): pin TIMESTAMP vs TIMESTAMP_NTZ parity` — both backends shift NTZ via `.In(loc)` (verified live); a parity case guards against a one-sided change.
- `test(kernel): pin VARIANT & GEOMETRY render parity` — both arrive as strings (verified live); nested string-leaf + top-level cases.
- `fix(kernel): surface kernel errors as DBExecutionError with sqlstate` — `errors.As` → `DBExecutionError`/`SqlState()` parity with Thrift (42P01/42703), via a neutral untagged constructor.
- `docs(kernel): document verified inherited features` — pool / telemetry / circuit-breaking inherited above the seam; drops the now-stale interval caveat.

Tagged (`cgo && databricks_kernel`), depend on #165:
- `feat(kernel): bind query parameters via the C ABI` — `backend.Param{Name,Type,Value}` → `kernel_statement_bind_parameter`; 10-case live parity (positional/named, each scalar, NULL, multi-param, predicate).
- `feat(kernel): surface server query id for EXECUTE_STATEMENT telemetry` — `StatementID()` → `kernel_executed_statement_query_id`, ungating EXECUTE_STATEMENT telemetry + QueryIdCallback.
- `build(kernel): pin KERNEL_REV to the PuPr statement-surface rev` — points the build recipe at #165's tip so the tagged CI links the new symbols; re-pin to #165's squash-merge SHA once it lands.

## Testing
Default `CGO_ENABLED=0` suite green; tagged build + unit tests green; every live behavior (intervals, NTZ, variant/geometry, error sqlstate, connection pool, param binding, query-id telemetry) verified against a staging warehouse. `golangci-lint` clean on all touched files.

**Isaac Review:** two batched passes, both landing clean.
- Untagged commits (data types, error sqlstate, docs) — 0 findings. (An earlier per-commit pass on the interval commit caught 1 MAJOR — an int64 overflow scaling large day-time intervals to nanoseconds — fixed by dividing in the native unit before scaling only the sub-second remainder, with a `Long.MaxValue` regression case.)
- Tagged commits (param binding, query-id telemetry) — 0 findings after removing a stale `StatementID()` doc-comment the reviewer flagged.

Several behaviors were pinned by live-probing both backends first rather than coding to assumption: TIMESTAMP_NTZ (both shift via `.In(loc)` — the "don't-shift NTZ" plan premise was wrong and would have broken parity) and VARIANT/GEOMETRY (both arrive as strings).

## Merge order
Kernel #163 → #165, then driver #393 → #395 → #396 → this PR; re-pin `KERNEL_REV` to #165's merge SHA at that point. The tagged/cgo CI links #165's branch tip in the meantime.


mani-mathur-arch:
superseded by #399

PR #399, opened 2026-07-14
feat(kernel): consolidated DAIS gap-closure + PuPr feature set + nightly workflows for e2e tests

mani-mathur-arch:
## What

Extends the SEA-via-kernel backend for the Go driver with the full DAIS gap-closure
+ PuPr feature set: authentication, session setup, type rendering, query execution,
telemetry, richer TLS, kernel-log routing, and CI. 

This PR consolidates what was a 5-PR stack (#400 richer-TLS, #401 log-unify,
#402 nightly-e2e, #403 c2-dispatch) into a single reviewable PR. The stack was linear,
so the branches were fast-forwarded into this one — every change is preserved as a
**distinct, logically-scoped commit** (23 commits), so it reads commit-by-commit rather
than as one blob. #400–#403 now show as merged (their commits live here).

## Features

**Authentication**
- **OAuth M2M / U2M** — the kernel drives its own auth flow from `cfg.Authenticator`
  (mirroring pyo3/napi). Adds a `kernel.Auth` descriptor + `resolveKernelAuth`, wired
  through the `set_auth_pat` / `_m2m` / `_u2m` C-ABI setters. `cfg.Authenticator` is the
  single source of truth (last-writer-wins, matching Thrift).
- **Unsupported auth is sentinel-wrapped** — token-provider / external / federated
  authenticators reject with `%w`-wrapped `ErrNotSupportedByKernel`, the same
  programmatic-fallback contract every other unsupported kernel option follows.

**Session setup**
- **Initial namespace** — post-connect `USE CATALOG` / `USE SCHEMA` with identifier
  quoting (the kernel C ABI has no namespace setter).
- **Metric-view metadata** — `config.EffectiveSessionParams()` folds the server conf in
  backend-neutrally, so the kernel path sends the identical conf Thrift does.

**Type rendering**
- **INTERVAL** day-time & year-month rendering in `internal/arrowscan` (kernel returns
  native arrow values; formatted Go-side to the Thrift path's string form). Handles the
  `math.MinInt64` / `math.MinInt32` negation bound without overflow.
- Parity coverage for **TIMESTAMP vs TIMESTAMP_NTZ** and **VARIANT / GEOMETRY**
  (live-probed on both backends).

**Query execution & telemetry**
- **Bound query parameters** via the kernel raw-param C ABI
  (`kernel_statement_bind_parameter`). The positional/named + SQL-NULL/empty-string
  decision is a pure `paramBindArg` seam, unit-tested under `CGO_ENABLED=0`.
- **Kernel errors surfaced as `DBExecutionError`** carrying sqlstate + server query id,
  so `errors.As` → `SqlState()` / `QueryId()` works on the kernel path. The execute
  path deliberately reports **non-retryable** (a sent statement may have committed — no
  double-write), matching Thrift.
- **Server query id** via a real `StatementID()` accessor
  (`kernel_executed_statement_query_id`), threaded into EXECUTE_STATEMENT telemetry, plus
  CLOSE_STATEMENT telemetry on the result-read path.
- **Cancelled execute** evicts a session-fatal conn and preserves the kernel error
  metadata (multi-`%w`: `errors.Is(DeadlineExceeded)` AND `errors.As(*KernelError)` both
  hold).

**Richer TLS (Go options)**
- CA-bundle + independent hostname-skip setters mapped to the kernel's own TLS knobs
  (defensive PEM copy; the wholesale custom `WithTransport` stays rejected).

**Kernel logging**
- Kernel log output routed through the driver log level with connId/corrId/queryId
  correlation; allocation-free above Debug (guards the Arrow-batch hot path).
- **scope note (please read before flagging):**
- This unifies the *control knob*, not the *sink*, and that boundary is intentional.
`DATABRICKS_LOG_LEVEL`/`SetLogLevel` now drives both the Go binding lines and the
kernel's Rust lines from one level. But they stay two physical streams: Go lines go
through `logger.Logger` (and honor `logger.SetLogOutput`), while the kernel's Rust
lines go straight to stderr — the kernel C ABI exposes **no sink/callback hook**, so
there is currently no way to route them through the Go logger. Consequences a
reviewer will notice, all expected:
- A custom `SetLogOutput`/file captures only the Go lines; Rust lines stay on stderr.
- With the default sink both land on stderr but only *race* there — not a single
  ordered, merged stream.
- The Rust subscriber's level is sampled once, at the first kernel connect.

Fully merging both into one ordered sink requires a kernel log callback and is
tracked separately (<TICKET>); it's out of scope here because it's a kernel-side
ABI change, not a driver one.

**CI**
- A scheduled + dispatchable **nightly E2E** workflow that runs the credential-gated
  e2e + Thrift-vs-kernel parity suites against a real warehouse (the ordinary CI injects
  no warehouse secrets, so those tests otherwise skip). SEA-via-kernel integration leg
  gated behind a kernel label.

## Notable behavior

- Bound query parameters are now supported (previously rejected at connect).
- **Staging operations remain rejected** — the kernel path can't perform the local file
  transfer and the C ABI surfaces no staging signal.

## Testing

- **Default pure-Go** (`CGO_ENABLED=0`): `go build` / `go vet` / `go test ./...` — 24
  packages ok.
- **Tagged kernel path** (`CGO_ENABLED=1 -tags databricks_kernel`, linked against a
  locally-built kernel `.a` at `KERNEL_REV`): 24 packages ok.
- **Live e2e + parity** against a real warehouse: all pass — data types (incl.
  variant/geometry/decimal), NTZ, params-vs-Thrift (10/10), CloudFetch, cancellation,
  connection pool (40/40), StatementID, initial namespace, and **M2M** (OAuth
  client-credentials).
- `gofmt` clean; Isaac Review clean (0 final comments).

## Merge gate

- **Re-pin `KERNEL_REV` at merge time.** It currently points at the kernel dependency's
  PR-head SHA (GC-able once that kernel change merges); bump it to the resulting kernel
  `main` SHA, re-sync the cgo drift assertions in `cgo.go` if any signatures changed, and
  run `make test-kernel` against the new rev.


internal/arrowscan/arrowscan.go
```diff
@@ -175,16 +177,84 @@ func ScanCellCached(col arrow.Array, row int, loc *time.Location, keys *StructKe
...
 	}
 }
 
+// formatDayTimeInterval renders an arrow duration (in the given time unit) as the
+// Thrift path's "D HH:MM:SS.nnnnnnnnn" — days, then zero-padded hours:minutes:seconds
+// with 9 fractional digits, negated with a leading '-'.
+func formatDayTimeInterval(v int64, unit arrow.TimeUnit) string {
+	neg := v < 0
+	if neg {
+		v = -v
```
mani-mathur-arch: This still overflows for the minimum negative interval value. For `math.MinInt64`, `v = -v` wraps and leaves `v` negative, so the formatter then computes negative components while also prepending `-` (the same issue exists for `math.MinInt32` in `formatYearMonthInterval`). These bounds are valid Arrow/Spark interval values, and the current tests cover `MaxInt64` plus small negatives but not the negative extreme. Can we avoid negating the full signed magnitude, or add explicit MinInt handling and regression cases?

kernel_config.go
```diff
@@ -61,51 +60,83 @@ func validateKernelConfig(cfg *config.Config) (token string, err error) {
...
+				token = p.AccessToken
+			}
+		}
+		if token == "" {
+			return kernel.Auth{}, errors.New("databricks: the kernel backend requires a personal access token; " +
+				"set one with WithAccessToken (or a *pat.PATAuth via WithAuthenticator)")
+		}
+		return kernel.Auth{Mode: kernel.AuthPAT, Token: token}, nil
+	default:
+		return kernel.Auth{}, errors.New("databricks: this authenticator is not supported by the kernel backend; " +
```
mani-mathur-arch: This unsupported-auth path no longer wraps `ErrNotSupportedByKernel`, even though `validateKernelConfig` documents that unsupported kernel options use that sentinel so callers can fall back with `errors.Is`. That means token-provider / external / static / federated auth failures behave differently from the other unsupported kernel options. Can this return a `%w`-wrapped `ErrNotSupportedByKernel` (or update the contract/tests if unsupported auth intentionally should not be programmatically detectable)?

kernel_parity_test.go
```diff
@@ -64,12 +64,67 @@ func TestKernelThriftParity(t *testing.T) {
...
+	{"pos_two", "SELECT ? + ? AS v", []any{int64(2), int64(40)}},
+	{"pos_in_predicate", "SELECT count(*) AS v FROM range(100) WHERE id < ?", []any{int64(10)}},
+	{"named_int", "SELECT :n AS v", []any{sql.Named("n", int64(7))}},
+	{"named_string", "SELECT :s AS v", []any{sql.Named("s", "world")}},
+	{"named_two", "SELECT :a AS a, :b AS b", []any{sql.Named("a", int64(1)), sql.Named("b", "x")}},
+}
+
+// TestKernelParamsVsThrift asserts parameterized queries produce byte-identical
+// output on the kernel and Thrift backends — the bound-parameter acceptance gate.
+func TestKernelParamsVsThrift(t *testing.T) {
```
mani-mathur-arch, line 122: This is the main test that proves kernel parameter binding actually works, but it self-skips without `DATABRICKS_HOST` / `DATABRICKS_HTTP_PATH` / `DATABRICKS_TOKEN`, and the kernel CI job intentionally runs without warehouse creds. As a result, regressions in named vs positional binding, NULL vs empty-string handling, type strings, or bind-before-execute ordering can pass required CI. Can we add some executing coverage for the bind mapping (for example a hermetic seam around the Param -> kernel bind arguments, or a required credentialed kernel parity run)?

mani-mathur-arch:
Two additional review findings from the PR pass could not be anchored inline because GitHub does not expose those exact lines in this PR diff:

1. Kernel retryability is still dropped from `DBExecutionError`. `KernelError.Retryable` is copied from the C ABI, but `kernelOp.ExecutionError` only forwards sqlstate/query id into `NewExecutionErrorWithState`; `DBError.IsRetryable()` is derived from `internal/errors.RetryableError` in the cause chain. A retryable kernel execution failure will therefore surface as non-retryable unless the cause is wrapped when `ke.Retryable` is true, with a test asserting `DBExecutionError.IsRetryable()`.

2. The cancelled execute branch skips session-fatal eviction. If `ctx.Err() != nil` while `kernel_statement_execute` returns a session-fatal `KernelError`, the code returns only `ctx.Err()` and bypasses `k.evictIfSessionFatal(execErr)`, leaving a dead session potentially valid in the pool and dropping kernel metadata from the returned error. It should still classify/evict on `execErr` while preserving cancellation semantics.

internal/arrowscan/arrowscan.go
mani-mathur-arch: Fixed in f5ea652. Both formatters now derive each component from the **signed** value and take its magnitude when formatting (`abs64`) instead of negating the full magnitude up front; `formatYearMonthInterval` widens to `int64` before negating so `math.MinInt32` doesn't wrap. Added regression cases for `math.MinInt64` (µs and ns units → `-106751991 04:00:54.775808000` / `-106751 23:47:16.854775808`) and `math.MinInt32` (`-178956970-8`).

kernel_config.go
mani-mathur-arch: Addressed one commit up the stack in #403 (`0ee3a85`), which `%w`-wraps `ErrNotSupportedByKernel` in the unsupported-auth `default` case so callers can detect it with `errors.Is`. Deliberately **not** duplicated here to avoid a conflicting edit on the same lines when #403 rebases onto this branch. Resolving since it's handled in the stack — flagging that it is not in this PR's own diff.

kernel_parity_test.go
mani-mathur-arch: Addressed in f5ea652. Extracted the positional/named + SQL-NULL/empty-string decision out of the cgo `bindParams` into a pure, untagged `paramBindArg` (`internal/backend/kernel/bindparams.go`) that the cgo path now consumes, and added `TestParamBindArg` covering named/positional values, `nil`→SQL NULL, empty-string≠NULL, and positional NULL. It runs under `CGO_ENABLED=0`, so the bind mapping now has PR-time coverage independent of the credentialed nightly parity run (`TestKernelParamsVsThrift`).

mani-mathur-arch:
Both points addressed in f5ea652 — one fixed, one deliberately declined:

**1. Retryability — intentionally NOT changed.** A follow-up Isaac review flagged wrapping the kernel `Retryable` flag here as **MAJOR**: `kernelOp.ExecutionError` is exclusively the post-submission (execute / result-read) path, where a network/unavailable failure may have already committed a non-idempotent INSERT/UPDATE/MERGE. Surfacing `IsRetryable() == true` there would invite an app keying retry on it (the doc.go recipe) to double-write, and it diverges from the Thrift path, which always builds a non-retryable execution error here. This mirrors `toStatementError` refusing `driver.ErrBadConn` for the same reason. Added `TestExecutionErrorNeverRetryable` pinning that a `Retryable` KernelError still reports `IsRetryable() == false`; sqlState/queryId extraction is unchanged. The connect-phase path (`toConnError`), where the retryable signal *is* safe, is unaffected.

**2. Cancelled-execute eviction — fixed.** Hoisted `k.evictIfSessionFatal(execErr)` above the `ctx.Err()` branch so a session-fatal failure racing a cancel still evicts the dead conn from the pool. Also wrapped **both** the ctx error and the kernel error with two `%w` verbs, so `errors.Is(err, context.DeadlineExceeded/Canceled)` still matches **and** the `*KernelError` (sqlstate/queryId) stays reachable via `errors.As` instead of being dropped.

kernel_config.go
mani-mathur-arch: Correcting my earlier reply: this is now fixed **in this PR** (`fe67c65`), not deferred to the stack. The unsupported-auth `default` case wraps `ErrNotSupportedByKernel` with `%w`, matching every other unsupported kernel option and the contract doc.go advertises, and the `non-PAT/non-OAuth authenticator rejected` test now asserts `errors.Is(err, dbsqlerr.ErrNotSupportedByKernel)` rather than only `err != nil`. The empty-PAT case stays unwrapped (missing token is misconfiguration, not an unsupported feature).

You were right that leaving it to #403 was the wrong call — #399 is the base of the stack, so it would have merged to main first and shipped a documented contract its own code broke. The `return` statement is byte-identical to #403's version, so when #403 rebases onto this the overlap is a trivial comment-only conflict and #403's now-redundant `kernel_config.go` hunk can be dropped.

doc.go
```diff
@@ -186,51 +186,70 @@ the C header) under the cgo link path; `make build-kernel` does both steps:
...
+    validation (finer-grained than WithSkipTLSHostVerify).
+
+Setting either without WithUseKernel fails Connect with an error wrapping the
+sentinel ErrRequiresKernelBackend, detectable with errors.Is.
+
+Features above the backend seam are inherited unchanged: the database/sql connection
+pool, per-connection telemetry (CREATE_SESSION / EXECUTE_STATEMENT / DELETE_SESSION),
+and the telemetry circuit breaker. Result types render byte-for-byte identical to the
+Thrift backend: scalars, DECIMAL (exact string), TIMESTAMP / TIMESTAMP_NTZ (shifted
+into the session time zone), INTERVAL, nested Array/Map/Struct and VARIANT (as JSON),
```
mani-mathur-arch, line 251: **[MEDIUM] INTERVAL rendering claimed byte-for-byte identical to Thrift, but has no cross-backend or live test**

*Flagged by 3 reviewers (devils-advocate, test, language).*

This line claims result types — including INTERVAL — "render byte-for-byte identical to the Thrift backend." But the arrow-based parity suite (`internal/rows/arrowbased/arrowscan_parity_test.go`) covers timestamp / variant / geometry / nested-struct and has **no** Duration/MonthInterval case, and the Thrift path can't emit a duration array in prod (native-interval off → server sends a pre-formatted string), so a Thrift-vs-kernel INTERVAL parity is structurally impossible. The only coverage is hand-written `want` strings in `internal/arrowscan/arrowscan_test.go`. If the server's actual interval string differs on a sign/format edge (`-0` prefix, year-month sign placement, fractional digits, day separator), kernel INTERVAL columns render wrong with no CI signal.

<details>

Add a live `TestKernelE2E*` case that `SELECT`s INTERVAL literals (e.g. `INTERVAL '1 01:01:01.5' DAY TO SECOND`, `INTERVAL '1-1' YEAR TO MONTH`) and asserts the scanned string, giving the format the same ground-truth check TIMESTAMP and VARIANT already have.

</details>
---

internal/backend/kernel/operation.go
```diff
@@ -167,10 +189,44 @@ func (k *KernelBackend) execute(ctx context.Context, req backend.ExecRequest) (b
...
+	for i, p := range params {
+		// The positional/named + NULL/empty-string decision is the pure paramBindArg
+		// (unit-tested under CGO_ENABLED=0 in bindparams_test.go); this loop only does
+		// the C-string marshaling of that decision.
+		a := paramBindArg(p)
+		name := newCStrOrNull(a.name) // empty name → NULL → positional
+		typ := newCStr(a.typ)
+		val := newCStrOrNull("") // NULL pointer: the SQL-NULL wire form
+		if !a.isNull {
+			val = newCStr(a.value) // a real value (empty string included), never NULL
```
mani-mathur-arch, line 232: **[MEDIUM] Bound string parameter silently truncated at an embedded NUL byte on the kernel path**

*Flagged by language reviewer.*

`bindParams` marshals a real string value via `newCStr(a.value)` → `C.CString`, which is NUL-terminated. If `kernel_statement_bind_parameter` takes the value as a NUL-terminated `const char*` with no length argument, a parameter value containing an interior NUL (`"a\x00b"`, valid in a Spark STRING) is read Rust-side only up to the first NUL. The Thrift path serializes `StringValue` length-prefixed, so it preserves the full string — a silent value divergence that contradicts the "bound parameters render byte-for-byte identical" contract, with no error raised. `TestKernelParamsVsThrift` does not cover NUL.

Caveat: the no-length-argument claim rests on the vendored kernel header (`databricks_kernel.h`), which is not in this repo — confirm the bind ABI signature before acting.

<details>

Reject a value containing a NUL before binding (parity-safe error), or move to a length-carrying bind ABI. Minimal guard in `bindParams`: `if !a.isNull && strings.IndexByte(a.value, 0) >= 0 { return fmt.Errorf("param %d: value contains a NUL byte, unsupported by the kernel bind ABI", i) }`

</details>
---

internal/backend/kernel/auth.go
```diff
@@ -0,0 +1,65 @@
...
+// interface, so the last WithX option applied wins, exactly as on Thrift.
+//
+// It lives in this internal package (not the public auth package) so the
+// secret-reading capability is never exposed on the driver's public API; the
+// unexported concrete m2m authenticator satisfies it structurally.
+type M2MCredentialsProvider interface {
+	// M2MCredentials returns the client id and client secret. (The kernel's C-ABI
+	// M2M setter takes no scopes — it applies its own default scope set, matching
+	// the Go authenticator's default — so scopes are not exposed here.)
+	M2MCredentials() (clientID, clientSecret string)
```
mani-mathur-arch, line 52: **[MEDIUM] M2M custom OAuth scopes silently dropped on the kernel path**

*Flagged by architecture reviewer.*

`M2MCredentialsProvider` models only `(clientID, clientSecret)`, so `resolveKernelAuth` structurally cannot forward scopes, and the kernel's `set_auth_m2m` applies its own default scope set. A user who supplies custom scopes via the public `WithAuthenticator(m2m.NewAuthenticatorWithScopes(id, secret, host, scopes))` gets those scopes on Thrift but silently the kernel default when they flip `WithUseKernel` — a silent authorization divergence (broader, or a failing token grant, for a least-privilege user). This violates the module's own loudly-advertised "nothing silently ignored" contract; every other unsupported knob is rejected with `ErrNotSupportedByKernel`. The interface comment justifies the omission only for the default-scope case.

<details>

Either widen `M2MCredentialsProvider` to expose scopes (`M2MScopes() []string`) and forward them via a kernel scopes setter, or — if the kernel setter can't accept scopes yet — detect a non-default scope set in `resolveKernelAuth` and reject it with `ErrNotSupportedByKernel`, matching the contract applied to every other unsupported option.

</details>
---

internal/backend/kernel/operation.go
```diff
@@ -138,25 +149,36 @@ func (k *KernelBackend) execute(ctx context.Context, req backend.ExecRequest) (b
...
-		klog("Execute failed: %v", execErr)
-		// A session-fatal status (expired token, dropped/unavailable session) means
-		// this conn is unusable: evict it so the pool doesn't hand it out again. We
-		// still return the PLAIN error (toStatementError, never ErrBadConn), so the
+		klogCtx(ctx, "Execute failed: %v", execErr)
+		// We still return the PLAIN error (toStatementError, never ErrBadConn), so the
 		// conn is discarded without database/sql re-running the statement.
-		k.evictIfSessionFatal(execErr)
 		op.close()
 		return op, fmt.Errorf("kernel: execute: %w", toStatementError(execErr))
```
mani-mathur-arch, line 191: **[MEDIUM] Failed kernel queries emit no EXECUTE_STATEMENT telemetry (asymmetry vs Thrift)**

*Flagged by operational-safety reviewer.*

`op.statementID` is captured only on the success path (the `kernel_executed_statement_query_id` block after a nil `execErr`); this execute-error return leaves it `""`. `connection.go` gates EXECUTE_STATEMENT telemetry (`AfterExecute` / `CompleteStatement`) on `statementID != ""`, so a failed kernel query records no statement metric. Thrift's `StatementID()` reads the `OperationHandle`, which IS present for post-submit failures (e.g. SQL errors), so Thrift emits failure telemetry. The server query id exists here — it rides in `KernelError.QueryID` — but is never copied to `op.statementID`. Result: error-rate dashboards keyed on EXECUTE_STATEMENT undercount kernel failures while the identical Thrift failures are visible, giving on-call a false "healthy" signal.

<details>

On the execute-error path, extract the query id from the `*KernelError` (via `errors.As`) and set `op.statementID` before returning, so the connection-layer telemetry gate fires and `CompleteStatement(..., true)` records the failure with its server query id.

</details>
---

.github/workflows/nightly-e2e.yml
```diff
@@ -0,0 +1,244 @@
...
+      - name: Run kernel E2E
+        env:
+          DATABRICKS_PECOTESTING_SERVER_HOSTNAME: ${{ secrets.DATABRICKS_PECOTESTING_SERVER_HOSTNAME }}
+          DATABRICKS_PECOTESTING_HTTP_PATH2: ${{ secrets.DATABRICKS_PECOTESTING_HTTP_PATH2 }}
+          DATABRICKS_PECOTESTING_TOKEN: ${{ secrets.DATABRICKS_PECOTESTING_TOKEN }}
+          DATABRICKS_PECOTESTING_TOKEN_PERSONAL: ${{ secrets.DATABRICKS_PECOTESTING_TOKEN_PERSONAL }}
+        run: |
+          CGO_ENABLED=1 go test -tags databricks_kernel -count=1 -v -timeout 30m \
+            -run 'TestKernelE2E|TestKernelThriftParity|TestKernelParamsVsThrift' \
+            -skip 'TestKernelE2EM2M' .
```
mani-mathur-arch, line 246: **[MEDIUM] M2M OAuth auth ships with no CI-runnable end-to-end verification**

*Flagged by test reviewer.*

This nightly `-skip 'TestKernelE2EM2M'` (plus the test's own self-skip without `DATABRICKS_CLIENT_ID`/`_SECRET`, and no service-principal secret in the workflow) means the one test that would catch a real M2M auth failure never runs. The unit test `TestSetAuthByMode` asserts only that `set_auth_m2m` returns nil — it can't observe the marshalled args, so a clientID/secret swap or wrong-order bind passes. So M2M authentication, a headline feature, has zero executable coverage; a broken credential mapping surfaces only in a real consumer's failed connect.

<details>

Provision the service-principal secrets and drop the `-skip 'TestKernelE2EM2M'` so the M2M path authenticates against a real warehouse at least nightly. Also tighten the `TestSetAuthByMode` docstring, which overclaims it "pins the marshalling."

</details>
---

.github/workflows/nightly-e2e.yml
```diff
@@ -0,0 +1,244 @@
...
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  thrift-e2e:
+    name: E2E (Thrift backend)
+    # Skips cleanly (no-op success) if the warehouse secret isn't provisioned yet,
+    # so the workflow doesn't hard-fail on a repo where the secrets haven't landed.
+    if: ${{ vars.NIGHTLY_E2E_ENABLED != 'false' }}
```
mani-mathur-arch, line 36: **[LOW] Job-gate comment contradicts the hard-fail credential guard**

*Flagged by 2 reviewers (devils-advocate, operational-safety).*

The comment above this gate claims the job "Skips cleanly (no-op success) if the warehouse secret isn't provisioned yet." But the `if:` keys on the repo variable `vars.NIGHTLY_E2E_ENABLED != 'false'` (which runs by default when the var is unset), not on secret presence — and the "Verify warehouse credentials are present" step `exit 1`s when a secret is missing. So on any repo where secrets haven't landed and nobody has set `NIGHTLY_E2E_ENABLED=false`, the 07:00 UTC cron fails red every night, not a clean skip. The only actual skip path is a manual variable toggle the comment never mentions.

<details>

Either gate on secret presence directly, or reword the comment to state the skip mechanism is the `NIGHTLY_E2E_ENABLED` variable (not automatic on a missing secret).

</details>
---

kernel_e2e_test.go
```diff
@@ -42,11 +47,16 @@ func TestKernelE2ESelect1(t *testing.T) {
 // counterpart to kernelTestDB.
 func kernelTestDBWith(t *testing.T, extra ...ConnOption) *sql.DB {
 	t.Helper()
-	host := os.Getenv("DATABRICKS_HOST")
-	httpPath := os.Getenv("DATABRICKS_HTTP_PATH")
-	token := os.Getenv("DATABRICKS_TOKEN")
+	// Reads the same DATABRICKS_PECOTESTING_* warehouse credentials as the Thrift
+	// E2E suite, so both backends run against one test warehouse from one secret set.
+	host := os.Getenv("DATABRICKS_PECOTESTING_SERVER_HOSTNAME")
```
mani-mathur-arch: **[LOW] Warehouse-credential block copy-pasted into a 3rd/4th test site instead of a shared helper**

*Flagged by maintainability reviewer.*

This 4-line `DATABRICKS_PECOTESTING_*` read + `_TOKEN_PERSONAL` fallback + skip-message block is byte-identical to blocks already in `driver_e2e_test.go` (×2), and the PR duplicates it again here and in `kernel_parity_test.go:16-21`. All four blocks (and their long skip strings) must now be kept in sync — the PR itself edits two of them in lockstep, and a future env-var rename touches four sites.

<details>

Extract `pecoTestingCreds(t) (host, httpPath, token string)` into an untagged `_test.go` in `package dbsql` (visible to both the tagged kernel files and the untagged `driver_e2e_test.go`); call it from all four sites.

</details>
---

mani-mathur-arch:

**Score**: 82/100 — MODERATE RISK

Clean, disciplined 5-PR consolidation of the SEA-via-kernel backend. No Critical or High findings, no injection, no cgo memory defect, and the TLS changes are security-positive (fail-loud when a `WithKernel*` option is set without `WithUseKernel`). Findings cluster on verification gaps and stale docs rather than broken logic: INTERVAL rendering is claimed byte-for-byte identical to Thrift but has no cross-backend or live test, M2M auth has no CI-runnable coverage, failed kernel queries emit no telemetry, and two doc/comment blocks contradict the code they document. None block merge on their own; F1–F5 are worth follow-ups. 9/9 reviewers delivered (language re-spawned after an API cancellation). security and performance found nothing at confidence >= 80.

---

Could not post inline comments for: F6 — see body below.

## Medium Findings

**[LOW] `execute` doc comment contradicts the param-binding it now performs** (F6 — `internal/backend/kernel/operation.go`:L35)

*Flagged by 2 reviewers (agent-compat, devils-advocate).*

The `execute` header still reads "Executes SQL text only; bound parameters are rejected up front by Execute." This PR adds a `bindParams(stmt, req.Params)` call inside `execute` and removed the rejection from `Execute`, so the comment now directly contradicts the code it documents (and the exported `Execute` doc in the same file). The numbered algorithm (steps 1–5) also omits the new bind step between `set_sql` and `canceller_new`. A maintainer trusting "params are rejected here" could reintroduce a rejection guard or mis-reason about the statement lifecycle — exactly the kind of stale invariant a 5-PR consolidation buries.

- **Suggested fix**: Update the comment to state that bound parameters (`req.Params`) are bound onto the statement via `bindParams` before execute, and add the bind as a numbered step in the algorithm.

doc.go
mani-mathur-arch: Addressed in 988bcb9. Added a live TestKernelE2EInterval, a kernel-vs-Thrift parity check on INTERVAL (day-time + year-month, each positive and negative). Thrift gets the server-formatted string, the kernel formats the native arrow value Go-side, and the two must match, so the sign/format edges now have a real ground-truth check. Passed live against a real warehouse.

internal/backend/kernel/operation.go
mani-mathur-arch: Fixed in 988bcb9. Confirmed the bind ABI against the vendored header (value is a NUL-terminated const char* with no length arg), so this is real. Added checkParamValue in the untagged bindparams.go (unit-tested under CGO_ENABLED=0) that rejects a bound value containing an interior NUL before binding, a parity-safe error rather than a silent truncation.

internal/backend/kernel/auth.go
mani-mathur-arch: Fixed in 988bcb9. Widened M2MCredentialsProvider with M2MScopes(); resolveKernelAuth now rejects a non-default scope set with ErrNotSupportedByKernel (the kernel set_auth_m2m takes no scopes and applies all-apis itself), instead of silently downgrading. doc.go updated and a reject test added.

internal/backend/kernel/operation.go
mani-mathur-arch: Fixed in 988bcb9. The execute-error path now sets op.statementID from the KernelErrors query id (statementIDFromError, untagged + unit-tested), so the conns StatementID()-gated EXECUTE_STATEMENT telemetry fires on failures too, matching Thrift.

.github/workflows/nightly-e2e.yml
mani-mathur-arch: Intentionally left as-is. The M2M service-principal secret is deliberately kept off this PUBLIC repo: the PAT already exercises the real-warehouse path for both backends, the M2M setter wiring is unit-covered, and M2M authenticates end-to-end in local/nightly-adjacent runs. Automated M2M coverage, if wanted, belongs in the internal driver-test dispatch, not the public repos secret surface. Tightened the TestSetAuthByMode wording separately.

.github/workflows/nightly-e2e.yml
mani-mathur-arch: Fixed in 988bcb9. Reworded the comment: the skip mechanism is the NIGHTLY_E2E_ENABLED repo variable, and the credential step hard-fails by design (a silent skip reporting green is exactly the gap this workflow closes), not a skip-on-missing-secret gate.

kernel_e2e_test.go
mani-mathur-arch: Fixed in 988bcb9. Extracted pecoTestingCreds(t) into an untagged _test.go in package dbsql and called it from all four sites (driver_e2e_test.go x2, kernel_e2e_test.go, kernel_parity_test.go).

internal/backend/kernel/rows.go
```diff
@@ -56,7 +61,14 @@ type kernelRows struct {
 // newKernelRows fetches the schema up front (for Columns()) and returns the row
 // iterator; batches are pulled lazily on Next.
 func newKernelRows(ctx context.Context, op *kernelOp, stream *C.kernel_result_stream_t, cb *dbsqlrows.TelemetryCallbacks) (driver.Rows, error) {
```
vikrantpuppala, line 63: **Medium — Failed result-schema fetch leaves EXECUTE_STATEMENT dangling; comment claims otherwise**

When `op.Results()` errors (schema fetch/import fails **after** a successful server execute), `QueryContext` returns `return rows, err` (connection.go:378-383) with no `AfterExecute`/`CompleteStatement` call. The completing EXECUTE_STATEMENT telemetry lives only in `closeCallback`, which is armed on the `Rows` — but on this path no `Rows` is armed (callbacks are deliberately left unset until construction succeeds), so the callback never fires.

Result: EXECUTE_STATEMENT is `BeforeExecute`'d + `FinalizeLatency`'d but never completed — a dangling, never-finalized statement metric for a query that executed server-side but whose result-schema fetch failed. This contradicts the `newKernelRows` comment: *"the construction error itself is surfaced to and recorded by the conn execute path."*

Caveat: this asymmetry is shared with the Thrift path (same `op.Results()` → `return rows, err` shape), and the trigger (schema fetch/import failing after a committed execute) is a narrow edge case.

<details>

on the `op.Results()` error path in `QueryContext`, finalize the metric (`AfterExecute` + `CompleteStatement(err)`) the same way the execute-error path at connection.go:282-291 does; and correct the `newKernelRows` comment to match actual behavior.

</details>
---

vikrantpuppala:

**Score**: 92/100 — LOW RISK

Consolidated kernel-backend feature PR (39 files, 3003 lines), reviewed by 9 parallel reviewers. Code is exceptionally well-documented; boundary cases are explicitly handled and unit-tested. No correctness or security defects that block merge. Posting the one Medium finding below; the remaining Low findings are edge-case/observability nits (several documented as deliberate, tracked deferrals) and are not posted.

---

kernel_config.go
```diff
@@ -61,51 +60,101 @@ func validateKernelConfig(cfg *config.Config) (token string, err error) {
...
+		clientID, clientSecret := a.M2MCredentials()
+		return kernel.Auth{Mode: kernel.AuthM2M, ClientID: clientID, ClientSecret: clientSecret}, nil
+	case kernel.U2MCredentialsProvider:
+		// Go sources only the client id for U2M; kernel.Auth.Scopes / RedirectPort
+		// are left zero so setAuth passes the kernel's defaults. Go exposes no
+		// user-facing option for U2M scopes or redirect port on either backend today
+		// (the native Thrift path hardcodes both), so there is nothing to forward —
+		// but the kernel.Auth fields + setAuth wiring model the full set_auth_u2m
+		// surface for if/when such an option is added (see kernel.Auth docs).
+		return kernel.Auth{Mode: kernel.AuthU2M, ClientID: a.U2MClientID()}, nil
```
msrathore-db, line 166: **[HIGH] OAuth U2M requests different scopes on kernel than on Thrift — silent, backend-specific auth failure**

The M2M path guards against scope divergence (`M2MScopesSupported` rejects a custom set), but the U2M branch forwards only `ClientID`, leaving `kernel.Auth.Scopes` nil → the kernel applies its own default scopes. The Thrift path unconditionally requests `offline_access` + `sql` (or, on Azure, `<tenant>/user_impersonation`) via `oauth.GetScopes` inside `GetConfig` (`auth/oauth/u2m/u2m.go:15`). On a workspace whose OAuth app grants only `sql`, kernel U2M fails (or over-requests) while Thrift U2M succeeds — contradicting the `doc.go` claim that kernel U2M uses "the same client id the Thrift path would." `U2MCredentialsProvider` exposes no `U2MScopes()` and the authenticator doesn't store its scopes, so the gap can't be closed without an interface change — yet nothing rejects the mismatch the way M2M does.

<details>

extend `U2MCredentialsProvider` with a `U2MScopes()` accessor and forward the resolved scopes into `kernel.Auth.Scopes`, or reject U2M-over-kernel with an explicit error until scope parity exists (mirroring the M2M custom-scope rejection).

</details>
---

kernel_config.go
```diff
@@ -61,51 +60,101 @@ func validateKernelConfig(cfg *config.Config) (token string, err error) {
...
+// satisfy structurally:
+//   - implements M2MCredentialsProvider → M2M (client id + secret)
+//   - implements U2MCredentialsProvider → U2M (browser/PKCE; kernel-owned flow)
+//   - PAT / nil / noop                  → PAT (from AccessToken or a *pat.PATAuth)
+//   - anything else                     → rejected loudly (token-provider / external
+//     / static / federated), so the failure names the cause instead of surfacing as
+//     an opaque Unauthenticated.
+func resolveKernelAuth(cfg *config.Config) (kernel.Auth, error) {
+	switch a := cfg.Authenticator.(type) {
+	case kernel.M2MCredentialsProvider:
```
msrathore-db, line 148: **[MEDIUM] Real OAuth authenticators bind to the kernel provider interfaces structurally with no compile-time assertion — a method rename silently disables M2M/U2M**

`authClient.M2MCredentials/M2MScopes` and `u2mAuthenticator.U2MClientID` satisfy `kernel.M2MCredentialsProvider` / `U2MCredentialsProvider` only structurally — there is no `var _ kernel.M2MCredentialsProvider = (*authClient)(nil)` assertion anywhere, and the unit tests bind `fake*Auth` types, never the real ones. The one live test (`TestKernelE2EM2M`) self-skips in CI. Renaming/reshaping a real method compiles clean and `go test ./...` stays green, but at runtime the authenticator no longer matches its `case`, falls to `default`, and every OAuth-over-kernel connect is rejected as unsupported.

<details>

add compile-time assertions (`var _ kernel.M2MCredentialsProvider = (*authClient)(nil)` and the U2M equivalent) in a package importing both, or drive `resolveKernelAuth` with the real `m2m`/`u2m` constructors in a unit test rather than fakes.

</details>
---

internal/backend/kernel/cgo.go
```diff
@@ -34,78 +34,121 @@ package kernel
...
+// re-scope or turn it off afterward. That is a kernel-ABI property, not a Go one.
+// A direct consequence: the driver level is sampled HERE, once, at the first kernel
+// session — a later dbsql.SetLogLevel re-levels the Go binding lines (klog/klogCtx
+// re-read GetLevel per call) but NOT the already-installed Rust subscriber. Set the
+// level before opening the first kernel connection to govern the Rust logs.
 func initKernelLogging() {
-	if !kdebug {
-		return
-	}
 	initLoggingOnce.Do(func() {
```
msrathore-db, line 136: **[MEDIUM] Kernel Rust log subscriber now installed unconditionally at WARN, to raw stderr, bypassing `logger.SetLogOutput`**

Previously the subscriber installed only when `DBSQL_KERNEL_DEBUG` was set (`if !kdebug { return }`, now removed); `initLoggingOnce.Do` now fires on the first `OpenSession` in every configuration, installing a process-wide, first-call-wins Rust subscriber at the mapped driver level (WARN default). Consequences: (1) kernel WARN/ERROR now prints to stderr for all users, bypassing a configured `SetLogOutput` sink; (2) a host process using its own in-process Rust `tracing` stack silently loses/conflicts (failed install only logged at Warn); (3) the level is sampled once — a later `SetLogLevel` can't re-quiet it.

<details>

consider mapping an unset driver level to OFF unless explicitly opted in, and prominently document the stderr split-sink limitation in the runbook.

</details>
---

kernel_backend.go
```diff
@@ -29,19 +29,32 @@ func newKernelBackend(_ context.Context, cfg *config.Config) (backend.Backend, e
...
 	// TLS: the driver honors TLSConfig only for InsecureSkipVerify (see
 	// internal/client), so map exactly that knob to the kernel.
 	if cfg.TLSConfig != nil && cfg.TLSConfig.InsecureSkipVerify {
 		kc.TLSSkipVerify = true
 	}
+	// Experimental kernel-only TLS knobs (WithKernelTrustedCerts /
+	// WithKernelSkipHostnameVerify), if any. These have no Thrift-path equivalent
+	// (the connector rejects them on that path) and are forwarded verbatim to the
+	// kernel C ABI in OpenSession.
+	if ke := cfg.KernelExperimental; ke != nil {
```
msrathore-db: **[MEDIUM] `KernelExperimental` TLS forwarding in `newKernelBackend` has no test**

Nothing asserts that `newKernelBackend` actually copies `cfg.KernelExperimental.{TLSTrustedCertsPEM,TLSSkipHostnameVerify}` into `kernel.Config`. Tests cover option→config, the cgo setters with a hand-built `Config`, and Thrift rejection — but deleting these two forwarding lines would silently pass all unit tests, dropping a caller's `WithKernelTrustedCerts` / `WithKernelSkipHostnameVerify`. No e2e exercises them either.

<details>

add a test asserting a `Config` built by `newKernelBackend` from a `cfg` with `KernelExperimental` set carries both TLS fields.

</details>
---

.github/workflows/trigger-integration-tests.yml
```diff
@@ -261,9 +284,14 @@ jobs:
                 pr_repo: context.repo.owner + '/' + context.repo.repo,
                 pr_url: `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/pull/${prNumber}`,
                 proxy_mode: 'replay',
+                // The required merge-queue gate runs the Thrift backend only. The
+                // kernel (sea) leg is previewed on demand via the kernel labels but
+                // is not a required gate — it becomes one once the SEA backend ships
+                // in a released driver and its recordings are captured.
+                go_mode: 'thrift',
```
msrathore-db, line 291: **[LOW] Kernel↔Thrift parity tests don't run on the required merge gate**

The merge-queue required gate hardcodes `go_mode: 'thrift'`; kernel runs only via the opt-in `integration-test-kernel` label or the nightly cron (`TestKernelThriftParity`, `TestKernelParamsVsThrift`). The byte-for-byte parity guarantee `doc.go` advertises rests on live tests excluded from the required merge path, so a Go-side kernel divergence (interval formatter, param mapping) can merge and is caught only on the next nightly.

<details>

consider adding a kernel replay/parity leg to the required gate once committed recordings exist.

</details>
---

.github/workflows/nightly-e2e.yml
```diff
@@ -0,0 +1,246 @@
...
+        run: |
+          if ! command -v make &> /dev/null ; then
+            apt-get update && apt-get install -y make
+          fi
+          go get -v -t -d ./...
+
+      # Guard: fail loudly if a credential is missing rather than letting the tests
+      # silently t.Skip() into a misleading green. The token may be provided as either
+      # _TOKEN or _TOKEN_PERSONAL (the tests accept either), so require at least one.
+      - name: Verify warehouse credentials are present
```
msrathore-db, line 65: **[LOW] Warehouse-credential guard duplicated byte-for-byte across both nightly-e2e jobs**

The "Verify warehouse credentials are present" step (env block + guard + `::error::` text) is copy-pasted into both `thrift-e2e` and `kernel-e2e` (also at lines ~204-214). Adding/renaming a required secret requires editing both; a one-sided edit drifts silently. The repo already uses composite actions (`./.github/actions/setup-jfrog`).

<details>

extract the guard into `.github/actions/verify-warehouse-creds/action.yml` and reference it from both jobs.

</details>
---

msrathore-db:

**Score**: 75/100 — MODERATE RISK

Large PR (39 files, ~3,200 lines) landing the full DAIS gap-closure feature set on the opt-in SEA-via-kernel backend. The kernel↔Thrift abstraction boundary is clean, pure decision logic is well-isolated and unit-tested under `CGO_ENABLED=0`, cgo lifetimes/frees check out, and the hot path stays allocation-free. Remaining risk is kernel↔Thrift behavioral parity — one confirmed auth-scope divergence plus several parity/observability gaps that hermetic tests and the required merge gate don't cover. No security, correctness, or performance blockers; all 10 prior-review findings remain resolved.

---

Could not post inline comments for: F2, F6 — see body below.

## Medium Findings

**F2** · `internal/backend/kernel/backend.go`:L104

**[MEDIUM] U2M session-open blocks in cgo up to ~120s, ignoring the connect-context deadline — can wedge the `sql.DB` pool**

`OpenSession` checks `ctx.Err()` only at entry; once inside `kernel_session_open` → the kernel-owned browser/PKCE flow, the goroutine parks in cgo until login or the kernel's ~120s callback timeout, and `context` cancellation has no effect (documented at `doc.go:225-228`). Under `sql.DB`, a process misconfigured with `authType=oauthU2M` can pile up connection attempts each blocked ~120s in cgo, with no log line naming the interactive-auth stall.

- **Suggested fix**: add a Go-side watchdog that logs at Warn when the connect deadline fires, or reject U2M when the connect ctx deadline is shorter than the kernel callback timeout so the failure names the cause instead of hanging.

## Low Findings

<details>
<summary>Low findings (1) — click to expand</summary>

**F6** · `internal/backend/kernel/operation.go`:L51

**[LOW] Query text not checked for interior NUL — silently truncated on kernel while Thrift sends it whole**

`bindParams` got an explicit interior-NUL guard (`checkParamValue`/`errParamNUL`, `bindparams.go:67`) precisely because the kernel C ABI is NUL-terminated and would truncate silently vs. Thrift's length-prefixed wire form. The identical reasoning applies to `newCStr(req.Query)` (SQL text), which has no such check — a NUL-containing statement is silently truncated on kernel, executing a different, shorter statement than Thrift.

- **Suggested fix**: apply the same interior-NUL check to `req.Query` before `set_sql`, failing loudly rather than diverging from Thrift.



</details>

.github/workflows/nightly-e2e.yml
mani-mathur-arch: Leaving it for a small follow-up rather than this PR as it does not break anything

.github/workflows/trigger-integration-tests.yml
mani-mathur-arch: Acknowledged and intentional for now as the kernel replay recordings do not exist yet as the features are still being implemented. the nighty and the integration tests are used for testing along with local testing

internal/backend/kernel/rows.go
mani-mathur-arch: Fixed in a7228f6 . On the op.Results() error path in QueryContext, we now finalize the metric instead of returning early, the completing telemetry previously lived only in closeCallback, which never fires when no Rows is armed. We use the detached telemetryCtx, not the caller's ctx, because a Results-after-execute failure is often caused by ctx cancellation and the export threads that ctx to the metric's HTTP request, the raw ctx would abort the very metric we're adding. Returns nil rows so the discarded Rows can't later double-fire. The newKernelRows comment is now accurate as written (the error is surfaced and recorded).

kernel_backend.go
mani-mathur-arch: Fixed in a7228f6. Extracted the config assembly into a pure, untagged buildKernelConfig seam so the KernelExperimental TLS forwarding is asserted under CGO_ENABLED=0 (TestBuildKernelConfig) deleting the forwarding lines now fails a test rather than silently passing.

kernel_config.go
mani-mathur-arch: Addressed as a doc fix in a7228f6. The scope difference is real (kernel requests all-apis+offline_access, Thrift sql+offline_access), but neither backend exposes a U2M-scopes option, so it's a fixed-vs-fixed default difference, not a dropped caller choice. Both use the same built-in databricks-sql-connector public client, which all-apis targets so the "Thrift succeeds / kernel fails" outcome depends on a server-side client registration that isn't reproducible from a custom sql-only app. Rather than forward sql (which would swap the kernel onto an untested scope set), I've corrected the overclaim: the U2MClientID comment and doc.go now state the client id is shared but the default scopes differ. If a U2M-scopes option is ever added, kernel.Auth.Scopes is already wired through set_auth_u2m to close it. Could take the complete fix with tests as a follow up.

internal/backend/kernel/cgo.go
mani-mathur-arch: Not changing this here. the substantive part is fixed one PR up the stack in #411. #411 swaps kernel_init_logging (stderr) for kernel_set_log_callback, routing kernel logs through the driver's logger.Logger and honoring SetLogOutput which resolves the "bypasses the configured sink" consequence. The residual items (process-global first-call-wins, level sampled once) are kernel-ABI properties, not fixable Go-side, and are documented in both PRs. Installing at WARN by default is intentional: WARN/ERROR should surface the problem was the destination, which #411 fixes. Re-gating here would just conflict with #411's rewrite of these lines.

kernel_config.go
mani-mathur-arch: Fixed in a7228f6. Added TestResolveKernelAuthRealAuthenticators, which drives resolveKernelAuth with the real m2m/u2m constructors (not the fakes) so a rename of M2MCredentials/M2MScopes/U2MClientID now fails the build/test instead of only breaking at runtime. The test-in-dbsql form covers both. Also verified live: M2M-over-kernel authenticates end-to-end against a real warehouse. 
The tests were skipped as I did not know if we should add the m2m creds in the repo as no other tests were using them so that is the only test that is done locally and not exercised in ci or the nightly runs.

vikrantpuppala (approved):

PR #400, opened 2026-07-14
feat(kernel): richer TLS Go options — CA bundle + independent hostname skip

mani-mathur-arch:
## Context for reviewers

Part of the **SEA-via-kernel** backend for the Go driver: an alternative to the Thrift backend that talks to Databricks SQL through the Rust `databricks-sql-kernel` static library over a C ABI (built into the driver behind the `databricks_kernel` build tag; the default pure-Go build is unaffected). This PR adds two TLS knobs to that backend.

**Stack position** (each stacked on the one below):
```
  └─ #400  richer TLS options            ← THIS PR
            └─ #402  nightly E2E workflow
```

**How to read the diff:** the new public surface is the two `With…` options in `connector.go` + their `config` fields; the interesting logic is (a) the Thrift-path *reject* and (b) the `OpenSession` forwarding via the new `cBytes` cgo helper. Everything else is wiring + tests.

---

## What

Exposes two kernel-only TLS options via an experimental-option idiom — **no kernel change** (both C-ABI setters already exist on kernel main, verified against the `origin/main` kernel header):

- **`WithKernelTrustedCerts(pem []byte)`** → `kernel_session_config_set_tls_trusted_certs`: adds a PEM CA bundle to the kernel's trust store on top of the system roots (corporate re-signing proxy / on-prem CA). Required rather than relying on `SSL_CERT_FILE`, which the kernel's rustls stack ignores.
- **`WithKernelSkipHostnameVerify()`** → `kernel_session_config_set_tls_skip_hostname_verification`: skips only the certificate hostname check while keeping chain validation (finer-grained than the blanket `WithSkipTLSHostVerify`).

Closes **PECOBLR-3651**.

## Design

- Knobs live on a non-exported `config.KernelExperimentalConfig` off `config.Config` (NOT `UserConfig`), so they stay off the stable DSN/exported surface — same treatment as `TLSConfig`, mirroring Node's non-exported `InternalConnectionOptions`. `DeepCopy` copies the CA byte slice.
- The default (Thrift) backend **rejects a non-nil block loudly** at connect, so a caller who sets one and forgets `WithUseKernel` learns the option had no effect rather than connecting with a weaker-than-intended trust store.
- `OpenSession` forwards each to the kernel C ABI via a new byte-buffer helper (`cBytes`, mirroring `cStr`). A reflective guard (`TestKernelExperimentalFieldsClassified`) fails if a new experimental field isn't classified forwarded/rejected — so a future option can't silently be neither forwarded nor rejected.

## Scope boundary (deliberately excluded, deferred)

Bundle 1b only (the no-kernel-dependency part of the richer-TLS work). The following were confirmed **absent from kernel main**, so they need a kernel C-ABI setter first and are correctly deferred:

- **mTLS** client cert/key (`set_tls_client_certificate`) → PECOBLR-3652 (K5).
- **CloudFetch enable/disable** (`set_cloudfetch_enabled`) → PECOBLR-3653 (K3).
- **Private-key auth** → PECOBLR-3605 (crate-blocked).

## Testing

- Default `CGO_ENABLED=0`: build + vet + `go test ./...` — 24 pkg ok, 0 fail.
- Tagged `CGO_ENABLED=1 -tags databricks_kernel` (linked against a locally-built kernel `.a`): build + vet + test — 24 pkg ok, 0 fail. New `TestSetKernelTLS` drives the real cgo setters; untagged `kernel_experimental_test.go` covers the option→config wiring, the classification guard, and `DeepCopy`.
- **Isaac Review: clean** (0 critical / 0 major / 0 minor). Two info-level items were reviewed and judged harmless: the empty→nil `[]byte` normalization, and the idempotent double hostname-skip call when both `WithSkipTLSHostVerify` + `WithKernelSkipHostnameVerify` are set.

## Stacking / merge order

Stacked on **#399** (`mani/sea-kernel-consolidated`). No `KERNEL_REV` change here (inherits #399's pin). #401 (kernel logging) is stacked on top of this.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"database/sql/driver"
-	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -16,6 +16,7 @@ import (
 	"github.com/databricks/databricks-sql-go/auth/pat"
 	"github.com/databricks/databricks-sql-go/auth/tokenprovider"
 	"github.com/databricks/databricks-sql-go/driverctx"
+	dbsqlerr "github.com/databricks/databricks-sql-go/errors"
 	"github.com/databricks/databricks-sql-go/internal/backend"
 	"github.com/databricks/databricks-sql-go/internal/backend/thrift"
 	"github.com/databricks/databricks-sql-go/internal/client"
@@ -50,9 +51,9 @@ func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
 		// learns the option had no effect instead of connecting with a
 		// weaker-than-intended (or unconfigured) TLS trust store.
 		if c.cfg.KernelExperimental != nil {
-			return nil, errors.New("databricks: the WithKernel* options " +
-				"(WithKernelTrustedCerts / WithKernelSkipHostnameVerify) require the kernel backend; " +
-				"add WithUseKernel(true) or remove them")
+			return nil, fmt.Errorf("databricks: a WithKernel* option "+
+				"(WithKernelTrustedCerts / WithKernelSkipHostnameVerify) %w; "+
+				"add WithUseKernel(true) or remove it", dbsqlerr.ErrRequiresKernelBackend)
 		}
 		be, err = thrift.New(ctx, c.cfg, c.client)
 	}
@@ -576,13 +577,25 @@ func kernelExperimental(c *config.Config) *config.KernelExperimentalConfig {
 // EXPERIMENTAL, kernel-only: the default (Thrift) backend rejects this at connect.
 func WithKernelTrustedCerts(pem []byte) ConnOption {
 	return func(c *config.Config) {
-		kernelExperimental(c).TLSTrustedCertsPEM = pem
+		// Copy defensively (matching KernelExperimentalConfig.DeepCopy) so a
+		// caller mutating pem between NewConnector and Connect can't change the
+		// trust store out from under us.
+		if len(pem) > 0 {
+			kernelExperimental(c).TLSTrustedCertsPEM = append([]byte(nil), pem...)
+		} else {
+			kernelExperimental(c).TLSTrustedCertsPEM = pem
+		}
 	}
 }
 
 // WithKernelSkipHostnameVerify skips only the certificate hostname-vs-SNI check on
 // the kernel backend, while keeping chain validation. This is finer-grained than
 // WithSkipTLSHostVerify, which relaxes both chain and hostname checks.
+// WARNING:
+// Skipping hostname verification still weakens TLS: a certificate issued by a
+// trusted CA for a different host will be accepted, opening a machine-in-the-middle
+// vector. Only use this when the hostname is an internal private-link hostname that
+// legitimately differs from the certificate's subject.
 //
 // EXPERIMENTAL, kernel-only: the default (Thrift) backend rejects this at connect.
 func WithKernelSkipHostnameVerify() ConnOption {
--- a/connector.go
+++ b/connector.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"database/sql/driver"
+	"errors"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -43,6 +44,16 @@ func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
 	if c.cfg.UseKernel {
 		be, err = newKernelBackend(ctx, c.cfg)
 	} else {
+		// The experimental WithKernel* TLS options have no Thrift-path equivalent —
+		// reject them loudly rather than silently ignore, so a caller who sets a
+		// trusted-CA bundle / an independent hostname skip and forgets WithUseKernel
+		// learns the option had no effect instead of connecting with a
+		// weaker-than-intended (or unconfigured) TLS trust store.
+		if c.cfg.KernelExperimental != nil {
+			return nil, errors.New("databricks: the WithKernel* options " +
+				"(WithKernelTrustedCerts / WithKernelSkipHostnameVerify) require the kernel backend; " +
+				"add WithUseKernel(true) or remove them")
+		}
 		be, err = thrift.New(ctx, c.cfg, c.client)
 	}
 	if err != nil {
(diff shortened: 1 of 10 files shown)


connector.go
```diff
@@ -43,6 +44,16 @@ func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
 	if c.cfg.UseKernel {
 		be, err = newKernelBackend(ctx, c.cfg)
 	} else {
+		// The experimental WithKernel* TLS options have no Thrift-path equivalent —
+		// reject them loudly rather than silently ignore, so a caller who sets a
+		// trusted-CA bundle / an independent hostname skip and forgets WithUseKernel
+		// learns the option had no effect instead of connecting with a
+		// weaker-than-intended (or unconfigured) TLS trust store.
+		if c.cfg.KernelExperimental != nil {
+			return nil, errors.New("databricks: the WithKernel* options " +
```
mani-mathur-arch: **nit (optional): programmatic sentinel**

Kernel rejections wrap `ErrNotSupportedByKernel` so callers can fall back without matching message text. This Thrift-side rejection uses a plain `errors.New`.

Probably fine for experimental options, but if you expect kernel-first / Thrift-fallback callers, consider a symmetric sentinel (e.g. `ErrRequiresKernelBackend`) wrapped with `%w` here — same detection pattern as `doc.go` documents for the kernel path.

connector.go
```diff
@@ -537,3 +548,45 @@ func WithFederatedTokenProviderAndClientID(baseProvider tokenprovider.TokenProvi
...
+
+// WithKernelTrustedCerts adds a PEM CA-certificate bundle to the kernel's TLS
+// trust store on top of the system roots — for a corporate re-signing proxy or an
+// on-prem CA. Required (rather than relying on SSL_CERT_FILE) because the kernel's
+// rustls stack does not read that environment variable.
+//
+// EXPERIMENTAL, kernel-only: the default (Thrift) backend rejects this at connect.
+func WithKernelTrustedCerts(pem []byte) ConnOption {
+	return func(c *config.Config) {
+		kernelExperimental(c).TLSTrustedCertsPEM = pem
```
mani-mathur-arch: **nit: slice aliasing**

`DeepCopy` defensively copies the PEM bytes for pooled connections, but this stores the caller's slice by reference. If the caller mutates `pem` between `NewConnector` and `Connect`, behavior can change.

Consider copying on set to match `DeepCopy`:

```go
if len(pem) > 0 {
    kernelExperimental(c).TLSTrustedCertsPEM = append([]byte(nil), pem...)
} else {
    kernelExperimental(c).TLSTrustedCertsPEM = pem
}
```

(Separately: `WithKernelTrustedCerts(nil)` / `[]byte{}` still allocates a non-nil `KernelExperimental`, so Thrift connect fails even though `applyKernelTLS` is a no-op — edge case, but worth being aware of.)

connector.go
```diff
@@ -537,3 +548,45 @@ func WithFederatedTokenProviderAndClientID(baseProvider tokenprovider.TokenProvi
...
+	}
+}
+
+// WithKernelSkipHostnameVerify skips only the certificate hostname-vs-SNI check on
+// the kernel backend, while keeping chain validation. This is finer-grained than
+// WithSkipTLSHostVerify, which relaxes both chain and hostname checks.
+//
+// EXPERIMENTAL, kernel-only: the default (Thrift) backend rejects this at connect.
+func WithKernelSkipHostnameVerify() ConnOption {
```
mani-mathur-arch, line 601: **nit: security warning consistency**

`WithSkipTLSHostVerify` carries an explicit MITM warning. This option weakens TLS too (hostname only, but still). A one-liner `WARNING:` comment here would help callers understand the trade-off, similar to the existing option above.

kernel_experimental_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+// The experimental WithKernel* options are rejected on the default (Thrift) path
+// so a caller who forgets WithUseKernel learns the option had no effect. The
+// option builders set config.KernelExperimental; the connector's backend-selection
+// branch is what rejects it. We assert the option→config wiring here (a non-nil
+// KernelExperimental after applying a WithKernel* option is the signal the Thrift
+// branch keys off).
+func TestWithKernelTLSOptionsSetExperimental(t *testing.T) {
```
mani-mathur-arch, line 57: **suggestion: add end-to-end Thrift rejection test**

The file comment says this covers the "Thrift fail-loud signal", but this test only asserts option → config wiring. It doesn't call `Connect()` on the Thrift path.

`TestKernelBackendNotCompiledIn` is the model — a small untagged test like:

```go
func TestWithKernelOptionsRejectedOnThriftPath(t *testing.T) {
    c, err := NewConnector(
        WithServerHostname("example.cloud.databricks.com"),
        WithPort(443),
        WithHTTPPath("/sql/1.0/endpoints/abc"),
        WithAccessToken("token"),
        WithKernelTrustedCerts([]byte("ca")),
    )
    if err != nil { t.Fatal(err) }
    _, err = c.Connect(context.Background())
    if err == nil {
        t.Fatal("expected Connect to reject WithKernel* on Thrift path")
    }
}
```

would lock in the behavior the PR description emphasizes.

mani-mathur-arch (commented):
## Summary

Solid, focused PR —  minor suggestions.

The experimental-option idiom (`KernelExperimental` off `Config`, `WithKernel*` prefix, fail-loud on Thrift) is the right design and consistent with the rest of the kernel backend work. `cBytes` / `applyKernelTLS` / `trySetKernelTLS` follow existing patterns well, and the reflective field-classification guard is a nice maintenance hook.

Main gap: no end-to-end test that `Connect()` actually rejects on the Thrift path (see inline on `kernel_experimental_test.go`). Everything else is polish / nice-to-have.

connector.go
mani-mathur-arch: Done in `2b7ba24` — added the `ErrRequiresKernelBackend` sentinel (mirror of `ErrNotSupportedByKernel`) in `errors/errors.go` and wrapped the Thrift-path rejection with `%w`, so kernel-first/Thrift-fallback callers can detect it via `errors.Is`. Documented the detection pattern in `doc.go`, and the new `TestWithKernelOptionsRejectedOnThriftPath` asserts on the sentinel rather than message text.

connector.go
mani-mathur-arch: Done in `2b7ba24` — `WithKernelTrustedCerts` now copies the PEM defensively (`append([]byte(nil), pem...)` for non-empty input), matching `DeepCopy`, so a caller mutating the slice between `NewConnector` and `Connect` can't change the trust store. Left the empty-`pem` → non-nil-block behavior as-is: that is the intended fail-loud signal on the Thrift path.

connector.go
mani-mathur-arch: Done in `2b7ba24` — added a `WARNING:` block to `WithKernelSkipHostnameVerify` explaining the machine-in-the-middle vector (a trusted-CA cert for a different host is accepted), consistent with `WithSkipTLSHostVerify`.

kernel_experimental_test.go
mani-mathur-arch: Done in `2b7ba24` — added `TestWithKernelOptionsRejectedOnThriftPath` (untagged, table-driven over both options): it calls `Connect()` with no `WithUseKernel` and asserts the error wraps `ErrRequiresKernelBackend`, modeled on `TestKernelBackendNotCompiledIn`. This exercises the connector reject branch end-to-end, so the file-header's "Thrift fail-loud signal" claim is now actually covered.

PR #401, opened 2026-07-14
feat(kernel): route kernel logging through the driver log level + correlate

mani-mathur-arch:
## Context for reviewers

Part of the **SEA-via-kernel** backend for the Go driver (an alternative to Thrift that talks to Databricks SQL through the Rust `databricks-sql-kernel` static library over a C ABI, behind the `databricks_kernel` build tag; the default pure-Go build is unaffected). This PR makes the kernel backend's logging behave like the rest of the driver's.

**Stack position:**
```
  └─ #400  richer TLS options
       └─ #401  kernel logging → driver log level    ← THIS PR
            └─ #402  nightly E2E workflow
```

**Where to focus review:** the hot-path allocation guard (see the review note below) — this is the load-bearing correctness point, since a debug log that allocates when disabled would regress the per-batch fetch path.

---

## What

Unify kernel-backend logging onto the driver's shared logger, so the one knob that controls the rest of the driver — `DATABRICKS_LOG_LEVEL` / `dbsql.SetLogLevel` — governs it, and each binding line carries the structured `connId`/`corrId`/`queryId` fields that let it be correlated in a multi-connection process.

Addresses @vikrantpuppala's review comment: `klog` and the kernel Rust subscriber emitted unstructured lines straight to `os.Stderr` — uncorrelatable, and gated on a separate `DBSQL_KERNEL_DEBUG` rather than the driver log level.

Closes **PECOBLR-3650**.

## Three parts (the ticket was scoped as one, the review surfaced two more)

The JIRA row was originally just "route kernel debug logging through `DATABRICKS_LOG_LEVEL` instead of `DBSQL_KERNEL_DEBUG`." The review raised a second, distinct problem (unstructured, uncorrelatable lines), so the complete fix is three parts:

- **Level** — `klog` now emits through `logger.Logger.Debug()` instead of raw `fmt.Fprintf(os.Stderr, ...)`. A cheap `kernelDebugOff()` front gate (`GetLevel() > Debug`) short-circuits before any formatting/allocation, so it stays a true no-op at the default Warn level, benchmarks included. Replaces the old `DBSQL_KERNEL_DEBUG` bool and the caller-less `KernelDebugEnabled`.
- **Correlation** — new `klogCtx(ctx, ...)` pulls `connId`/`corrId`/`queryId` off ctx via `logger.WithContext` (the exact idiom the Thrift/conn path uses at `connection.go:139`); the conn layer already stuffs those IDs into ctx before calling the backend. Every hot-path site (`execute`, `nextBatch`, `newKernelRows`, `Open`/`CloseSession`) uses it; ctx-less sites (`bindParams`, `kernelOp.close`, `lastError`) degrade to `klog`. The `WithContext` allocation is behind the same up-front level gate, so it never runs below Debug.
- **Kernel Rust logs** — `initKernelLogging` maps the driver's zerolog level → the `kernel_init_logging` level string (`kernelLogLevel`), so `DATABRICKS_LOG_LEVEL` drives the kernel's Rust logs too. The accepted level strings were verified against kernel source (`src/c_abi/session.rs`; a bad level safely falls back to `RUST_LOG`). `DBSQL_KERNEL_DEBUG` is kept as an advanced override (forces the subscriber on, defers to `RUST_LOG` for kernel-only verbosity).

### Why `DBSQL_KERNEL_DEBUG` is retained (not just `RUST_LOG`)

The kernel reads `RUST_LOG` **only when it's passed a NULL level** — the moment a concrete level is supplied it builds the filter from that and never consults `RUST_LOG` (`src/logging.rs`). Since the normal path now always passes the mapped driver level, `RUST_LOG` alone is inert; `DBSQL_KERNEL_DEBUG` is the one switch that makes the Go side pass NULL, unlocking `RUST_LOG`. That buys two things the driver level structurally can't: a kernel-Rust verbosity **decoupled** from the driver level, and **per-target** filtering — including the HTTP stack (`RUST_LOG=debug` → `hyper`/`reqwest`), which the driver-level mapping (scoped to `databricks::sql::kernel`) never surfaces.

## Known limitation (documented)

The kernel's **Rust** log lines are still plain text — they do not yet carry the structured fields, because they're emitted inside Rust. Fully structuring them needs the kernel log-callback ABI (**PECOBLR-3654 / K4**). Only the Go binding lines are structured today. `doc.go` states this.

## Testing

- Default `CGO_ENABLED=0` and tagged `CGO_ENABLED=1 -tags databricks_kernel`: build + vet + test — 24 pkg ok, 0 fail each.
- `TestKernelLogLevel` pins the zerolog→kernel level mapping (incl. fatal/panic→ERROR, default→WARN).
- `TestKernelLogNoAllocWhenOff` uses `testing.AllocsPerRun` to prove `klog`/`klogCtx` allocate **0 times** at the default Warn level — the hot-path zero-cost guarantee, kept as a regression guard.

### Review note (why the alloc guard exists)

**Isaac Review, 2 passes:** the first pass caught 2 legitimate MAJORs — `klogCtx` built `logger.WithContext` (which eagerly `make([]byte, 0, 500)`s inside zerolog's `With()`) *before* the `.Debug()` gate, so it allocated ~500 B per call even at Warn, on the per-batch `nextBatch` path. `logger.Logger.Debug().Msgf(...)` is **not** allocation-free when off if the arguments themselves allocate — the `.Debug()` no-op only discards the *event*, after the arg expressions have already run. Fixed with the up-front `kernelDebugOff()` gate + the `AllocsPerRun` regression test; **second pass clean** (0/0/0). Two info items filtered: a non-issue nil-ctx path, and a pre-existing global-logger data race that predates this change and affects the Thrift path equally.

## Stacking

Stacked on **#400** (`mani/sea-kernel-richer-tls`), which is stacked on #399. No `KERNEL_REV` change (inherits #399's pin).

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -186,16 +186,30 @@ the C header) under the cgo link path; `make build-kernel` does both steps:
 In a build without the tag, WithUseKernel(true) returns a clear error at connect
 time rather than silently using Thrift.
 
-To see the kernel's own logs interleaved with the driver's, set DBSQL_KERNEL_DEBUG
-to any non-empty value. That single flag turns on the driver's binding step tracer
-AND installs the kernel's Rust log subscriber, so both write to the same stderr in
-execution order. It is off by default (and must stay off during benchmarks — debug
-logging perturbs latency). The kernel's verbosity is then controlled by RUST_LOG,
-which the kernel honors directly; filter on the target databricks::sql::kernel
-(note the colons — it is the kernel's explicit log target, NOT the crate module
-path databricks_sql_kernel, which would match nothing):
+Kernel-backend logging follows the same knob as the rest of the driver. The
+binding-layer trace (session/statement/batch steps) is emitted through the shared
+logger at Debug level, so DATABRICKS_LOG_LEVEL=debug or dbsql.SetLogLevel("debug")
+turns it on — no separate switch — and each line carries the same structured
+connId/corrId/queryId fields as the driver's other logs, so it can be correlated in
+a multi-connection process. At the default Warn level the lines are suppressed with
+no cost (zerolog no-ops a below-level event), including during benchmarks.
 
-	# kernel logs only:
+	dbsql.SetLogLevel("debug") // or DATABRICKS_LOG_LEVEL=debug
+
+The driver's log level is also mapped into the kernel's own Rust log subscriber, so
+the same knob governs the kernel's internal (Rust) logs; they write to stderr and
+interleave with the driver's output. Note the kernel's Rust lines are currently
+plain text and do NOT yet carry the structured connId/corrId/queryId fields — that
+needs a kernel log-callback ABI (tracked separately); only the Go binding lines are
+structured today.
+
+For finer control of the kernel's Rust verbosity independent of the driver level,
+set DBSQL_KERNEL_DEBUG to any non-empty value: it forces the kernel subscriber on
+and defers to RUST_LOG for the kernel's threshold. Filter on the target
+databricks::sql::kernel (note the colons — the kernel's explicit log target, NOT
+the crate module path databricks_sql_kernel, which would match nothing):
+
+	# kernel logs only, at the kernel's own verbosity:
 	DBSQL_KERNEL_DEBUG=1 RUST_LOG=databricks::sql::kernel=debug ./your_app 2>&1
 	# kernel logs plus its HTTP stack (hyper/reqwest):
 	DBSQL_KERNEL_DEBUG=1 RUST_LOG=debug ./your_app 2>&1
--- a/doc.go
+++ b/doc.go
@@ -189,19 +189,35 @@ time rather than silently using Thrift.
 Kernel-backend logging follows the same knob as the rest of the driver. The
 binding-layer trace (session/statement/batch steps) is emitted through the shared
 logger at Debug level, so DATABRICKS_LOG_LEVEL=debug or dbsql.SetLogLevel("debug")
-turns it on — no separate switch — and each line carries the same structured
-connId/corrId/queryId fields as the driver's other logs, so it can be correlated in
-a multi-connection process. At the default Warn level the lines are suppressed with
-no cost (zerolog no-ops a below-level event), including during benchmarks.
+turns it on — no separate switch — and it lands in the same sink the rest of the
+driver uses (including a custom one set via logger.SetLogOutput). Request-scoped
+binding lines carry the same structured connId/corrId/queryId fields as the
+driver's other logs where a context is in scope (session open/close, execute,
+result-batch fetch), so they can be correlated in a multi-connection process; a few
+context-less lines (parameter binds, operation teardown) are emitted without those
+fields. At the default Warn level the lines are suppressed with no cost (zerolog
+no-ops a below-level event), including during benchmarks.
 
 	dbsql.SetLogLevel("debug") // or DATABRICKS_LOG_LEVEL=debug
 
 The driver's log level is also mapped into the kernel's own Rust log subscriber, so
-the same knob governs the kernel's internal (Rust) logs; they write to stderr and
-interleave with the driver's output. Note the kernel's Rust lines are currently
-plain text and do NOT yet carry the structured connId/corrId/queryId fields — that
-needs a kernel log-callback ABI (tracked separately); only the Go binding lines are
-structured today.
+the same knob governs the kernel's internal (Rust) verbosity. Two caveats apply to
+the Rust logs specifically, both consequences of the kernel's process-wide logging
+ABI rather than the Go layer:
+
+  - They are written to stderr directly and are NOT affected by
+    logger.SetLogOutput — an app that routes the driver's Go logs to a file or an
+    app logger will still find the Rust lines on stderr, not interleaved with that
+    sink.
+  - The Rust verbosity is fixed at the level in effect when the FIRST kernel
+    session in the process is opened (the kernel subscriber is process-wide and
+    installed once); a later dbsql.SetLogLevel changes the Go binding lines but not
+    the already-installed Rust subscriber. Set the level before opening the first
+    kernel connection to control the Rust logs.
+
+The kernel's Rust lines are also currently plain text and do NOT yet carry the
+structured connId/corrId/queryId fields — that needs a kernel log-callback ABI
+(tracked separately); only the Go binding lines are structured today.
 
 For finer control of the kernel's Rust verbosity independent of the driver level,
 set DBSQL_KERNEL_DEBUG to any non-empty value: it forces the kernel subscriber on
--- a/internal/backend/kernel/cgo.go
+++ b/internal/backend/kernel/cgo.go
@@ -36,7 +36,6 @@ import "C"
 import (
 	"context"
 	"fmt"
-	"os"
 	"runtime"
 	"sync"
 	"unsafe"
@@ -135,16 +134,14 @@ var initLoggingOnce sync.Once
 // level before opening the first kernel connection to govern the Rust logs.
 func initKernelLogging() {
 	initLoggingOnce.Do(func() {
-		// DBSQL_KERNEL_DEBUG override: force the subscriber on and let RUST_LOG tune
-		// it (level=NULL). Otherwise map the driver's current log level in, so the
-		// one DATABRICKS_LOG_LEVEL knob governs kernel verbosity too.
+		// resolveKernelLogArg decides the level (or NULL for the DBSQL_KERNEL_DEBUG
+		// override, which lets the kernel honor RUST_LOG). The pure decision lives in
+		// logging_level.go so it's unit-tested without cgo.
 		var level cStr
-		if os.Getenv("DBSQL_KERNEL_DEBUG") != "" {
-			level = cStr{c: nil} // NULL → kernel honors RUST_LOG
-		} else {
-			level = newCStr(kernelLogLevel(logger.Logger.GetLevel()))
+		if lvl, useNULL := resolveKernelLogArg(); !useNULL {
+			level = newCStr(lvl)
 			defer level.free()
-		}
+		} // else level stays {c: nil} → NULL → kernel honors RUST_LOG
 		if err := call(func() C.KernelStatusCode {
 			return C.kernel_init_logging(level.c, nil)
 		}); err != nil {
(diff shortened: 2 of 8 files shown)


mani-mathur-arch (commented):
The main theme is that the Go binding logs now follow the live driver logger, but the Rust subscriber remains process-wide/first-call-wins and stderr-only, so the public logging contract should be tightened or the implementation needs a stronger sink/level API.

internal/backend/kernel/cgo.go
```diff
@@ -34,82 +34,146 @@ package kernel
...
-// installs the subscriber (and its stderr output) for ALL subsequent kernel
-// sessions in that process, and there is no way to scope it to one connection or
-// turn it off afterward. The "off during benchmarks" guarantee therefore depends
-// on DBSQL_KERNEL_DEBUG being unset before package init. Making this a
-// per-process runtime knob (rather than an init-time env read) is tracked with
-// the logging-unification follow-up.
+// uninstalled — in a long-lived multi-tenant process the first kernel session's
+// level/destination applies to ALL subsequent kernel sessions, with no way to
+// re-scope or turn it off afterward. That is a kernel-ABI property, not a Go one.
 func initKernelLogging() {
```
mani-mathur-arch, line 135: This samples `logger.Logger.GetLevel()` only inside `sync.Once`, so Rust kernel logging does not actually follow runtime `dbsql.SetLogLevel` changes after the first kernel session. A process that opens one kernel connection at `warn` and later flips to `debug` will get Go binding debug logs but still no Rust debug logs; the reverse also leaves Rust debug logging enabled after the Go logger is lowered.

Suggested fix: either document this as first-kernel-session/process-start configuration (`SetLogLevel`/`DATABRICKS_LOG_LEVEL` must be set before the first kernel connection), or keep Rust logs behind an explicit `DBSQL_KERNEL_DEBUG` style opt-in until the kernel ABI supports reloading the Rust subscriber/filter.

doc.go
```diff
@@ -186,16 +186,30 @@ the C header) under the cgo link path; `make build-kernel` does both steps:
...
+logger at Debug level, so DATABRICKS_LOG_LEVEL=debug or dbsql.SetLogLevel("debug")
+turns it on — no separate switch — and each line carries the same structured
+connId/corrId/queryId fields as the driver's other logs, so it can be correlated in
+a multi-connection process. At the default Warn level the lines are suppressed with
+no cost (zerolog no-ops a below-level event), including during benchmarks.
+
+	dbsql.SetLogLevel("debug") // or DATABRICKS_LOG_LEVEL=debug
+
+The driver's log level is also mapped into the kernel's own Rust log subscriber, so
+the same knob governs the kernel's internal (Rust) logs; they write to stderr and
```
mani-mathur-arch: This makes the logging model sound more unified than the implementation is. The driver supports `logger.SetLogOutput`, but `kernel_init_logging(..., nil)` sends Rust kernel logs to stderr regardless of the driver's configured sink. In apps that route driver logs to stdout, a file, or an app logger, Rust logs will not interleave with the driver's output.

Suggested fix: document Rust kernel logs as stderr-only and not affected by `SetLogOutput`, or avoid presenting them as part of the shared logger contract until there is a kernel callback/file-sink ABI that can route them through the same output.

doc.go
```diff
@@ -186,16 +186,30 @@ the C header) under the cgo link path; `make build-kernel` does both steps:
...
-logging perturbs latency). The kernel's verbosity is then controlled by RUST_LOG,
-which the kernel honors directly; filter on the target databricks::sql::kernel
-(note the colons — it is the kernel's explicit log target, NOT the crate module
-path databricks_sql_kernel, which would match nothing):
-
-	# kernel logs only:
+Kernel-backend logging follows the same knob as the rest of the driver. The
+binding-layer trace (session/statement/batch steps) is emitted through the shared
+logger at Debug level, so DATABRICKS_LOG_LEVEL=debug or dbsql.SetLogLevel("debug")
+turns it on — no separate switch — and each line carries the same structured
```
mani-mathur-arch: The `each line carries ... connId/corrId/queryId` claim is stronger than the current code. Most changed call sites now use `klogCtx`, but `bindParams` still emits `klog("bound param ...")`, and `kernelOp.close` logs via `klog` while `Close(ctx)` ignores its context. Those are useful traces in multi-connection debugging and remain uncorrelated.

Suggested fix: thread `ctx` into `bindParams` and operation teardown logging, or soften the docs to say request-scoped binding lines carry correlation fields where context is available.

internal/backend/kernel/cgo.go
```diff
@@ -34,82 +34,146 @@ package kernel
...
+	switch l {
+	case zerolog.TraceLevel:
+		return "TRACE"
+	case zerolog.DebugLevel:
+		return "DEBUG"
+	case zerolog.InfoLevel:
+		return "INFO"
+	case zerolog.WarnLevel:
+		return "WARN"
+	case zerolog.ErrorLevel, zerolog.FatalLevel, zerolog.PanicLevel:
```
mani-mathur-arch: Mapping `fatal`/`panic` to Rust `ERROR` makes the Rust side noisier than the Go driver for users who explicitly choose those levels. With `DATABRICKS_LOG_LEVEL=fatal`, Go error logs are suppressed, but the Rust subscriber can still emit error lines to stderr.

Suggested fix: map `FatalLevel`/`PanicLevel` to `OFF` unless the kernel supports equivalent fatal/panic thresholds, or call out the mismatch explicitly in docs/tests.

internal/backend/kernel/kernel_test.go
```diff
@@ -70,6 +73,58 @@ func TestSetKernelTLS(t *testing.T) {
...
+	}
+}
+
+// klog / klogCtx must be allocation-free at the default (above-Debug) log level —
+// the "no hot-path cost, safe during benchmarks" guarantee. klogCtx is the one that
+// matters: logger.WithContext eagerly allocates (zerolog's With() does
+// make([]byte, 0, 500)) BEFORE the .Debug() gate, so without the up-front
+// kernelDebugOff() guard this would allocate ~500 B per call — per Arrow batch on the
+// nextBatch hot path. Guards against reintroducing that (an Isaac Review MAJOR).
+func TestKernelLogNoAllocWhenOff(t *testing.T) {
```
mani-mathur-arch, line 88: The allocation guard is valuable, but the tests still do not prove the positive logging behavior this PR is changing. This would still pass if `klogCtx` stopped emitting at debug level or stopped attaching `connId`/`corrId`/`queryId`.

Suggested test additions: add a capture test that sets `logger.SetLogOutput(&buf)` plus debug level, calls `klogCtx` with conn/corr/query IDs, parses the emitted log, and asserts the message and fields; then set warn and assert no output. Also consider a small seam around `kernel_init_logging` so tests can verify the mapped level is actually passed, `DBSQL_KERNEL_DEBUG=1` passes NULL for `RUST_LOG`, and the first-call-wins behavior is documented by test rather than only by comments.

internal/backend/kernel/cgo.go
mani-mathur-arch: Fixed in 5314b80. You're right — the Rust level is sampled once, inside `initLoggingOnce.Do`, so a later `SetLogLevel` re-levels the Go binding lines (`klog`/`klogCtx` re-read `GetLevel()` per call) but not the already-installed Rust subscriber. Documented this as first-kernel-session/process-start configuration rather than pretending it's live: the `initKernelLogging` comment now states the level is sampled here once at the first session, and `doc.go` tells users to set the level before opening the first kernel connection to govern the Rust logs. Left it as documentation (not an explicit opt-in) since the process-wide/first-call-wins constraint is a kernel-ABI property — a runtime-reloadable Rust filter needs the log-callback ABI (PECOBLR-3654 / K4).

doc.go
mani-mathur-arch: Fixed in 5314b80. Agreed — `kernel_init_logging(..., nil)` sends the Rust lines to stderr unconditionally, so `logger.SetLogOutput` (which only redirects the Go zerolog sink) does not affect them. `doc.go` now documents the Rust logs as stderr-only and explicitly NOT routed through `SetLogOutput`, called out as a consequence of the kernel's process-wide logging ABI rather than presenting them as part of the shared-sink contract. The `initKernelLogging` comment says the same.

doc.go
mani-mathur-arch: Fixed in 5314b80. Correct — `bindParams` and `kernelOp.close` (whose `Close(ctx)` discards its ctx) still emit via context-less `klog`, so those lines are uncorrelated. Rather than thread ctx through those two sites (bindParams runs pre-execute and teardown is ctx-less by the C-ABI ordering), I softened the docs: `doc.go` now says request-scoped binding lines carry the correlation fields where a context is in scope (session open/close, execute, result-batch fetch), and explicitly notes the context-less parameter-bind / teardown lines are emitted without them.

internal/backend/kernel/cgo.go
mani-mathur-arch: Fixed in 5314b80. Good catch on the asymmetry. Rather than a docs caveat I took the stronger fix: `kernelLogLevel` now maps `FatalLevel`/`PanicLevel` -> `OFF` (not `ERROR`). At `DATABRICKS_LOG_LEVEL=fatal` the Go driver suppresses even its own `Error()` lines, and the kernel has no fatal/panic threshold, so the Rust subscriber must stay at least as quiet as the driver the user configured — mapping to `OFF` guarantees it never emits stderr output the user explicitly asked to silence. `TestKernelLogLevel` is updated to pin `fatal`/`panic` -> `OFF`.

internal/backend/kernel/kernel_test.go
mani-mathur-arch: Fixed in 5314b80. Added `TestKernelLogCtxEmitsCorrelation`: it sets `logger.SetLogOutput(&buf)` + debug level, calls `klogCtx` with conn/corr/query IDs, parses the emitted JSON line, and asserts both the message (`[kernel] step execute`) and the `connId`/`corrId`/`queryId` fields; then flips to warn and asserts no output. That covers the positive half the alloc guard can't see — it would now fail if `klogCtx` stopped emitting at debug or stopped attaching the fields. I did not add the `kernel_init_logging` seam in this pass (it needs cgo indirection to intercept the C call); the level mapping itself is pinned by `TestKernelLogLevel`, and the first-call-wins behavior is documented — happy to add the seam as a follow-up if you'd like the passed-level / NULL-on-DBSQL_KERNEL_DEBUG behavior asserted by test too.

mani-mathur-arch: Follow-up in ef7e5ec — added the seam after all. Extracted the pure level-resolution decision into an untagged `logging_level.go` as `resolveKernelLogArg() (level string, useNULL bool)`, and added `TestResolveKernelLogArg` in an untagged `logging_level_test.go` (mirrors the existing `errors_classify.go` pattern, so both it and the moved `TestKernelLogLevel` now run under `CGO_ENABLED=0` — no kernel lib needed). It pins the two invariants: `DBSQL_KERNEL_DEBUG` (non-empty) → `useNULL=true` so the kernel honors `RUST_LOG`, and otherwise the driver level is mapped in (incl. fatal→OFF); empty value is treated as unset.

I stopped short of asserting the string is *actually* handed to `C.kernel_init_logging` and the `sync.Once` first-call-wins — that needs a function-pointer seam over the cgo call plus resetting a package global to test one assignment + stdlib `sync.Once`, which is where the cost/value flips. Left that documented rather than tested. Re-resolving.

PR #403, opened 2026-07-14
ci: dispatch the SEA-via-kernel integration leg behind kernel labels + kernel CLOSE_STATEMENT telemetry & auth-rejection sentinel

mani-mathur-arch:
## Context for reviewers

Two changes for the **SEA-via-kernel** backend (the alternative backend that talks to Databricks SQL through the Rust `databricks-sql-kernel` static lib over a C ABI, behind the `databricks_kernel` build tag):

1. **CI**: extend the `databricks-driver-test` dispatch so the Go integration suite can run against the SEA-via-kernel backend, selected by label.
2. **Kernel backend**: record `CLOSE_STATEMENT` telemetry from `kernelRows.Close()`, and make the unsupported-authenticator rejection programmatically detectable via the `ErrNotSupportedByKernel` sentinel.

The two are unrelated in subject but small and both review-driven; they share this PR because the telemetry/sentinel commit was folded onto this branch. Reviewable independently — the CI change touches only `.github/workflows/`, the kernel change only `internal/backend/kernel/` + `kernel_config*.go`.

**Stack position (top of the stack):**
```
  └─ #400  richer TLS options
            └─ #402  nightly E2E workflow
                 └─ #403  CI dispatch label + CLOSE_STATEMENT telemetry   ← THIS PR
```

---

## 1. CI: kernel integration-test dispatch

Extend the `databricks-driver-test` dispatch so the Go integration suite can run against the SEA-via-kernel backend, selected by label — the analogue of the existing Thrift labels, not a copy.

| Label | Backend | Proxy mode |
|---|---|---|
| `integration-test` | Thrift | replay |
| `integration-test-full` | Thrift | passthrough |
| `integration-test-kernel` | sea (kernel) | **passthrough** |

- The label resolves both `proxy_mode` and a new `go_mode` (thrift vs sea).
- `go_mode` rides in the `repository_dispatch` `client_payload`. The receiver in driver-test reads it to decide whether to build the kernel static lib and run the `databricks_kernel`-tagged leg. Thrift labels send `go_mode: thrift` and are unchanged.
- **Only one kernel label, and it's passthrough:** there are no committed Go-sea proxy recordings yet, so a sea+replay label would have nothing to replay and produce a red, misleading run. A sea replay label can be added later once recordings are captured.
- The new label is added to the on-new-commit label-drop list and the skip-stub guidance.
- The **required merge-queue gate stays Thrift-only** (`go_mode` pinned to `thrift`): the kernel leg is previewable on demand but not a required gate until the SEA backend ships in a released driver.

### Two-repo change

This is the driver-side half. The receiver-side half (read `go_mode`, sea-gated kernel build prerequisites, thread `DATABRICKS_GO_MODE`, guard sea+replay) is a separate PR in `databricks-driver-test`. The runner + harness already handle `DATABRICKS_GO_MODE=sea` end-to-end. On the receiver side the review gate caught (and fixed before merge) two real defects worth knowing about: a **backend-aware `RECORDING_NAME`** (`databricks-go-${go_mode}`) so a sea run can't clobber/serve the committed thrift recordings, and a **narrowed git-token `insteadOf` scope** (exact `databricks-sql-kernel` repo, not the whole `databricks/` org) so the kernel App-token isn't applied to unrelated org fetches.

---

## 2. Kernel backend: CLOSE_STATEMENT telemetry + auth-rejection sentinel

Two review-driven, Go-only fixes on the kernel backend.

### CLOSE_STATEMENT telemetry

`conn` records `CLOSE_STATEMENT` / close latency / statement success-or-failure **only when the rows' `OnClose` callback fires** (the Thrift path fires it). `kernelRows.Close()` never called it, so kernel queries emitted no close telemetry — including failures — a production observability blind spot.

- `Close()` now times the teardown and fires `OnClose(latencyMs, chunkCount, iterationErr, nil)`. `closeErr` is nil because the kernel teardown (C stream/statement close) has no fallible close RPC to report, unlike Thrift's `CloseOperation`.
- `Next()` is split into a thin wrapper over `next()` that records the first non-EOF error as `iterationErr` (`io.EOF` is normal drain, not a failure), which `Close()` passes to the callback — mirroring the Thrift path's `iterationErr`.
- The callback is armed **only after `newKernelRows` finishes constructing successfully**, so the cleanup `Close()` on a schema-fetch/import failure does not record a falsely-successful `CLOSE_STATEMENT` (matching the Thrift `NewRows`, which sets the close callback just before returning).
- Idempotent: the existing `closed` guard means a double-`Close` fires `OnClose` exactly once.

### Auth-rejection sentinel

`resolveKernelAuth`'s unsupported-authenticator branch (token-provider / external / federated) returned a bare `errors.New`, so fallback logic ("kernel can't honor this auth → use Thrift") could only fragile substring-match. It now wraps `ErrNotSupportedByKernel` via `%w` so callers can `errors.Is` it, matching the other kernel-config rejections.

Deliberately **not** wrapped: the "requires a personal access token" (empty-token) error — that is missing-required-config, not an unsupported feature; wrapping it would wrongly signal "fall back to Thrift" for a forgotten token.

---

## Testing

- Workflow YAML validated; label resolution dry-run: `integration-test`→thrift/replay, `integration-test-full`→thrift/passthrough, `integration-test-kernel`→sea/passthrough.
- `TestKernelRowsCloseFiresOnClose` (tagged): success path (nil errs), iterationErr propagation, idempotent double-close, nil-callbacks-safe, construction-failure-no-success-close.
- `TestValidateKernelConfig`: strengthened the non-PAT authenticator rejection to assert `errors.Is(err, ErrNotSupportedByKernel)`.
- Default `CGO_ENABLED=0` + tagged `databricks_kernel`: build + vet + test — 24 packages ok, 0 fail each.

## Stacking

Stacked on **#402** (`mani/sea-kernel-nightly-e2e`).

### DIFF
--- a/internal/backend/kernel/kernel_test.go
+++ b/internal/backend/kernel/kernel_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/databricks/databricks-sql-go/driverctx"
 	dbsqlerr "github.com/databricks/databricks-sql-go/errors"
 	"github.com/databricks/databricks-sql-go/internal/backend"
+	dbsqlrows "github.com/databricks/databricks-sql-go/internal/rows"
 	"github.com/databricks/databricks-sql-go/logger"
 )
 
@@ -345,3 +346,78 @@ func strPtr(s string) *string { return &s }
 // CGO_ENABLED=0 build; see arrowscan_test.go. The decimal formatter lives in
 // internal/decimalfmt. This file keeps the kernel-specific tests: error mapping,
 // bad-connection classification, and the bound-params rejection.
+
+// kernelRows.Close() must fire the OnClose telemetry callback so the kernel path
+// records CLOSE_STATEMENT / latency / statement success-or-failure — conn gates
+// that recording on OnClose being called, and the Thrift path fires it. Before this
+// wiring, kernel queries emitted no close telemetry (a production blind spot). A bare
+// kernelRows is safe: Close() nil-guards cur/stream/op.
+func TestKernelRowsCloseFiresOnClose(t *testing.T) {
+	t.Run("success path reports nil iterErr", func(t *testing.T) {
+		var got struct {
+			called            bool
+			chunkCount        int
+			iterErr, closeErr error
+		}
+		r := &kernelRows{
+			chunkCount: 3,
+			callbacks: &dbsqlrows.TelemetryCallbacks{
+				OnClose: func(latencyMs int64, chunkCount int, iterErr, closeErr error) {
+					got.called, got.chunkCount, got.iterErr, got.closeErr = true, chunkCount, iterErr, closeErr
+				},
+			},
+		}
+		if err := r.Close(); err != nil {
+			t.Fatalf("Close() = %v, want nil", err)
+		}
+		if !got.called {
+			t.Fatal("OnClose was not fired")
+		}
+		if got.chunkCount != 3 {
+			t.Errorf("OnClose chunkCount = %d, want 3", got.chunkCount)
+		}
+		if got.iterErr != nil || got.closeErr != nil {
+			t.Errorf("OnClose errs = (%v, %v), want (nil, nil)", got.iterErr, got.closeErr)
+		}
+	})
+
+	t.Run("iterationErr is reported", func(t *testing.T) {
+		sentinel := errors.New("boom")
+		var gotIter error
+		fired := 0
+		r := &kernelRows{
+			iterationErr: sentinel,
+			callbacks: &dbsqlrows.TelemetryCallbacks{
+				OnClose: func(_ int64, _ int, iterErr, _ error) { fired++; gotIter = iterErr },
+			},
+		}
+		_ = r.Close()
+		if !errors.Is(gotIter, sentinel) {
+			t.Errorf("OnClose iterErr = %v, want %v", gotIter, sentinel)
+		}
+		// Idempotent: a second Close must not re-fire (conn/database-sql may double-close).
+		_ = r.Close()
+		if fired != 1 {
+			t.Errorf("OnClose fired %d times across two Close() calls, want 1", fired)
+		}
+	})
+
+	t.Run("nil callbacks is safe", func(t *testing.T) {
+		r := &kernelRows{} // no callbacks
+		if err := r.Close(); err != nil {
+			t.Errorf("Close() with nil callbacks = %v, want nil", err)
+		}
+	})
+
+	t.Run("construction-failure Close must not fire a success OnClose", func(t *testing.T) {
+		// newKernelRows arms the callback only after a successful build, so its
+		// cleanup Close() on a schema-fetch/import failure has callbacks==nil and
+		// must NOT record a (falsely successful) CLOSE_STATEMENT. Model that state.
+		fired := false
+		r := &kernelRows{ /* callbacks intentionally nil, as during construction */ }
+		_ = r.Close()
+		if fired {
+			t.Error("OnClose fired for a rows object that never finished construction")
+		}
+	})
+}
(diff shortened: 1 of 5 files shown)


internal/backend/kernel/kernel_test.go
```diff
@@ -264,3 +265,78 @@ func strPtr(s string) *string { return &s }
...
+			t.Errorf("Close() with nil callbacks = %v, want nil", err)
+		}
+	})
+
+	t.Run("construction-failure Close must not fire a success OnClose", func(t *testing.T) {
+		// newKernelRows arms the callback only after a successful build, so its
+		// cleanup Close() on a schema-fetch/import failure has callbacks==nil and
+		// must NOT record a (falsely successful) CLOSE_STATEMENT. Model that state.
+		fired := false
+		r := &kernelRows{ /* callbacks intentionally nil, as during construction */ }
```
mani-mathur-arch: This subtest is currently a no-op: `fired` is initialized to `false` and nothing can ever set it to `true`, so the assertion always passes.

Suggested fix: either remove this subtest and let the existing `nil callbacks is safe` case cover `callbacks == nil`, or make it assert the actual invariant by constructing a rows value that *would* fire if callbacks were accidentally armed in this state. For example, the useful regression coverage is at the construction path boundary: inject/mock the schema failure in `newKernelRows` and assert the supplied `OnClose` callback is not called during cleanup.

internal/backend/kernel/rows.go
```diff
@@ -56,7 +61,14 @@ type kernelRows struct {
 // newKernelRows fetches the schema up front (for Columns()) and returns the row
 // iterator; batches are pulled lazily on Next.
 func newKernelRows(ctx context.Context, op *kernelOp, stream *C.kernel_result_stream_t, cb *dbsqlrows.TelemetryCallbacks) (driver.Rows, error) {
-	r := &kernelRows{ctx: ctx, op: op, stream: stream, callbacks: cb, keyCache: arrowscan.NewStructKeyCache()}
+	// The telemetry callback is deliberately NOT set on r yet: the two cleanup
+	// r.Close() calls below run when construction FAILS (schema fetch/import), and a
+	// Close() with the callback set would fire OnClose as a *successful* close for a
+	// statement that never produced rows — masking the failure in CLOSE_STATEMENT
+	// telemetry. Assign it only on the success path (matching the Thrift NewRows,
+	// which sets closeCallback just before returning); the construction error itself
```
mani-mathur-arch: The behavior here is right, but this parenthetical is a little inaccurate: Thrift's `NewRows` assigns `closeCallback` during construction rather than literally just before returning.

Suggested fix: reword this to avoid relying on the Thrift timing comparison, e.g. `Assign it only on the success path so cleanup Close() on schema/import failure does not record a falsely successful CLOSE_STATEMENT; the construction error itself is surfaced to and recorded by the conn execute path.`

internal/backend/kernel/kernel_test.go
mani-mathur-arch: Fixed in d49b653. You're right that the subtest was a no-op — `fired` could never be set, so it just duplicated the nil-callbacks case. Rather than delete it, I took the construction-path-boundary option you suggested: the test now drives the real `newKernelRows` failure path. A nil result stream makes `kernel_result_stream_get_schema` return a defined `InvalidArgument` (the kernel null-checks the handle at `result.rs:161` — `(stream as *mut _).as_mut()`, never UB), so `newKernelRows` takes its `r.Close()` cleanup branch and returns an error, and the supplied `OnClose` is asserted **not** to fire. No warehouse/creds needed. Mutation-verified: arming the callback before the schema import makes the test fail as intended.

---

internal/backend/kernel/rows.go
mani-mathur-arch: Fixed in d49b653. You're right — Thrift's `NewRows` assigns `closeCallback` during construction (rows.go:172), not just before returning, so the comparison was inaccurate. Reworded to drop the Thrift timing comparison and state the invariant directly, per your suggestion: "Assign it only on the success path so cleanup Close() on a schema/import failure does not record a falsely successful CLOSE_STATEMENT; the construction error itself is surfaced to and recorded by the conn execute path."

---

PR #409, opened 2026-07-16
feat(kernel): mTLS client-certificate option + ABI-version check

mani-mathur-arch:
## What

Two experimental, kernel-only capabilities on the SEA-via-kernel Go backend,
matching the slimmed kernel PR ([databricks-sql-kernel#169](https://github.com/databricks/databricks-sql-kernel/pull/169)):

**mTLS client certificate**
- `WithKernelClientCertificate(cert, key)` — configure a client certificate +
  private key for mutual TLS. Both PEM halves are required together (they map to
  the single paired `kernel_session_config_set_tls_client_certificate`, so
  cert-without-key is unrepresentable); the private key is never logged. An
  incomplete request (cert-without-key, key-without-cert, or the option called
  with empty bytes from a failed PEM load) is rejected loudly at connect rather
  than silently connecting with no client identity. EXPERIMENTAL, kernel-only:
  the default (Thrift) backend rejects it at connect with `ErrRequiresKernelBackend`.
- Python (pyo3 `tls_client_cert`), Node (napi `client_cert_pem`), and JDBC
  (`SSLKeyStore`) already expose the equivalent — this brings the Go/kernel path
  to parity.

**ABI-version handshake**
- `checkABIVersion()` runs once at `OpenSession`, before any other kernel call:
  it compares the linked kernel library's `kernel_abi_version()` against the
  `DATABRICKS_KERNEL_ABI_VERSION` the driver's header declares and refuses to
  open on a mismatch, guarding against a driver built against one header linked
  against a differently-built prebuilt `.a`.

`KERNEL_REV` is pinned to the kernel head carrying these two symbols.

## Scope note

This PR was scoped down during review (thanks @vikrantpuppala on the kernel
side). Earlier revisions also wired ctx-deadline cancellation, a CloudFetch
toggle, and reverse-call log forwarding. Those kernel C-ABI symbols were removed
and deferred (they were purely additive with no merged consumer), so the Go
wiring is removed here to match:

- **ctx-deadline cancellation** (connect + mid-fetch cancel token / `ctxWatcher`)
  — Gap 6/7 (PECOBLR-3645 / PECOBLR-3646). `OpenSession` / `nextBatch` revert to
  the plain (non-cancellable) kernel calls; query cancellation still works via
  the pre-existing statement canceller. Re-add when the deadline-honoring path is
  actually needed (CloudFetch downloads already carry their own ~60s read-timeout).
- **CloudFetch toggle** (`WithKernelCloudFetch`) — the server gates
  `can_cloud_download` (`CONFIG_NOT_AVAILABLE` / `42K0I`), so there's no working
  e2e path until SEA inline-Arrow ships server-side. (CloudFetch itself is
  unaffected — only the disable knob is dropped.)
- **log forwarding** (reverse-call `kernel_set_log_callback` bridge) —
  `initKernelLogging` reverts to `kernel_init_logging` (kernel Rust logs to
  stderr / a file), the pre-existing path; the unified-sink forwarding was an
  ergonomics upgrade, not required.

## How we know it works

- **Untagged** (`CGO_ENABLED=0`): build + vet + `go test` green — the config
  wiring, experimental-field classification guard, and DeepCopy.
- **Tagged** (`cgo && databricks_kernel`, against the pinned kernel `.a`): build +
  vet + kernel-backend unit tests green — `TestSetKernelTLS` (incl. the mTLS
  case), `TestABIVersionMatches`, execute/error contracts.
- **Live e2e** against a warehouse: Select1, DataTypes (×18 incl.
  decimal/variant/geometry), **CloudFetch**, **Cancellation** (query cancel still
  works), M2M, TLSSkipVerify, Interval, MetricViewMetadata — all PASS.

mTLS is implemented and wired end-to-end in the kernel (verified against an
`openssl s_server -Verify` client-cert-required server: with the identity the
handshake completes; without it the server rejects with `certificate required`).
No live mTLS e2e here — no Databricks endpoint demands a client cert — so the
`kernel_mtls_test.go` coverage is option-wiring + the kernel-side handshake proof.

Co-authored-by: Isaac


internal/backend/kernel/backend.go
```diff
@@ -219,9 +213,22 @@ func (k *KernelBackend) OpenSession(ctx context.Context) error {
...
+	defer watcher.stop()
+	err := call(func() C.KernelStatusCode {
+		return C.kernel_session_open_cancellable(cfg, &sess, watcher.tokenPtr())
+	})
 	consumed = true
 	if err != nil {
+		// Prefer the caller's ctx error when the connect was interrupted by the
+		// deadline (database/sql convention), keeping the kernel error as cause.
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return fmt.Errorf("kernel: session_open cancelled: %w", ctxErr)
```
mani-mathur-arch: This drops the kernel error on the cancelled-connect path. The comment says we keep the kernel error as the cause, and the execute/fetch paths preserve both errors with two `%w`s, but this only wraps `ctxErr`. If the kernel returns useful details while the context deadline/cancel also fires, callers lose the underlying `*KernelError` diagnostics. Consider mirroring the execute/next_batch shape, e.g. wrap both `ctxErr` and `toConnError(err)`.

kernel_e2e_test.go
```diff
@@ -406,6 +406,85 @@ func TestKernelE2ECancellation(t *testing.T) {
...
+	start := time.Now()
+	// A deliberately slow, HIGH-CARDINALITY query: an un-aggregated cross join
+	// streams billions of rows, so the result keeps flowing far past the deadline
+	// (unlike a COUNT(*), which collapses to one row that returns instantly and
+	// exercises no mid-fetch cancel). An honored cancel — on the execute path or
+	// mid-fetch on the read path — must return well before it could complete.
+	var err error
+	rows, qerr := db.QueryContext(ctx,
+		"SELECT a.id, b.id FROM range(1000000000) a CROSS JOIN range(1000000) b")
+	if qerr != nil {
```
mani-mathur-arch: This can pass via the `qerr != nil` branch before result rows are ever read, so it does not deterministically prove the new `kernel_result_stream_next_batch_cancellable` path. That may be fine as an end-to-end smoke test, but I’d add or adjust coverage so at least one test reliably reaches `rows.Next()` after `QueryContext` succeeds and then observes deadline cancellation during fetch.

internal/backend/kernel/backend.go
mani-mathur-arch: Fixed in 7c094f8. The cancelled-connect path now wraps both the ctx error and the kernel error (two %w), so errors.Is still matches while the *KernelError stays reachable via errors.As. Matches the execute/next_batch paths.

kernel_e2e_test.go
mani-mathur-arch: Fixed in 7c094f8. Added TestKernelE2ECancelDuringFetch, which reads a row after QueryContext succeeds, then cancels and keeps draining so a later nextBatch deterministically hits the mid-fetch cancel path.

mani-mathur-arch: Follow-up: the first pass still tripped the pre-fetch ctx.Err() guard, so it never reached kernel_result_stream_next_batch_cancellable. Reworked in 7becbdf — a watcher fires the cancel only once a rows.Next() has been blocked past a buffered-row return (i.e. inside the fetch), and the test asserts the "next_batch cancelled" wrapper the C-call path emits (vs the guard's bare context error) to prove the in-flight token abort.

kernel_config.go
```diff
@@ -117,13 +138,17 @@ func buildKernelConfig(cfg *config.Config, kauth kernel.Auth) kernel.Config {
...
+	// WithKernelSkipHostnameVerify / WithKernelClientCertificate /
+	// WithKernelCloudFetch), if any. These have no Thrift-path equivalent (the
+	// connector rejects them on that path) and are forwarded verbatim to the
 	// kernel C ABI in OpenSession.
 	if ke := cfg.KernelExperimental; ke != nil {
 		kc.TLSTrustedCertsPEM = ke.TLSTrustedCertsPEM
 		kc.TLSSkipHostnameVerify = ke.TLSSkipHostnameVerify
+		kc.TLSClientCertPEM = ke.TLSClientCertPEM
+		kc.TLSClientKeyPEM = ke.TLSClientKeyPEM
+		kc.CloudFetchEnabled = ke.CloudFetchEnabled
```
mani-mathur-arch: **High — `buildKernelConfig` field forwarding is untested; a dropped `CloudFetchEnabled` line would ship green**

`TestBuildKernelConfig` (kernel_config_test.go:411) asserts only that `TLSTrustedCertsPEM` and `TLSSkipHostnameVerify` forward. The three new forwarded lines here — `TLSClientCertPEM` / `TLSClientKeyPEM` / `CloudFetchEnabled` (kernel_config.go:149-151) — have no runtime-copy assertion. `TestKernelExperimentalFieldsClassified` only checks the disposition *map*, not the actual copy (its own docstring warns of exactly this).

Deleting `kc.CloudFetchEnabled = ke.CloudFetchEnabled` makes `WithKernelCloudFetch(false)` a silent no-op (CloudFetch stays on) with **zero test failures** — the precise silent-option-drop the classified/forwarded machinery exists to prevent. Same exposure for the two mTLS halves.

<details>

Extend the "experimental TLS fields forwarded" subtest (or add one) to set `TLSClientCertPEM`, `TLSClientKeyPEM`, and `CloudFetchEnabled` on `KernelExperimental` and assert all three land in the returned `kernel.Config`, plus a `CloudFetchEnabled == nil` (unset) case.

</details>
---

kernel_e2e_test.go
```diff
@@ -406,6 +407,210 @@ func TestKernelE2ECancellation(t *testing.T) {
 	t.Logf("cancelled after %v with err=%v", elapsed, err)
 }
 
+// TestKernelE2EConnectHonorsCancelledCtx proves the connect-side ctx bridge:
+// opening a session under an already-cancelled ctx must fail, not connect. This
+// covers the OpenSession → kernel_session_open_cancellable path (the execute-path
+// TestKernelE2ECancellation above does not reach connect, which happens before any
+// statement runs).
+func TestKernelE2EConnectHonorsCancelledCtx(t *testing.T) {
```
mani-mathur-arch: **High — connect-cancel e2e never reaches the `kernel_session_open_cancellable` path it claims to cover**

This test passes an *already-cancelled* ctx to the first query. `OpenSession` returns at its entry guard `if err := ctx.Err(); err != nil { return err }` (backend.go:203) — pre-existing behavior — long before the new `kernel_session_open_cancellable` + ctxWatcher and the dual-wrap "session_open cancelled" branch (backend.go:199-217). `database/sql` may even return `ctx.Err()` before calling the driver at all. The wrong-error-type check is only `t.Logf`, not a failure, so the docstring's claim ("covers the OpenSession → kernel_session_open_cancellable path") is false.

The headline connect-side mid-connect cancellation (ctxWatcher firing during a blocking connect, and the branch that preserves both the ctx error and `*KernelError`) is **unexercised** — it could be deleted or broken and every test still passes. Contrast the rigorous timing dance in the mid-fetch test.

<details>

Add a test that fires the deadline *during* a slow connect (cold-start warehouse / unreachable host with a short ctx timeout) and assert the terminal error contains "session_open cancelled" and matches `errors.Is(context.DeadlineExceeded)` — mirroring the fetch test's mid-call proof.

</details>
---

kernel_config.go
```diff
@@ -86,6 +86,27 @@ func validateKernelConfig(cfg *config.Config) (kernel.Auth, error) {
...
+	// halves required together / never silently dropped" contract.
+	//
+	// "Incomplete" covers three cases: cert-without-key, key-without-cert, and —
+	// critically — the option being invoked with both halves empty/nil (e.g. from
+	// a failed PEM load), which the marker distinguishes from the option never
+	// being called. We also treat a bare cert/key present as a request even
+	// without the marker, so a config assembled without the option can't fail open
+	// either.
+	if ke := cfg.KernelExperimental; ke != nil {
+		certLen, keyLen := len(ke.TLSClientCertPEM), len(ke.TLSClientKeyPEM)
```
mani-mathur-arch, line 116: **Medium — `WithCloudFetch(false)` is silently ignored on the kernel path, violating this backend's "nothing silently ignored" contract** *(flagged independently by 3 reviewers: architecture, agent-compat, devil's-advocate)*

`validateKernelConfig` rejects every other unsupported neutral option loudly (Port, Transport, Timeout, retries-disabling WithRetries — all wrap `ErrNotSupportedByKernel`), and the mTLS pair is validated right here. But the backend-neutral `WithCloudFetch(false)` (sets `cfg.UseCloudFetch`) is classified `inert` (kernel_config_test.go:359), never read by `buildKernelConfig`, and never rejected here — so it is **silently dropped and CloudFetch stays on** (the kernel default).

doc.go:230 claims "Nothing is silently ignored" and lists only `WithMaxRows`/`WithRetries` as inert; `WithCloudFetch`'s inertness is documented nowhere. An operator disabling CloudFetch via the standard documented option (e.g. to bypass a proxy that blocks the S3 pre-signed-URL path) silently still uses it; `WithCloudFetch(false)` + `WithKernelCloudFetch(true)` is contradictory config resolved silently in favor of the kernel knob.

<details>

`UseCloudFetch` defaults to `true`, so an explicit `false` is distinguishable from unset — reject `UseCloudFetch == false` on the kernel path (wrapping `ErrNotSupportedByKernel`, pointing at `WithKernelCloudFetch(false)`), OR forward it into `CloudFetchEnabled` when `WithKernelCloudFetch` is unset. At minimum, add `WithCloudFetch` to doc.go's inert list.

</details>
---

internal/backend/kernel/logforward.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	switch level {
+	case kernelLevelError:
+		s.log.Error().Str("target", target).Msg(message)
+	case kernelLevelWarn:
+		s.log.Warn().Str("target", target).Msg(message)
+	case kernelLevelInfo:
+		s.log.Info().Str("target", target).Msg(message)
+	case kernelLevelDebug:
+		s.log.Debug().Str("target", target).Msg(message)
```
mani-mathur-arch: **Medium — kernel WARN/ERROR logs now inflate the driver's shared Warn/Error stream, unthrottled and unclassified**

This PR changes the kernel's Rust log destination from stderr to the shared `logger.Logger` sink. At the default Warn level the kernel forwards WARN+ERROR, and `forward` re-emits them at driver Warn/Error with no rate-limiting, sampling, dedup, or user-fault/infra reclassification. Kernel-internal transient events (retry "retrying request" WARN lines during a 503 / S3 storm) now interleave into the driver's own Warn/Error stream.

This reintroduces exactly the operational noise the `lastError` `isUserFault` split (cgo.go:241 — "don't inflate the WARN rate on-call alerts key on") was built to suppress, through a different door. A transient outage across many concurrent connections can make log-based alerting keyed on the driver's WARN/ERROR rate fire on recoverable events.

<details>

Either forward kernel events one severity lower by default (kernel ERROR→Warn, WARN→Info) and require `DBSQL_KERNEL_DEBUG` to preserve native severity, or add coarse rate-limiting/dedup in `forward`. Document that enabling kernel forwarding raises the driver's Warn/Error volume.

</details>
---

internal/backend/kernel/rows.go
```diff
@@ -185,12 +186,31 @@ func (r *kernelRows) nextBatch() error {
 		r.cur.Release()
 		r.cur = nil
 	}
+	// Bridge ctx onto a cancel token for the duration of this fetch, so a deadline
+	// firing mid-fetch drops the in-flight download future. A non-cancellable ctx
+	// yields a nil watcher → NULL token → the plain fetch path (no overhead).
+	watcher := newCtxWatcher(r.ctx)
```
mani-mathur-arch: **Low — a fresh ctxWatcher (goroutine + channel + 2 cgo token calls) is allocated per Arrow batch on any cancellable context** *(devil's-advocate conf 86; performance conf 64)*

`nextBatch` calls `newCtxWatcher(r.ctx)` + `defer watcher.stop()` on **every** batch. For a cancellable ctx (the idiomatic request-scoped / `WithTimeout` case) each batch does `kernel_cancel_token_new`, spawns a goroutine, and `stop()` does `close(done)` + `wg.Wait()` + `kernel_cancel_token_free`. The "no watcher overhead on the common background-context path" comment is true only for `context.Background()`; the common service case pays per-batch churn. The execute path (operation.go), by contrast, creates exactly one canceller for the whole statement.

A multi-GB CloudFetch result (thousands of batches) under a deadline ctx incurs thousands of goroutine spawn/teardown cycles, chan+WaitGroup syncs, and cgo token alloc/free. Dominated by per-batch network latency in the normal case (likely <2%), but strictly wasted work on the driver's known-sensitive large-result path — and if the kernel returns buffered/prefetched batches sub-millisecond, the fixed watcher tax approaches the per-batch cost.

<details>

Create one `ctxWatcher` per `kernelRows` (tied to `r.ctx`) in `newKernelRows`, reuse its token across all `nextBatch` calls, and `stop()` it in `Close` — matching the execute path's once-per-statement amortization. A fired token correctly aborts the current and all subsequent fetches.

</details>
---

internal/backend/kernel/backend.go
```diff
@@ -178,9 +192,27 @@ func (k *KernelBackend) OpenSession(ctx context.Context) error {
...
+	// non-cancellable ctx yields a nil watcher → NULL token → the plain open path,
+	// so there is no watcher overhead on a background context.
+	watcher := newCtxWatcher(ctx)
+	defer watcher.stop()
+	err := call(func() C.KernelStatusCode {
+		return C.kernel_session_open_cancellable(cfg, &sess, watcher.tokenPtr())
+	})
 	consumed = true
 	if err != nil {
+		// Prefer the caller's ctx error when the connect was interrupted by the
```
mani-mathur-arch: **Low — the cancelled-path error-wrap rationale and dual-`%w` line are triplicated across backend.go / rows.go / operation.go**

The same ~6-line rationale comment ("Prefer the caller's ctx error… Wrap BOTH… so `errors.Is` … AND `errors.As` … Mirrors the execute path") plus the identical `fmt.Errorf("kernel: X cancelled: %w (kernel error: %w)", ctxErr, conv(err))` dual-`%w` line now appears in three places: here (backend.go:206-214), rows.go:205-213, and the pre-existing operation.go:176-186. The dual-`%w` trick (making both `errors.Is` and `errors.As` work) is the load-bearing, subtle part.

Changing the wrap contract (add queryId, adapt to a Go multi-`%w` semantics change) is a 3-site edit, and the three prose copies drift independently.

<details>

Extract `cancelledErr(op string, ctxErr, kernelErr error) error` holding the format + dual-`%w`, document the rationale once on it, and have the call sites use it.

</details>
---

mani-mathur-arch:

**Score**: 69/100 — HIGH RISK

The C-ABI/FFI mechanics are sound — security and language reviewers independently verified cert handling, memory ownership, cancel-token lifecycle, and the log trampoline against the kernel header at the pinned rev, both clean. Risk is concentrated in two test-coverage gaps (a dropped forwarding line or a broken connect-cancel path would ship green) and a UX/contract inconsistency where the standard WithCloudFetch(false) is silently ignored on the kernel path (flagged by 3 reviewers). No confirmed production correctness bug; the HIGH RISK band is driven by the test gaps, not by shipped defects.

---

## Medium Findings

**M2** — `doc.go` explicitly denies the headline mid-fetch-cancellation feature this PR ships (`doc.go:275-277`, unchanged by this diff)

This PR adds mid-fetch cancellation: `rows.go` `nextBatch` now calls `kernel_result_stream_next_batch_cancellable` with a ctx-driven cancel token (proven by the new `TestKernelE2EQueryCancelDuringExecution`). But the read-path cancellation paragraph in doc.go still asserts the opposite: *"On the read path, context cancellation is honored at result-batch boundaries, not mid-fetch: an in-flight CloudFetch batch runs to completion before the cancel takes effect."*

`go doc github.com/databricks/databricks-sql-go` on the merged commit would print a mid-fetch-cancel disclaimer directly contradicted by code in the same commit. Users read the stale doc, assume a hung CloudFetch GET is uninterruptible, and design around a limitation that no longer exists.

This finding has no in-diff line to anchor an inline comment (the paragraph sits just past the last doc.go hunk, which ends at line 266).

- **Suggested fix**: Replace the paragraph with the now-true behavior (mid-fetch cancel via the kernel cancel token; a NULL token on a non-cancellable ctx preserves the plain path).

internal/backend/kernel/backend.go
mani-mathur-arch: Fixed in `4137103`. Extracted `cancelledErr(op string, ctxErr, kernelErr error)` holding the format + dual-`%w` wrap and its rationale once; `OpenSession`, `execute` (operation.go), and `nextBatch` (rows.go) now all call it. It lives in the tagged `cancel.go` rather than the untagged `errors_classify.go` so the `unused` linter (which runs on the default CGO_ENABLED=0 build) can see its tagged callers.

internal/backend/kernel/logforward.go
mani-mathur-arch: Fixed in `4137103`. Forwarded kernel ERROR/WARN are now demoted one severity by default (kernel ERROR→driver Warn, WARN→Info), each tagged with a `kernelLevel` field so the true origin stays visible; `DBSQL_KERNEL_DEBUG` (the same knob that widens kernel verbosity) preserves the native severities for a debugging session. So a transient kernel-internal retry storm no longer inflates the driver's Warn/Error stream that log-based alerting keys on. INFO/DEBUG/TRACE are unchanged (already below the alert threshold).

internal/backend/kernel/rows.go
mani-mathur-arch: Fixed in `4137103`. Now one `ctxWatcher` per `kernelRows` — created once in `newKernelRows` (tied to `r.ctx`), its token reused across every `nextBatch`, and `stop()`ed in `Close`, matching the execute path's once-per-statement amortization. A fired token correctly aborts the current and all subsequent fetches. Verified against staging that the mid-fetch cancel path still works: `TestKernelE2ECancelDuringFetch` still reaches the `next_batch cancelled` cancellable path on the first attempt.

kernel_config.go
mani-mathur-arch: Fixed in `4137103`. Added a `TestBuildKernelConfig` subtest that sets `TLSClientCertPEM`, `TLSClientKeyPEM`, and `CloudFetchEnabled` on `KernelExperimental` and asserts all three land in the returned `kernel.Config`, plus a `CloudFetchEnabled == nil` (unset stays nil / kernel default) case and an extended nil-`KernelExperimental` zero case. Deleting any of the three forwarding lines now fails the test rather than shipping green.

kernel_config.go
mani-mathur-arch: Fixed in `4137103`. `validateKernelConfig` now rejects a bare `WithCloudFetch(false)` on the kernel path (wrapping `ErrNotSupportedByKernel`, steering to `WithKernelCloudFetch(false)`), and `UseCloudFetch` is reclassified inert→rejected in the disposition map. When `WithKernelCloudFetch` is set it stays authoritative, so it isn't rejected. I confirmed both connection entry points (`NewConnector` via `config.WithDefaults()` and `ParseDSN`) apply the `UseCloudFetch=true` default, so the rejection only fires on an explicit `false`, never on an unset config. doc.go's "nothing silently ignored" list now names `WithCloudFetch(false)`.

kernel_e2e_test.go
mani-mathur-arch: Fixed in `4137103`. Added a tagged, hermetic `TestOpenSessionHonorsCtxDeadlineMidConnect` (new `cancel_connect_test.go`): a local black-hole TLS listener accepts the connection but never completes the handshake, so `kernel_session_open_cancellable` blocks; a short ctx deadline must then fire the cancel token mid-connect and surface as `session_open cancelled` matching `context.DeadlineExceeded`, promptly. It drives `OpenSession` directly (not database/sql, which can short-circuit an already-cancelled ctx), runs in the `build-and-test-kernel` CI job with no warehouse, and is `-race` clean. I also corrected this e2e test's docstring to state it only covers the pre-connect guard.

mani-mathur-arch:
Re **M2** (from the review summary — `doc.go` read-path paragraph denied the mid-fetch cancellation this PR ships): fixed in `4137103`. The paragraph now states that context cancellation is honored both at result-batch boundaries and mid-fetch (a cancel/deadline firing while a CloudFetch batch is in flight aborts that fetch), with the non-cancellable-context zero-overhead note. The stale "an in-flight CloudFetch batch runs to completion before the cancel takes effect" sentence is gone. (No inline thread to resolve — the paragraph sat past the last doc.go hunk.)

All 6 inline findings (#1–#6) are fixed and their threads resolved; details in each thread. Head is now `3c0a7c2`, which also adds a hermetic mTLS/custom-CA/hostname-skip handshake test. CI is green.

PR #411, opened 2026-07-16
feat(kernel): forward kernel logs into the driver logger via callback

mani-mathur-arch:
## What

Wire the driver side of the kernel reverse-call log callback (behind the `databricks_kernel` build tag), completing the one-knob logging story:

- Register a cgo-exported trampoline (`kernelLogTrampoline`) as the kernel's log sink via `kernel_set_log_callback`, so kernel-internal `tracing` events flow into the driver's logger — the unified Go+kernel stream, honouring `logger.SetLogOutput` — instead of only reaching stderr.
- The reverse-call machinery: a `runtime/cgo.Handle` round-trips the `*logSink` through the C `void* ctx` (cgo pointer rules), and a `defer recover()` firewall converts any panic into a dropped line (a panic across the cgo boundary would abort the process). The handle is a deliberate process-lifetime pin (no detach in the v0 ABI).
- `initKernelLogging` now installs the callback **instead of** calling `kernel_init_logging`. It passes the driver's mapped log level (`resolveKernelLogArg`, unchanged) as the callback's `level`, so the kernel's Rust lines follow `DATABRICKS_LOG_LEVEL` — the same knob PECOBLR-3650 established, now against the callback sink rather than stderr. `DBSQL_KERNEL_DEBUG` still defers to `RUST_LOG` (NULL level).
- The two C-ABI logging routes share the one process-global subscriber and are mutually exclusive; a non-`Success` install (host already installed a subscriber) is logged, never fatal to connect.

Closes [PECOBLR-3654](https://databricks.atlassian.net/browse/PECOBLR-3654) (driver side).

> Stacked on the Tier 2 driver PR (`mani/sea-kernel-tier2-features`). Draft — CI build is red until `KERNEL_REV` is bumped to a merged kernel SHA carrying `kernel_set_log_callback`.

## Tests

- Untagged (`CGO_ENABLED=0`): `logSink.forward` level mapping + nil-sink no-op (`logforward_test.go`), pure Go so it runs without the kernel linked.
- Tagged: `TestLogCallbackRoundTrip` drives the C→Go trampoline via a C invoke seam and asserts delivery (handle unwrap + recover firewall + sink dispatch); `TestLogCallbackTrampolineNilCtxSafe` proves a bad handle can't crash the process. Live round-trip verified — a forwarded WARN line lands in the driver's zerolog sink.

Co-authored-by: Isaac

[PECOBLR-3654]: https://databricks.atlassian.net/browse/PECOBLR-3654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -620,6 +620,13 @@ func WithKernelClientCertificate(cert, key []byte) ConnOption {
 		// caller mutating the buffers between NewConnector and Connect can't change
 		// the identity out from under us.
 		k := kernelExperimental(c)
+		// Record that mTLS was explicitly requested, independent of whether the
+		// caller passed non-empty bytes. Without this marker an empty cert+key
+		// pair (e.g. from a failed PEM load) is indistinguishable from the option
+		// never being called, and validateKernelConfig would accept it while
+		// applyKernelTLS silently skipped the setter — connecting with no client
+		// identity. The marker lets validation reject the incomplete request loudly.
+		k.TLSClientCertConfigured = true
 		if len(cert) > 0 {
 			k.TLSClientCertPEM = append([]byte(nil), cert...)
 		} else {
--- a/doc.go
+++ b/doc.go
@@ -195,19 +195,31 @@ level the lines are suppressed with no cost.
 
 	dbsql.SetLogLevel("debug") // or DATABRICKS_LOG_LEVEL=debug
 
-The same level is mapped into the kernel's internal (Rust) log subscriber, with two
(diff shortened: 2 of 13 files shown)


mani-mathur-arch:
Consolidated into #409 — all commits from this PR were folded (DCO sign-offs added) into `mani/sea-kernel-ctx-cancel` at 7e11982, which is now the single PR merging into main. Closing as redundant.

PR #412, opened 2026-07-18
feat(kernel): SEA-via-kernel PuPr tail: advanced proxy, Geography, conn/error telemetry, token-caching docs, configurable backoff

mani-mathur-arch:
The genuinely-new SEA-via-kernel driver work from the new-items backlog: the short PuPr telemetry/config/Geography tail plus advanced proxy and configurable backoff, followed by a datatype-parity + decimal-perf + RSS pass. Each is re-implemented clean on `main` (post-#399) — the pre-#399 POC was the spec, not the source. Every commit was reviewed with Isaac Review before commit.

### PuPr new-items tail
- **`feat(kernel): advanced HTTP proxy via WithKernelProxy`** (PECOBLR-3607) — fill the username / password / bypass_hosts args into the kernel's `set_proxy` setter (the kernel path passed the URL only). New `WithKernelProxy(url, user, pass, bypassHosts)` connector option; explicit proxy takes precedence over the env-var path.
- **`test(kernel): cover Geography as the WKT sibling of Geometry`** (PECOBLR-3620) — verify-and-close: GEOGRAPHY maps to the same Arrow Utf8/WKT shape as GEOMETRY (no renderer needed). Parity + live-e2e (skip-on-reject) coverage.
- **`feat(kernel): emit connection-config telemetry + fix connection-scoped error drop`** (PECOBLR-3627 / 3628) — emit `DriverConnectionParameters` at connect (kernel-path only; `mode=SEA`, `auth_mech`/`auth_flow` closed enums, proxy/arrow/query-tags/metric-view). Additive on top of #399's per-statement telemetry (not duplicated). The aggregator's `"error"` metric branch is routed correctly (a connection-scoped, no-statement error flushes standalone rather than being dropped), but it is **defensive**: no exported API emits a standalone `"error"` metric today, so the branch guards routing for a future producer rather than changing observable behavior now (clarified in review — see below).
- **`docs(kernel): note token caching + client reuse are inherited from the kernel`** (PECOBLR-3606 / 3626) — verify-and-close, doc-only: both are handled inside the kernel below the C ABI.
- **`feat(kernel): configurable HTTP retry / backoff`** (PECOBLR-3598) — wire `WithRetries` (RetryWaitMin/Max/RetryMax, incl. the disable form) into the kernel's new `set_retry_config` C-ABI setter, plus a kernel-only `WithKernelRetryOverallTimeout`.

### Datatype parity + decimal perf + RSS
- **`feat(kernel): report column-type metadata on the kernel Rows path`** (PECOBLR-3692) — `kernelRows` implemented only `driver.Rows`, so `database/sql` fell back to `""` DatabaseTypeName / `interface{}` ScanType for every column vs Thrift's typed metadata (a silent regression for ORMs / typed-scan tooling that value-level tests can't see). New shared pure-Go `arrowscan.ColumnTypeInfoFor` maps Arrow → `{DatabaseTypeName, ScanType, Length}` matching Thrift for all 21 types; `kernelRows` now implements the 4 optional `RowsColumnType*` interfaces. Pure-Go guard + live `TestKernelThriftColumnTypeParity`.
- **`perf(decimal): alloc-free exact decimal renderer (shared Thrift + kernel)`** (PECOBLR-3691) — rewrite `decimalfmt.ExactString` (on the shared render path) to allocate only the returned string: uint64 fast path renders into a stack scratch, `big.Int` only for true 128-bit magnitudes. Byte-identical output, pinned by an oracle-parity test (old impl kept verbatim as reference); 38.9ns/18B/1alloc → ~12ns/0B/0alloc. Live 50k-row multi-batch parity vs Thrift.
- **`feat(kernel): tune CloudFetch in-memory chunks via WithKernelMaxChunksInMemory`** (PECOBLR-3691, RSS) — bound how many decompressed CloudFetch chunks the kernel holds in memory (default 16), the knob that trades throughput for peak RSS on large results (measured 265 MB @ chunks=4 vs 856 MB @ chunks=32 on a 3M-row query). Kernel-only option forwarded as the client-only `cloudfetch_max_chunks_in_memory` session conf, which the kernel folds into its result config and strips before the SEA wire — never reaching the server or the Thrift path. No new C-ABI symbol.

## Kernel dependency

Two commits need new kernel behavior on the same kernel branch (PR #178):
- `set_retry_config` C-ABI symbol (backoff commit), and
- the `apply_client_result_overrides` reader that applies `cloudfetch_max_chunks_in_memory` (chunk-knob commit).

`KERNEL_REV` is pinned to that branch's head (`ffa73c0`), so the tagged CI links both. **Re-pin to the squash-merge SHA once the kernel PR lands.** The other commits are driver-only and need no kernel change.

## Testing

Both build paths green (`CGO_ENABLED=0 go test ./...` + `CGO_ENABLED=1 go test -tags databricks_kernel ./...`), `go vet` clean, gofmt clean, `golangci-lint run ./...` 0 issues. Pure-Go tests (option wiring, field-classification guards, ColumnType mapper, decimal oracle-parity, retry/proxy/telemetry resolvers, telemetry full-loop) run under `CGO_ENABLED=0`; tagged tests drive the real cgo setters; live e2e tests (proxy-rejects-bad-URL, proxy-credential-forwarding, geography, retry config/disable/overall-timeout, ColumnType parity, 50k-row decimal parity, CloudFetch chunk-knob drain) run against the staging warehouse. Each commit was reviewed with Isaac Review and its findings addressed before commit.

## Review findings addressed

`fix(kernel): address code-review findings` (`29edbba`) resolves all 9 findings from the code review; a follow-up `isaac review` came back clean (0 posted comments). All threads are resolved.

- **[Medium] `WithRetries(N, 0, 0)` silently reverted to the kernel default** — `kernelRetryConfig` now forwards the caller's `RetryMax` (with placeholder waits) on a degenerate range instead of returning `nil`. Because `WithDefaults()` runs before options, `WithRetries(n, 0, 0)` zeroed the waits and dropped the attempt count to the kernel's 5-retry default — a silent cross-backend divergence from Thrift. Only a zero-value range with no attempt count still falls back to the kernel default. New regression subtest.
- **[Medium] Stale Thrift-path rejection message** — the message named only the two TLS options, but all five `WithKernel*` options trip the same gate; generalized to name the `WithKernel*` family, and the test now guards against a stale subset.
- **[Medium] Connection-scoped error-drop branch targeted a dead path** — reframed the aggregator `"error"` branch and its test as **defensive** (no current producer), rather than an active behavior change.
- **[Low] Uint types omitted from the coverage guard** — added `UINT8/16/32/64 → BIGINT/int64` to `ColumnTypeInfoFor` (matching the scanner's int64 widening) and to both guard tests.
- **[Low] Doc comment named a nonexistent test** — replaced with the real guard names.
- **[Low] `WithKernelMaxChunksInMemory` absent from the `doc.go` catalog** — added the bullet.
- **[Low] No pure-Go coltype↔Thrift drift guard** — new `TestColumnTypeInfoMatchesThriftMapping` cross-checks `ColumnTypeInfoFor` against the Thrift `getScanType` / `GetDBTypeName` / length logic, so drift fails in `CGO_ENABLED=0` PR CI (not just the warehouse-gated nightly).
- **[Low] Proxy credential/bypass forwarding untested** — new `TestKernelE2EProxyForwardsCredentials` captures the `Proxy-Authorization` a local CONNECT proxy sees and asserts it decodes to exactly `user:pass` (a slot swap would show `pass:user`).
- **[Low] History-reference comments** — dropped the "(rejected before…)" / "Previously rejected —" clauses.


kernel_config.go
```diff
@@ -124,10 +123,101 @@ func buildKernelConfig(cfg *config.Config, kauth kernel.Auth) kernel.Config {
...
+			MinWait:        kernelRetryDisableWaitMin,
+			MaxWait:        kernelRetryDisableWaitMax,
+			MaxRetries:     0,
+			OverallTimeout: overall,
+		}
+	}
+
+	// Non-disabling: a degenerate range means WithDefaults didn't run — leave the
+	// kernel's default policy in place rather than fail the connect on a bad range.
+	if cfg.RetryWaitMin <= 0 || cfg.RetryWaitMax < cfg.RetryWaitMin {
```
mani-mathur-arch: **[Medium] `WithRetries(N, 0, 0)` silently reverts to the kernel's default retry policy**

The non-disabling branch returns `nil` (→ the kernel keeps its built-in 5-retry default) whenever `RetryWaitMin <= 0 || RetryWaitMax < RetryWaitMin`, dropping the caller's `RetryMax`. Because `WithDefaults()` runs *before* options and `WithRetries(10, 0, 0)` overwrites the waits to 0, the idiomatic `WithRetries(10, 0, 0)` (per its own godoc, which promises sane wait defaults) leaves `RetryWaitMin == 0`, hits this guard, and runs the kernel's default 5 retries instead of the requested 10 — silently, with no log. The in-code comment ("unusual outside tests / a Config without WithDefaults") is inaccurate: the normal option path reaches this. This also contradicts `doc.go`'s claim that `WithRetries` is forwarded to the kernel's HTTP retry config.

Flagged independently by 3 reviewers (language, ops, devil's-advocate); the Thrift path (go-retryablehttp) honors the same input, so it's a silent cross-backend divergence.

<details>

For `RetryMax >= 0` with a degenerate range, substitute the placeholder wait bounds (as the disable branch already does) while keeping the caller's `MaxRetries`, rather than returning `nil`. Alternatively emit a `logger.Warn` naming the dropped values.

</details>
---

telemetry/aggregator.go
```diff
@@ -165,11 +165,17 @@ func (agg *metricsAggregator) recordMetric(ctx context.Context, metric *telemetr
...
 		} else {
-			// Buffer non-terminal errors with statement
-			if stmt, exists := agg.statements[metric.statementID]; exists {
-				stmt.errors = append(stmt.errors, metric.errorType)
-			}
+			// A non-terminal error with no statement to attach to — e.g. a
+			// connection-scoped error (empty statementID), or one whose statement
+			// was already flushed. Emit it standalone (flushed immediately) rather
+			// than dropping it silently, so a connection-level error still lands.
+			agg.batch = append(agg.batch, metric)
```
mani-mathur-arch, line 184: **[Medium] The "connection-scoped error-drop fix" targets a path no production code produces**

This new `else` branch flushes a non-terminal `"error"` metric standalone instead of dropping it — but no production code emits `metricType: "error"`. The only `recordMetric` callers are `AfterExecute` ("statement"), `RecordOperation` ("operation"), `RecordConnectionConfig` ("connection"), and `completeStatement` ("statement"); connection failures surface as *operation* metrics with `errorType` set, never this branch. `TestConnectionScopedError_NotDropped` reaches it only by calling `recordMetric` directly, a path unreachable from any exported API — so the advertised fix changes no observable behavior today.

Two concerns: (1) reviewers/operators may believe connection-level errors now land, when nothing routes here; (2) latent — if a producer is ever added, every non-terminal error triggers an immediate un-batched `flushUnlocked` (one HTTP request per error). Note this also modifies the *shared* telemetry path (both backends) inside a kernel-focused PR.

<details>

Either wire a real producer for the `"error"` metric type, or reframe the fix/test to describe the branch as defensive (no current producer) so it isn't read as an active behavior change.

</details>
---

internal/arrowscan/coltype_test.go
```diff
@@ -0,0 +1,98 @@
...
+		arrow.PrimitiveTypes.Float32, arrow.PrimitiveTypes.Float64,
+		arrow.BinaryTypes.String, arrow.BinaryTypes.Binary,
+		arrow.FixedWidthTypes.Date32, arrow.FixedWidthTypes.Timestamp_us,
+		&arrow.Decimal128Type{Precision: 10, Scale: 2},
+		arrow.ListOf(arrow.PrimitiveTypes.Int64),
+		arrow.MapOf(arrow.BinaryTypes.String, arrow.PrimitiveTypes.Int64),
+		arrow.StructOf(arrow.Field{Name: "x", Type: arrow.PrimitiveTypes.Int64}),
+		&arrow.DurationType{Unit: arrow.Microsecond},
+		arrow.FixedWidthTypes.MonthInterval,
+	}
```
mani-mathur-arch, line 98: **[Low] Coverage guard omits the unsigned Arrow types the scanner handles, so its completeness claim is false**

`TestColumnTypeInfoScanTypeCoversScanner` asserts "every Arrow type the value scanner (ScanCellCached) handles must also have a non-fallback entry in ColumnTypeInfoFor." But `ScanCellCached` handles `Uint8/16/32/64` (returning `int64`), while `ColumnTypeInfoFor` has no `arrow.UINT*` case — they fall to the default arm (`DatabaseTypeName ""`, `ScanType *interface{}`). This `scannerTypes` list silently omits every Uint type, so the stated invariant is already violated yet the test passes.

Practical impact is low (Databricks emits no unsigned types; the scanner arms are documented as defensive/unreachable), but the "guard" overclaims and would not catch a future scalar added to the scanner without a matching metadata entry.

<details>

Either add a `Uint*` case to `ColumnTypeInfoFor` (mapping to BIGINT/int64, matching the scanner's output) or drive this list from a shared enumeration; add the Uint rows and assert the guard actually fails on an uncovered type.

</details>
---

internal/arrowscan/coltype.go
```diff
@@ -0,0 +1,136 @@
...
+// sql.ColumnType — the DatabaseTypeName, ScanType, and Length the optional
+// driver.RowsColumnType* interfaces return. The kernel backend derives it from
+// the result's Arrow schema; ColumnTypeInfoFor is the single mapping both the
+// value scanner (ScanCellCached) and the type-metadata reporter agree on, so a
+// column's reported type can never drift from what a row actually scans into.
+//
+// The mapping mirrors the Thrift backend (internal/rows/rows.go getScanType /
+// ColumnTypeDatabaseTypeName / ColumnTypeLength) so a query reports byte-identical
+// column metadata on either backend — the same "identical results across backends"
+// contract the value renderers hold. TestColumnTypeInfoMatchesThrift and the live
```
mani-mathur-arch: **[Low] Doc comment names a guard test that does not exist**

This comment claims "TestColumnTypeInfoMatchesThrift and the live TestKernelThriftColumnTypeParity are the guards," but no `TestColumnTypeInfoMatchesThrift` exists anywhere in the repo (grep across the tree returns only this comment). The actual pure-Go guards are `TestColumnTypeInfoFor` and `TestColumnTypeInfoScanTypeCoversScanner`. A maintainer chasing the named guard to understand or extend the parity contract finds nothing.

<details>

Replace `TestColumnTypeInfoMatchesThrift` with the real names `TestColumnTypeInfoFor` / `TestColumnTypeInfoScanTypeCoversScanner`.

</details>
---

doc.go
```diff
@@ -233,30 +234,52 @@ backend exposes a U2M-scopes option, so this is a fixed difference between the t
...
+    otherwise mirrors. Use it for the advanced fields the env-var path can't express:
+    a structured bypass (no-proxy) list, or basic-auth credentials supplied out of
+    band rather than embedded in the URL. An explicit proxy takes precedence over the
+    environment; empty credentials / bypass are passed to the kernel as unset.
+  - WithKernelRetryOverallTimeout(d) sets the cumulative retry budget across all
+    attempts — the 4th retry knob, alongside the backoff bounds and max attempts
+    carried by the backend-neutral WithRetries (also honored on the kernel path). It
+    is kernel-only because the Thrift WithRetries surface has no overall-budget
+    equivalent; it mirrors the pyo3/napi retry_overall_timeout knob. Zero keeps the
+    kernel's default budget (900s).
```
mani-mathur-arch, line 257: **[Low] `WithKernelMaxChunksInMemory` is absent from the doc.go experimental-options catalog**

This "Experimental kernel-only options" list enumerates `WithKernelTrustedCerts`, `WithKernelSkipHostnameVerify`, `WithKernelProxy`, and `WithKernelRetryOverallTimeout`, but omits `WithKernelMaxChunksInMemory` — a public `ConnOption` added in this PR (connector.go:672) with complete godoc of its own. Nothing in doc.go mentions the CloudFetch in-memory-chunk / memory-tuning knob, so a consumer (or agent) using the package overview as the authoritative option catalog won't discover it.

<details>

Add a `WithKernelMaxChunksInMemory(n)` bullet to the list, mirroring the other four.

</details>
---

internal/arrowscan/coltype_test.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// TestColumnTypeInfoFor pins the Arrow → column-metadata mapping the kernel
+// backend reports through sql.ColumnType, so it stays byte-identical to the
+// Thrift path (internal/rows/rows.go). This is a pure-Go, no-warehouse guard: the
+// gap it regresses (PECOBLR-3692) was invisible to the value-parity suites
+// because they compare scanned VALUES, not Rows.ColumnType* metadata. The
+// expected DatabaseTypeName / ScanType / Length values are the ground truth
+// captured from the Thrift backend on a live warehouse.
+func TestColumnTypeInfoFor(t *testing.T) {
```
mani-mathur-arch, line 20: **[Low] Coltype↔Thrift parity has no pure-Go drift guard (nightly-only)**

`ColumnTypeInfoFor` (plus the duplicated `scanType*` vars) restates the Thrift path's `getScanType` / `GetDBTypeName` / `ColumnTypeLength` decisions from `internal/rows/rows.go`, but `TestColumnTypeInfoFor` pins its own hardcoded expectations rather than deriving them from the Thrift mapping. A change to the Thrift mapping (e.g. DECIMAL's scan type) leaves `coltype.go` silently divergent while CI stays green — the "byte-identical across backends" contract is then enforced only by the warehouse-gated `TestKernelThriftColumnTypeParity` (nightly, not PR CI), so parity regressions ship until a live run catches them. (The two switch on different type systems — Arrow IDs vs Thrift TTypeIds — so a shared map isn't trivial, but a cross-check test is.)

<details>

Have a pure-Go test assert `ColumnTypeInfoFor` against the Thrift mapping functions directly for the shared Databricks types, so drift fails in CGO_ENABLED=0 PR CI.

</details>
---

internal/backend/kernel/kernel_test.go
```diff
@@ -76,6 +77,63 @@ func TestSetKernelTLS(t *testing.T) {
...
+	}{
+		{"url only", Config{ProxyURL: "http://proxy:3128"}},
+		{"url + credentials", Config{ProxyURL: "http://proxy:3128", ProxyUsername: "u", ProxyPassword: "p"}},
+		{"url + bypass", Config{ProxyURL: "http://proxy:3128", ProxyBypassHosts: "localhost,*.internal"}}, //nolint:gosec // G101: test literals, not real credentials
+		{"all fields", Config{ProxyURL: "http://proxy:3128", ProxyUsername: "u", ProxyPassword: "p", ProxyBypassHosts: "localhost,*.internal"}}, //nolint:gosec // G101: test literals, not real credentials
+		{"none (no-op)", Config{}},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if err := trySetProxy(c.cfg); err != nil {
```
mani-mathur-arch, line 98: **[Low] Proxy credential/bypass forwarding to the kernel is never verified**

`TestSetProxy` asserts only `err == nil`. `kernel_session_config_set_proxy` stores the strings and returns OK regardless, so a swap or drop of the `ProxyUsername` / `ProxyPassword` / `ProxyBypassHosts` args is undetectable. The e2e `TestKernelE2EProxyRejectsBadURL` passes empty credentials and only proves the URL slot reaches the kernel; `TestResolveKernelProxy` / the round-trip test stop at config plumbing. A bug mapping `ProxyUsername` → the wrong slot (or dropping `ProxyBypassHosts`) would pass every test and silently break proxy basic-auth in production.

<details>

Add an e2e case routing through a proxy that requires basic-auth (assert success with correct creds, failure with wrong), or expose a kernel-config readback seam to assert the credential/bypass args.

</details>
---

kernel_config.go
```diff
@@ -78,14 +80,11 @@ func validateKernelConfig(cfg *config.Config) (kernel.Auth, error) {
...
-	// kernel provides retries (just not user-tunable), documented in doc.go.
-	if cfg.RetryMax < 0 {
-		return kernel.Auth{}, fmt.Errorf("databricks: disabling retries via WithRetries is %w "+
-			"(the kernel retries internally); omit it or use the default (Thrift) backend", dbsqlerr.ErrNotSupportedByKernel)
-	}
+	// WithRetries (RetryWaitMin / RetryWaitMax / RetryMax) is honored on the kernel
+	// path: newKernelBackend forwards it via kernelRetryConfig →
+	// kernel_session_config_set_retry_config, and a negative RetryMax (the disable
+	// form) maps to zero kernel retries. So it is neither rejected nor silently
+	// ignored here (it was rejected before the kernel exposed a retry-config setter).
```
mani-mathur-arch: **[Low] History-reference comment belongs in the commit message**

This comment ends "(it was rejected before the kernel exposed a retry-config setter)," and the companion test opens "Previously rejected —". Both describe prior behavior; the code should describe the present contract, and the historical rationale is already captured by the diff/commit. These lines rot as the "before" recedes.

<details>

Drop the parenthetical (and the test's "Previously rejected —" clause), keeping only the present-tense statement that `WithRetries` (including the disable form) is forwarded on the kernel path.

</details>
---

mani-mathur-arch:

**Score**: 85/100 — MODERATE RISK

Large additive PR closing out the SEA-via-kernel new-items backlog (~8 features). The high-risk internals are solid: the alloc-free decimal renderer is correct at the uint64↔128-bit boundary (benchmarked 12.4ns/0-alloc), carries no injection surface, and proxy credentials/tokens never reach logs or telemetry — no Critical and no security findings. Remaining findings cluster around one real correctness gap (WithRetries silently ignored on a user-reachable input) plus advertised-but-thin items (a stale rejection message, a telemetry fix targeting a currently-dead path) and documentation/test-coverage polish. A separately-tracked KERNEL_REV re-pin (acknowledged by the author, to be fixed before finalizing) was dismissed from this review. All findings verified against head SHA 1cff39c.

---

## Medium Findings

**[Medium] Thrift-path rejection message names a stale 2-option subset** — `connector.go:54-55`

(Posted here rather than inline: the cited lines are pre-existing code not present in this PR's diff, so GitHub can't anchor an inline comment there.)

The rejection error hardcodes `"(WithKernelTrustedCerts / WithKernelSkipHostnameVerify)"`, but the three options added in this PR — `WithKernelProxy`, `WithKernelRetryOverallTimeout`, `WithKernelMaxChunksInMemory` — all allocate `KernelExperimental` and hit the same `KernelExperimental != nil` gate. A caller who sets `WithKernelProxy` and forgets `WithUseKernel` is told about two TLS options they never touched, misdirecting a human or agent debugging the connect failure. `TestWithKernelOptionsRejectedOnThriftPath` only asserts `errors.Is(ErrRequiresKernelBackend)`, so the staleness is untested. Flagged by 3 reviewers (agent-compat, architecture, ops).

<details>

Generalize the message to "a WithKernel* option" (drop the specific list) or enumerate all five current options; update the guarding comment (connector.go:48-52) to drop "TLS."

</details>

---

kernel_config.go
mani-mathur-arch: Fixed in 29edbba. `kernelRetryConfig` now substitutes the placeholder wait bounds while **keeping the caller's `RetryMax`** whenever `RetryMax > 0` on a degenerate range, instead of returning `nil`. Only a fully zero-value range with no attempt count (a hand-built `Config` without `WithDefaults`) still falls back to the kernel default, since there's nothing to preserve there.

So `WithRetries(10, 0, 0)` — which `WithDefaults`-then-options zeroes the waits on — now forwards 10 retries to the kernel, matching the Thrift path; the silent cross-backend divergence is gone. New subtest `WithRetries(n,0,0) honors the attempt count despite zero waits` guards the regression, and the `kernelRetryDisableWait*` constants were renamed `kernelRetryPlaceholderWait*` since both paths now use them.

telemetry/aggregator.go
mani-mathur-arch: Fixed in 29edbba by reframing, per the second suggested option. No exported API emits a standalone `"error"` metric today (errors ride on `"operation"`/`"statement"`), so the branch is now documented as **defensive** — it guards the routing for a future producer rather than being read as an active behavior change. The comment also flags that a future producer would get one un-batched `flushUnlocked` per non-terminal error, so batching should happen at the producer. `TestConnectionScopedError_NotDropped`'s doc now states it drives `recordMetric` directly to pin the routing, not a currently-reachable path. No fake producer was wired.

internal/arrowscan/coltype_test.go
mani-mathur-arch: Fixed in 29edbba. Added a `UINT8/16/32/64 -> BIGINT/int64` case to `ColumnTypeInfoFor`, matching `ScanCellCached`'s int64 widening of unsigned integers, so the reported metadata stays in lockstep with the scanned value. Added the four Uint rows to `TestColumnTypeInfoFor` and to the `scannerTypes` list in `TestColumnTypeInfoScanTypeCoversScanner`, so the coverage invariant now actually holds.

internal/arrowscan/coltype.go
mani-mathur-arch: Fixed in 29edbba. Replaced the nonexistent `TestColumnTypeInfoMatchesThrift` in the doc comment with the real pure-Go guards — `TestColumnTypeInfoFor`, `TestColumnTypeInfoScanTypeCoversScanner`, and the new `TestColumnTypeInfoMatchesThriftMapping` — alongside the live `TestKernelThriftColumnTypeParity`.

doc.go
mani-mathur-arch: Fixed in 29edbba. Added a `WithKernelMaxChunksInMemory(n)` bullet to the experimental-options catalog in `doc.go`, mirroring the other four and covering the CloudFetch in-memory-chunk / memory-tuning knob and its client-only session-conf forwarding.

internal/arrowscan/coltype_test.go
mani-mathur-arch: Fixed in 29edbba. Added a pure-Go drift guard, `TestColumnTypeInfoMatchesThriftMapping` (in package `rows`, white-box so it can reach the unexported `getScanType`). It enumerates the shared Databricks types as `{Arrow type, equivalent Thrift TColumnDesc}` pairs and asserts `ColumnTypeInfoFor(arrow)` matches the Thrift `getScanType` / `rowscanner.GetDBTypeName` / length decisions directly — so a Thrift-mapping change (e.g. DECIMAL's scan type) now fails in `CGO_ENABLED=0` PR CI instead of waiting for the warehouse-gated nightly parity test.

internal/backend/kernel/kernel_test.go
mani-mathur-arch: Fixed in 29edbba. Added the tagged e2e `TestKernelE2EProxyForwardsCredentials`: it stands up a local CONNECT proxy, routes the kernel through it with `WithKernelProxy(url, "proxyuser", "proxypass", "")`, captures the `Proxy-Authorization` header the kernel's HTTP stack sends, and asserts it decodes to exactly `proxyuser:proxypass` — a username/password slot swap would decode to `proxypass:proxyuser` and fail. The capture is hermetic (needs no real proxy round-trip); the warehouse only gives the kernel a reason to issue CONNECT. This closes the gap `TestSetProxy` couldn't cover against the opaque kernel C config.

kernel_config.go
mani-mathur-arch: Fixed in 29edbba. Dropped the parenthetical "(it was rejected before the kernel exposed a retry-config setter)" in `kernel_config.go` and the test's "Previously rejected —" clause, leaving only the present-tense statement that `WithRetries` (including the disable form) is forwarded on the kernel path.

mani-mathur-arch:
Fixed in 29edbba (the last remaining Medium — the stale rejection message at `connector.go`).

All five `WithKernel*` options allocate `KernelExperimental` and trip the same `KernelExperimental != nil` gate, so the message now names the family generically instead of the two-TLS-option subset:

```
databricks: a WithKernel* option requires the kernel backend; add WithUseKernel(true) or remove it
```

The guarding comment was de-scoped from "TLS" to cover the trusted-CA bundle, hostname-verify skip, proxy, retry budget, and CloudFetch chunk cap. `TestWithKernelOptionsRejectedOnThriftPath` now asserts the message names the `WithKernel*` family and does **not** name a stale subset (`WithKernelTrustedCerts` / `WithKernelSkipHostnameVerify`), so the message can't silently drift as options are added.

With this, all 9 findings from the review are addressed; a follow-up `isaac review --uncommitted` came back clean (0 posted comments).

kernel_config.go
```diff
@@ -124,10 +123,120 @@ func buildKernelConfig(cfg *config.Config, kauth kernel.Auth) kernel.Config {
...
+
+	// Degenerate wait range (WithRetries(n, 0, 0), or a Config assembled without
+	// WithDefaults). If the caller asked for a positive attempt count, honor it with
+	// the placeholder waits — dropping it to the kernel default would silently ignore
+	// the caller's RetryMax. Only a zero RetryMax (no attempt count to preserve) falls
+	// back to the kernel's default policy.
+	minWait, maxWait := cfg.RetryWaitMin, cfg.RetryWaitMax
+	if minWait <= 0 || maxWait < minWait {
+		if cfg.RetryMax <= 0 {
+			return nil
```
msrathore-db, line 237: **[Medium] `WithKernelRetryOverallTimeout` is silently discarded when retries are zeroed** — *reproduced*

`overall` is computed up-front from `KernelExperimental.RetryOverallTimeout`, but this degenerate-range branch `return nil` whenever `RetryMax <= 0` and the waits are non-positive — throwing the caller's explicit overall budget away. The kernel then keeps its 900s default. This is the same "silently ignored option" failure the `WithKernel*` rejection gate was built to prevent, but here on an *accepted* option.

Reproduced under `CGO_ENABLED=0` against this PR head:
```
WithRetries(0,0,0) + WithKernelRetryOverallTimeout(5*time.Minute)
→ kernelRetryConfig returns nil → the 5-minute budget never reaches set_retry_config.
```
<details>
<summary><strong>Fix</strong></summary>

in the `RetryMax <= 0` degenerate case, when `overall > 0` still return a `&kernel.RetryConfig{MinWait: placeholder, MaxWait: placeholder, MaxRetries: 0, OverallTimeout: overall}` instead of `nil`, so the budget is forwarded even with zero retries.

</details>
---

internal/backend/kernel/backend.go
```diff
@@ -237,6 +250,56 @@ func (k *KernelBackend) applyKernelTLS(cfg *C.KernelSessionConfig) error {
...
+func (k *KernelBackend) applyRetry(cfg *C.KernelSessionConfig) error {
+	r := k.cfg.Retry
+	if r == nil {
+		return nil
+	}
+	if err := call(func() C.KernelStatusCode {
+		// Via the go_kernel_set_retry_config shim in cgo.go — cgo drops the direct
+		// declaration of the underlying symbol (see the shim's comment).
+		return C.go_kernel_set_retry_config(cfg,
+			C.uint64_t(r.MinWait.Milliseconds()), C.uint64_t(r.MaxWait.Milliseconds()),
```
msrathore-db, line 302: **[Medium] Sub-millisecond `RetryWaitMin` passes the guard, truncates to 0ms, and the kernel rejects the connect** — *reproduced*

`kernelRetryConfig` validates the wait range in `time.Duration` space (`minWait <= 0`), but `applyRetry` forwards `r.MinWait.Milliseconds()`. Any `0 < MinWait < 1ms` passes the guard yet becomes `min_wait_ms == 0`, which the kernel setter rejects as `InvalidArgument` (`src/c_abi/session.rs:372` on the pinned rev, "retry min_wait_ms must be > 0"). Same config connects fine on the Thrift path (go-retryablehttp accepts any Duration) — a cross-backend divergence contradicting the "WithRetries is honored on the kernel path" doc claim.

Reproduced: `WithRetries(5, 999*time.Microsecond, 30*time.Second)` → `min_wait_ms=0` → kernel Connect fails.
<details>
<summary><strong>Fix</strong></summary>

clamp `MinWait`/`MaxWait` up to at least 1ms before forwarding (or perform the guard in millisecond space), so a valid sub-ms Duration can't degrade to a rejected 0.

</details>
---

KERNEL_REV
```diff
@@ -1 +1 @@
-4d301455f7e70de9cb747e0f1143b8a3df8c5b48
+f29957733c8f6d79a7212bc8a50cf4e5ffa8dd69
```
msrathore-db: **[Medium] `KERNEL_REV` pins an unmerged PR-branch head → non-reproducible / breakable CI**

`f299577…` is the head of kernel PR #178 (OPEN). The required symbols (`set_retry_config`, `apply_client_result_overrides`) exist only on that branch, not kernel `main` (verified). A PR head is a mutable ref: a squash-merge / rebase / force-push moves it and GitHub GCs the orphaned commit, after which the kernel-lib fetch of `refs/pull/*/head` can no longer reach this SHA and every cold (cache-miss) CI build goes red with no source change here.

The PR body already commits to re-pinning to the squash-merge SHA once #178 lands — this should block merge, and ideally #178 merges first so the pin points at a `main` commit.

---

kernel_config.go
```diff
@@ -124,10 +123,120 @@ func buildKernelConfig(cfg *config.Config, kauth kernel.Auth) kernel.Config {
...
 		kc.TLSTrustedCertsPEM = ke.TLSTrustedCertsPEM
 		kc.TLSSkipHostnameVerify = ke.TLSSkipHostnameVerify
+		// Kernel-only CloudFetch in-memory-chunk knob (WithKernelMaxChunksInMemory).
+		// Injected into the kernel backend's own SessionConf ONLY (not via
+		// EffectiveSessionParams, which both backends share) as the client-only key
+		// the kernel reads and strips before the SEA wire — so it never leaks to the
+		// server or the Thrift path. Zero/negative keeps the kernel default (16).
+		if ke.MaxChunksInMemory > 0 {
+			// EffectiveSessionParams returned a fresh map, so mutating it is safe.
+			kc.SessionConf[config.KernelMaxChunksInMemoryConfKey] = strconv.Itoa(ke.MaxChunksInMemory)
```
msrathore-db, line 147: **[Medium] `MaxChunksInMemory` rides a stringly-typed session-conf key coupled to an unmerged ABI, with no build/unit guard**

The knob is injected as the magic string `cloudfetch_max_chunks_in_memory` into the kernel `SessionConf`; correctness depends entirely on the kernel's `apply_client_result_overrides` recognizing and stripping it before the SEA wire. I verified the kernel does this correctly on the pinned rev (strips before `filter_session_conf`, clamps to a ceiling of 256, ignores invalid/zero) — so it works today. But the sibling retry knob got a typed C setter with compile-time status assertions, whereas this one is guarded only by the nightly warehouse test. If the merged kernel renames the client key, the value silently no-ops or leaks to the server, and no CGO=0 / build-time check catches it.

<details>
<summary><strong>Alternative design</strong></summary>

give it a dedicated typed C setter (mirroring `set_retry_config`) so a key/signature drift is a link/compile error, or add a kernel-side ABI accessor the driver can assert the key against at build time.

</details>
---

kernel_telemetry_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		if p := kernelConnectionTelemetry(cfg); !p.UseProxy {
+			t.Error("UseProxy = false, want true when WithKernelProxy is set")
+		}
+	})
+}
+
+// kernelAuthMech maps each resolvable auth form to the closed AuthMech/AuthFlow
+// enums. M2M and U2M are both OAUTH but differ in flow; PAT is PAT with no flow.
+func TestKernelAuthMech(t *testing.T) {
```
msrathore-db, line 112: **[Medium] Telemetry auth-mech OAUTH/M2M/U2M arms untested; proxy slot-correctness proven only in the nightly warehouse test**

`kernelAuthMech` maps M2M→(OAUTH, CLIENT_CREDENTIALS) and U2M→(OAUTH, BROWSER_BASED_AUTHENTICATION), but `TestKernelAuthMech` exercises only the PAT/default arm. The OAUTH mech and both flow strings — the values that silently NULL server-side if wrong — are never asserted, and both arms are pure-Go testable via a config that resolves to M2M/U2M auth. Separately, proxy credential slot-ordering is proven only by the cgo+warehouse `TestKernelE2EProxyForwardsCredentials` (verified genuine, not hollow), so it is unguarded in `CGO_ENABLED=0` PR CI.

<details>

add `TestKernelAuthMech` subtests for M2M and U2M configs asserting the exact (mech, flow) pairs; if a cgo-without-warehouse CI lane exists, run the proxy-forwarding test there (the local CONNECT-proxy capture needs no real warehouse).

</details>
---

internal/rows/coltype_thrift_parity_test.go
```diff
@@ -0,0 +1,111 @@
...
+					{PrimitiveEntry: &cli_service.TPrimitiveTypeEntry{Type: id}},
+				},
+			},
+		}
+	}
+
+	// thriftLength mirrors rows.ColumnTypeLength's variable-length classification
+	// (the method needs a *rows with a live client, so replicate its switch here on
+	// the same TTypeId set — the exact list the method checks).
+	thriftLength := func(id cli_service.TTypeId) (int64, bool) {
```
msrathore-db: **[Low] The Length dimension of the coltype drift-guard re-copies the Thrift rule instead of calling it** *(flagged independently by test + maintainability reviewers)*

`TestColumnTypeInfoMatchesThriftMapping` genuinely cross-checks ScanType and DatabaseTypeName by calling the real Thrift `getScanType` / `rowscanner.GetDBTypeName`, but Length is compared against `thriftLength`, a hand-copied replica of `rows.ColumnTypeLength`'s type switch. So for the Length dimension both sides can drift together and stay green — the exact silent-divergence class PECOBLR-3692's guard was added to prevent. There are now three copies of the variable-length rule.

<details>

drive the length expectation from the real classifier (`rowscanner.GetDBTypeID` + the shared var-length set) rather than a copied switch, so a Thrift-side change fails this test. Interval types (DURATION / INTERVAL_MONTHS) are also absent from the cross-check cases.

</details>
---

connector.go
```diff
@@ -603,3 +614,65 @@ func WithKernelSkipHostnameVerify() ConnOption {
...
+// credentials supplied out of band rather than embedded in the URL userinfo.
+// username / password / bypassHosts may be empty (passed to the kernel as NULL,
+// i.e. unset). An empty url is a no-op — the environment-derived proxy, if any,
+// stays in effect.
+//
+// An explicit WithKernelProxy takes precedence over the environment: consulting
+// both would be ambiguous, and an explicit proxy is a deliberate override.
+//
+// EXPERIMENTAL, kernel-only: the default (Thrift) backend rejects this at connect.
+func WithKernelProxy(url, username, password, bypassHosts string) ConnOption {
```
msrathore-db: **[Low] `WithKernelProxy(url, username, password, bypassHosts)` is four adjacent same-typed string params — swap-prone**

Four consecutive `string` parameters give a caller (human or codegen/agent) nothing but argument order to get right; a username/password swap or a misplaced bypassHosts compiles cleanly and fails only at runtime with wrong proxy credentials. The authors' own `TestKernelE2EProxyForwardsCredentials` explicitly guards against a "slot swap," confirming the footgun is real.

<details>

accept an options struct (e.g. `WithKernelProxy(KernelProxy{URL, Username, Password, BypassHosts})`) or split into single-purpose options so each field is named at the call site; the C-ABI mapping stays internal.

</details>
---

internal/decimalfmt/decimalfmt_test.go
```diff
@@ -34,10 +36,150 @@ func TestExactString(t *testing.T) {
...
+		}
+	}
+}
+
+// BenchmarkExactString / BenchmarkAppendReused document the alloc profile the fix
+// targets: ExactString allocates only the returned string, and Append into a
+// reused buffer allocates nothing (the digit scratch stays on the stack). Run:
+//
+//	go test -run '^$' -bench 'ExactString|AppendReused' -benchmem ./internal/decimalfmt/
+func BenchmarkExactString(b *testing.B) {
```
msrathore-db, line 178: **[Low] The "0 alloc" `ExactString` claim is a benchmark artifact; production still allocates 1 string/cell**

`BenchmarkExactString` does `_ = ExactString(n, 2)`, so the compiler proves the returned string doesn't escape and elides the heap allocation (0 B/0 allocs). Every production caller keeps the result, so the string escapes — measured ~14ns/5B/1 alloc in the escaping case vs the benchmark's 8.1ns/0B/0 allocs. Still a genuine win over the old 38.9ns/18B/1 alloc, but not alloc-free on the render path. The truly zero-alloc `Append` variant has no production callers.

<details>

add a sink-assigned benchmark (`var sink string; sink = ExactString(...)`) to report the real 1-alloc production profile, and/or wire the arrowbased/kernel render loops to `Append` into a reused batch-scoped buffer to actually capture the 0-alloc path.

</details>
---

kernel_newitems_e2e_test.go
```diff
@@ -0,0 +1,195 @@
...
+	if want := wantUser + ":" + wantPass; string(decoded) != want {
+		t.Errorf("proxy credentials = %q, want %q (a username/password slot swap would show %q)",
+			decoded, want, wantPass+":"+wantUser)
+	}
+}
+
+// TestKernelE2ERetryConfig proves a tuned WithRetries policy (backoff bounds + max
+// attempts) reaches the kernel's HTTP retry config and the connection still works:
+// the setter accepts the range and a normal query succeeds.
+func TestKernelE2ERetryConfig(t *testing.T) {
```
msrathore-db, line 154: **[Low] Retry / overall-timeout e2e tests are smoke tests that pass whether or not the policy is applied**

`TestKernelE2ERetryConfig` / `TestKernelE2ERetryOverallTimeout` assert only that `SELECT 1` returns 1. A silently-dropped or ignored retry/overall-timeout config produces the identical result, so the docstring claim "reaches the kernel's HTTP retry config" is unproven — the test can't distinguish applied-vs-ignored (near-zero signal over connect-succeeds).

<details>

soften the docstrings to "connect succeeds with the option set," or add fault injection (unreachable-then-recover) to observe retry behavior. `TestKernelE2ERetryDisabled` is fine — it guards the `validateKernelConfig` acceptance change.

</details>
---

kernel_config.go
```diff
@@ -124,10 +123,120 @@ func buildKernelConfig(cfg *config.Config, kauth kernel.Auth) kernel.Config {
...
+// resolveKernelProxy fills the kernel Config's proxy fields. An explicit
+// WithKernelProxy (KernelExperimental.ProxyURL non-empty) wins verbatim,
+// including its out-of-band credentials and bypass list; otherwise the
+// endpoint's environment-derived proxy URL is used (with no credentials or
+// bypass list — the env path folds credentials into the URL userinfo and
+// consumes NO_PROXY during resolution, per Go's proxy-env convention). Kept
+// untagged here (like buildKernelConfig) so the explicit-over-env precedence is
+// asserted under CGO_ENABLED=0 (see TestResolveKernelProxy); newKernelBackend
+// calls it after buildKernelConfig so it stays a thin assembler.
+func resolveKernelProxy(cfg *config.Config, kc *kernel.Config) {
```
msrathore-db, line 267: **[Low] Config-validation failures surface as opaque wrapped kernel strings, not `errors.Is`-able sentinels**

`resolveKernelProxy` copies `ProxyURL` verbatim with no `url.Parse`; a malformed proxy URL or a degenerate retry range surfaces only as an opaque `fmt.Errorf("kernel: set_proxy: %w", …)` / `set_retry_config` wrap. Unlike the Thrift-path rejection (which wraps the `ErrRequiresKernelBackend` sentinel for `errors.Is`), there's no typed error distinguishing "your config was invalid" from a transient connect failure, so a caller can't branch on it.

<details>

validate the proxy URL in the Go layer (`url.Parse`) and wrap config-validation failures in a distinct sentinel analogous to `ErrRequiresKernelBackend`.

</details>
---

msrathore-db:

**Score**: 76/100 — MODERATE RISK

Clean, well-tested kernel-integration work — the author pre-ran Isaac Review and fixed 9 findings before submitting, and the security and language passes came back clean after deep verification. Two genuine logic bugs survive in the retry-config resolver (`kernelRetryConfig`), both reproduced with executable pure-Go tests against the PR head (F1, F2). The dominant structural risk is that `KERNEL_REV` pins an unmerged kernel PR branch head (#178), which the PR body already flags for re-pinning. No Critical/High findings.

---

Could not post inline comments for: F3, F9 — see body below.

## Medium Findings

**F3** — `connector.go` — **[Medium] `WithRetries(n,0,0)` retry pacing diverges: retry storm on Thrift vs paced backoff on kernel** — *verified end-to-end*

Options apply *after* `WithDefaults()` (`connector.go` runs defaults then the option loop), so `WithRetries(n,0,0)` lands with 0/0 waits. On the Thrift path both `internal/client/client.go` `backoff` and `internal/retry/retry.go` `Backoff` yield **0 sleep** with min=0 → n near-immediate retries (a retry storm). The kernel path substitutes 1s/60s placeholders. Same option, materially different server pressure per backend.

Arguably the kernel's paced behavior is the *safer* one — consider normalizing the waits once (clamp to the documented 1s/30s defaults when a positive `RetryMax` is set with zero waits) so both backends share one policy.

## Low Findings

<details>
<summary>Low findings (1) — click to expand</summary>

**F9** — `internal/backend/kernel/backend.go` — **[Low] Resolved retry policy and `MaxChunksInMemory` are neither logged nor emitted in telemetry**

The resolved retry policy (RetryMax / wait bounds / RetryOverallTimeout) and MaxChunksInMemory are forwarded to the kernel but never logged and are absent from the `DriverConnectionParameters` telemetry event (which does capture proxy/mode/arrow). If a customer reports ~15-min hung connects or a pod OOM on large results, on-call has no log line or telemetry field showing what the customer actually configured.

- **Suggested fix**: emit a Debug klog in `OpenSession` of the resolved retry policy + chunk cap, and add RetryMax / RetryOverallTimeout / MaxChunksInMemory to the `DriverConnectionParameters` payload alongside the existing proxy field.



</details>

kernel_config.go
mani-mathur-arch: Fixed in bd84ae8.

The problem: with `WithRetries(0,0,0)` the retry count is 0 and the waits are zeroed, so we returned `nil` and your `WithKernelRetryOverallTimeout` was thrown away the kernel just kept its 900s default.

Now, if you set an overall timeout, we still send it to the kernel (with placeholder waits and 0 retries) instead of returning `nil`. We only fall back to the kernel default when there's genuinely nothing to keep no retry count *and* no timeout. Added a regression test.

internal/backend/kernel/backend.go
mani-mathur-arch: Fixed in bd84ae8.

We checked the wait range in `time.Duration` terms but hand the kernel milliseconds, so a wait under 1ms (like `999µs`) became `0ms` and the kernel rejected the connect even though the Thrift path is happy with it.

Now we clamp `MinWait`/`MaxWait` up to a 1ms floor before forwarding, so a sub-ms wait can't degrade to a rejected `0`. Added a regression test.

kernel_config.go
mani-mathur-arch: Acknowledged, kept the string-key design (a typed C setter needs a new kernel ABI symbol, which is heavier than this knob warrants), but closed the "no cheap guard" gap in bd84ae8.

Added a pure-Go test that pins the exact key literal `cloudfetch_max_chunks_in_memory`. So an accidental change to the Go constant now fails in `CGO_ENABLED=0` PR CI (not just the nightly warehouse test), and the test is the greppable anchor a kernel-side rename has to update in lockstep.

kernel_telemetry_test.go
mani-mathur-arch: Fixed in bd84ae8.

Added M2M and U2M subtests to `TestKernelAuthMech`, asserting the exact pairs, M2M → `(OAUTH, CLIENT_CREDENTIALS)`, U2M → `(OAUTH, BROWSER_BASED_AUTHENTICATION)`. Both run pure-Go (they reuse the existing fake authenticators, since the real `u2m.NewAuthenticator` does live OIDC discovery at construction). So the enum values that NULL server-side if wrong are now guarded.

internal/rows/coltype_thrift_parity_test.go
mani-mathur-arch: Fixed in bd84ae8.

You're right, the Length side was comparing against a hand-copied `thriftLength` switch, so both copies could drift together and stay green (the third copy of that rule).

I pulled the rule out of `ColumnTypeLength` into one shared helper, `columnTypeLengthForID`, and now both the method and this test call it, no replica. A change to the Thrift-side length rule now fails this test. Also added the interval types (`DURATION` / `INTERVAL_MONTHS`) to the cross-check cases.

connector.go
mani-mathur-arch: Fixed in bd84ae8.

Agreed, four adjacent strings is a footgun. Changed the signature to a named struct: `WithKernelProxy(KernelProxy{URL, Username, Password, BypassHosts})`, so a username/password swap can't compile. Updated all call sites and the `doc.go` catalog. The C-ABI mapping stays internal.

internal/decimalfmt/decimalfmt_test.go
mani-mathur-arch: Fixed in bd84ae8.

Right, the benchmark's `_ = ExactString(...)` lets the compiler prove the string never escapes, so it reports 0 allocs; real callers keep the result and pay 1 alloc.

Added `BenchmarkExactStringSink` that assigns to a package-level var so the string escapes, reporting the true ~1-alloc/5B production profile alongside the escape-elided one, and clarified the comment to read the two together. (Still a solid win over the old 38.9ns/18B renderer; the truly 0-alloc `Append` form has no production caller yet, as noted.)

kernel_newitems_e2e_test.go
mani-mathur-arch: Fixed in bd84ae8, per your suggestion.

Softened both docstrings to "connect succeeds with the option set", a happy-path `SELECT 1` can't tell applied-from-ignored, so I stopped claiming otherwise. The pure-Go `TestKernelRetryConfig` is what actually pins the `WithRetries` → `kernel.RetryConfig` mapping. Left `TestKernelE2ERetryDisabled` as-is since it genuinely guards the `validateKernelConfig` acceptance change.

kernel_config.go
mani-mathur-arch: Fixed in bd84ae8.

Now `validateKernelConfig` runs `url.Parse` on the `WithKernelProxy` URL (and requires a scheme + host) in the Go layer, and wraps a malformed one in a new sentinel, `errors.ErrInvalidKernelConfig`. So a bad proxy URL is an `errors.Is`-able config error a caller can branch on — instead of surfacing as an opaque `kernel: set_proxy: …` wrap at connect. Added tests for the reject cases.

KERNEL_REV
mani-mathur-arch: Yes this is already called out in the pr description. This will be taken care of while merging.

mani-mathur-arch:
**F3 (retry pacing divergence): not fixing in this PR.**
 
Both backends diverge in the behaviour but the kernel's paced behavior is the safer of the two. The reason we're leaving it: the only way to make the two share one policy is to normalize the zero-wait case in the shared WithRetries option (or WithDefaults), which would change the Thrift path's long-standing retry pacing for every existing user of this driver, a behavior change well outside the scope of this kernel PR, and one that deserves its own change with its own testing and release note. 
  
**F9 (resolved retry policy + chunk cap not observable), fixed.** 
  
- Debug log in OpenSession: logs the resolved retry policy (maxRetries / wait bounds / overallTimeout) and the CloudFetch chunk cap right before connect. So if a customer reports a ~15-min hung connect or a large-result OOM, on-call can see from the debug log what they actually configured.
- Telemetry: added retry_max_attempts, retry_overall_timeout_ms, and max_chunks_in_memory to DriverConnectionParameters, populated on the kernel path alongside the existing proxy/mode/arrow fields. They're omitempty and the Thrift path leaves them zero, so nothing changes for that backend.
- Added test coverage for the telemetry fields (set-and-reflected, plus the zero-when-unset case).

msrathore-db (approved):
LGTM

PR #414, opened 2026-07-20
Add ErrorCategory type and constants for telemetry error classification

prathamesh-db:
## What
Adds an internal `ErrorCategory` string type in `internal/errors` with the
category constants used to classify driver errors for telemetry:

## Why
`classifyError` currently infers a category by substring-matching the error
*message* which is fragile and lossy. The goal is to declare the category at the error
source and have `classifyError` read it directly. This PR adds only the
category type and constants.

Jira: PECOBLR-3537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release History
 
+## Unreleased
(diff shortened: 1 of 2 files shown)


samikshya-db (approved):
Congrats on your first PR, @prathamesh-db 🥳  This is a good start for now, let's polish this as we wire this in the code base.

PR #416, opened 2026-07-20
fix(kernel): Thrift-parity fixes + GetArrowBatches + decimal-as-float opt-in + U2M scopes/telemetry + SPOG warehouse-id routing

mani-mathur-arch:
## What

SEA-via-kernel↔Thrift parity fixes surfaced by the Go comparator (two `database/sql` connections against the same warehouse, differing only in `useKernel`), plus the public Arrow batch API, U2M scope/browser fixes, an auth-logging cleanup, and SPOG routing. All kernel-path-only and entirely Go-side (no kernel/C-ABI change).

**Parity fixes**
- **TIMESTAMP out-of-nanosecond-range wrapping.** Arrow's `ToTime` forms an `int64`-ns intermediate that overflows for microsecond TIMESTAMPs outside ~1678–2262 (`'0001-01-01'`→`1754-08-30`). Now uses the unit-specific `time.Unix*` constructors, covering 0001–9999. Thrift unaffected (gets a preformatted string).
- **DDL affected-rows `-1` vs `0`.** The C ABI returns `-1` for unknown/NA counts (DDL/SELECT); Thrift reports `0`. Fold the `-1` sentinel to `0`; real DML counts pass through.
- **User-Agent not forwarded.** The kernel's built-in UA leaked into query history. Now forwards the driver UA via `set_custom_header`, matching Thrift; `WithUserAgentEntry` still customizes.
- **VOID/NULL columns typed `NULL` vs `STRING`.** Server stringifies VOID over the wire (even `SELECT NULL` reports `STRING_TYPE`), so `arrow.NULL` now maps to `STRING`, matching Thrift. Corrects two tests that pinned `NULL_TYPE` from the enum name, not live.

**Arrow batch API**
- **`GetArrowBatches` not exposed on kernel.** `kernelRows` now implements it over the zero-copy `next_batch` pull (prefetch for exact `HasNext`; transfer record ownership to the caller). `GetArrowIPCStreams` returns `ErrNotSupportedByKernel` (kernel exports via C Data Interface, not IPC). Compile-time `rows.Rows` assertion guards the gap.

**Decimal-as-float opt-in (`WithKernelDecimalAsFloat`)**
- New kernel-only connection option that scans **top-level DECIMAL** columns to a lossy `float64` (arrow-go `Num.ToFloat64`) instead of the exact fixed-point string. The kernel still receives native Arrow `Decimal128` over the wire — only the Go scanner changes, skipping the per-cell string materialization for a cheap scalar. **Off by default** (behavior byte-identical to today when unset); mirrors the Thrift driver's pre-`UseArrowNativeDecimal` default, and is rejected on the Thrift path (use `WithArrowNativeDecimal` there). Nested decimals (inside list/struct/map) still render exactly.
- **Why:** the kernel's fetch/decode transport is already the faster of the two backends; its only large-result deficit was the per-cell decimal string render. Skipping it surfaces the transport win — on decimal-heavy large results (`tpcds_sf100.catalog_sales`, 15×`DECIMAL(7,2)`, 1.3M–31M rows) the option is **~25% faster than Thrift** (0.75–0.76×) and **up to ~37% faster than the exact-string kernel path** (0.63×), consistently across sizes and reps. Lossy beyond ~15–17 digits, so it's an opt-in speed knob, not a default.

**U2M**
- **Scope parity.** Kernel U2M forwarded no scopes → kernel applied `all-apis`, rejected where the built-in client lacks it. `resolveKernelAuth` now populates `Auth.Scopes` from the same `oauth.GetScopes` Thrift uses.
- **Double-browser fix.** The telemetry/feature-flag client wrapped the interactive authenticator, launching a second browser at connect (blocking up to ~120s). Telemetry init is now skipped on the kernel+U2M path (detected via the `U2MClientID` structural interface); best-effort telemetry is dropped rather than made to prompt, matching Python's default. Unauthenticated telemetry is the follow-up — **PECOBLR-3839**.

**Auth logging**
- M2M/U2M/OAuth-config/token-provider paths logged through global `zerolog/log` (default Debug), ignoring `WithLogLevel` on both backends. Routed through the driver `logger` so they honor the configured level; auth logic untouched.

**SPOG routing (`WithWarehouseID` path)**
- On a unified/SPOG host the kernel needs `x-databricks-org-id`, which it injects **only** via `set_http_path` → `from_http_path` (parses the warehouse id + `?o=`). But Go calls `set_warehouse` (→ `ConnectionConfig::new`, no org id) whenever a warehouse id is set, and the kernel refuses a caller-supplied org-id header — so a `WithWarehouseID` SPOG session 303'd to `/login`. Now, when a warehouse id is set **and** the HTTP path is a canonical `warehouses`/`endpoints` path with a non-empty `?o=`, route via `set_http_path` (kernel parses the same id plus the org id). Guarded by `httpPathCarriesOrgRouting`, so non-SPOG connections keep the existing `set_warehouse` routing untouched. **Go-only** — Python/Node/ODBC always use the `http_path` route.

## Testing

- Unit: `TestScanCellTimestampOutOfNanoRange`, `TestNormalizeAffectedRows`, `TestColumnTypeInfoFor`/`…MatchesThriftMapping`/`…ScanTypeCoversScanner` (VOID→STRING), `TestScanCellDecimalAsFloat` (float64 arm vs unchanged exact-string default), U2M scope assertions in `TestValidateKernelConfig`/`TestResolveKernelAuthRealAuthenticators`, `TestInteractiveU2MAuthenticatorDetection`, `TestHTTPPathCarriesOrgRouting`. `./auth/...` green after the logger swap (no import cycle).
- Live (staging kernel backend): `GetArrowBatches` — 100,000 rows / 32 zero-copy batches, schema + count pass; `GetArrowIPCStreams` returns the sentinel. U2M double-browser prompt confirmed gone. **SPOG live on `peco.azuredatabricks.net`** — the `WithWarehouseID` routing that previously 303'd now returns `SELECT 1 == 1` / `current_user()` with the org-id header injected; non-SPOG queries unchanged. `WithKernelDecimalAsFloat` verified live: default returns `DECIMAL`/`string`, option-on returns `STRING`/`float64` for the same value, both `99.98`.
- Both build paths green: pure-Go (`CGO_ENABLED=0`) and tagged cgo (`-tags databricks_kernel`).
- **Comparator** (Thrift vs SEA): **9 diffs** (from 13 pre-fix, vs `main`'s 41). The 5 `dt_interval_column` metadata diffs are fixed here (INTERVAL_DAY_TIME→STRING). Of the 9 remaining, 8 are not driver bugs (NaN==NaN artifact ×3, non-deterministic EXPLAIN node ids, by-design volume-staging rejection, raw-`fetch_arrow` shape ×2, a permission-gated CREATE PROCEDURE `sqlState` detail); 1 is a deterministic `DESCRIBE TABLE` extra-trailing-rows diff that's server output (likely a comparator-side `skip_rows` filter, not this repo). Native decimal/interval on the raw-`fetch_arrow` path is a tradeoff, tracked as **PECOBLR-3838** (needs a kernel `KERNEL_REV` bump).

Stacked on `mani/sea-kernel-new-items` (#412).

Co-authored-by: Isaac


internal/backend/kernel/affectedrows.go
```diff
@@ -0,0 +1,21 @@
...
+
+// normalizeAffectedRows maps the kernel C ABI's modified-row count onto the value
+// database/sql's Result.RowsAffected() reports, matching the Thrift path.
+//
+// The C ABI returns -1 for "not applicable / unknown": DDL, SELECT, or a warehouse
+// that doesn't surface the counter (the kernel's num_modified_rows Option<i64> is
+// None). The Thrift path reports 0 in exactly those cases (TGetOperationStatusResp
+// defaults NumModifiedRows to 0), so the -1 sentinel is folded to 0 for parity.
+// A real DML count (>= 0) passes through unchanged.
+func normalizeAffectedRows(n int64) int64 {
```
msrathore-db, line 16: make this change in kernel not here

internal/backend/kernel/backend.go
```diff
@@ -132,6 +132,20 @@ func (k *KernelBackend) OpenSession(ctx context.Context) error {
 		return err
 	}
 
+	// User-Agent so query history attributes the kernel path to this driver.
+	if k.cfg.UserAgent != "" {
```
msrathore-db, line 178: please verify from query history

kernel_parity_test.go
```diff
@@ -116,7 +116,9 @@ func TestKernelThriftColumnTypeParity(t *testing.T) {
 		"CAST('2020-01-01 00:00:00' AS TIMESTAMP) a_ts, CAST('abc' AS BINARY) a_binary, " +
 		"array(1,2,3) a_array, map('k',1) a_map, named_struct('x',1) a_struct, " +
 		`parse_json('{"a":1}') a_variant, INTERVAL '1' DAY a_iv_dt, ` +
-		"INTERVAL '1' MONTH a_iv_ym, st_point(1,2) a_geom"
+		// VOID: the server declares it STRING_TYPE over Thrift; asserts the kernel
+		// arrow.NULL→STRING mapping matches live, not just in the pure-Go tests.
+		"INTERVAL '1' MONTH a_iv_ym, st_point(1,2) a_geom, CAST(NULL AS VOID) a_void"
```
msrathore-db, line 121: can you please confirm this NULL to string in other drivers too?

msrathore-db (approved):
Please confirm the similar issues in other drivers over kernel like node and python so we can decide the location of these fixes

msrathore-db (commented):
Verify other drivers

internal/backend/kernel/backend.go
mani-mathur-arch: verified, the behavior is sound and matches how the python driver passes its user agent and what the query history shows

internal/backend/kernel/affectedrows.go
mani-mathur-arch: changing this in the kernel will break the other drivers. other drivers have different specs for each "not applicable" row counts (python has -1, node has none). Go thrift has 0 for the database/sql spec.

kernel_parity_test.go
mani-mathur-arch: The other drivers (node and python) also have this gap,  the kernel returns nulls as void but thrift returns as string. i think this is an intentional design decision.  verified by the python comparator, node one is currently broken due to dependency differences. I can remove it if that should be taken up as a separate fix for all the drivers or at the kernel level

internal/backend/kernel/spog_routing_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+package kernel
+
+import "testing"
+
+// httpPathCarriesOrgRouting decides whether OpenSession reroutes a warehouse-id
+// connection through set_http_path so the kernel receives the SPOG ?o= org id.
+// It must fire ONLY for a canonical warehouses/endpoints path with a non-empty
+// o= value, so a reroute can never hand set_http_path a path from_http_path would
+// reject. Runs under CGO_ENABLED=0 (pure Go, no kernel tag).
```
msrathore-db: **(P2, carried over from previous push) Inaccurate build-tag comment.** This says "Runs under CGO_ENABLED=0 (pure Go, no kernel tag)", but the file carries `//go:build cgo && databricks_kernel` — I confirmed `CGO_ENABLED=0 go test -run TestHTTPPathCarriesOrgRouting` reports "no tests to run", so this predicate only runs under the kernel tag (nightly), not in default PR CI.

Since `httpPathCarriesOrgRouting` is pure string logic, consider moving it + this test out from behind the build tag — the way `affectedrows.go` is deliberately kept tag-free for pure-Go unit testing — so the SPOG-routing predicate runs in default PR CI. Otherwise, please just fix the comment.

internal/arrowscan/coltype.go
```diff
@@ -100,46 +77,38 @@ func ColumnTypeInfoFor(dt arrow.DataType) ColumnTypeInfo {
 	case arrow.TIMESTAMP:
 		return ColumnTypeInfo{DatabaseTypeName: "TIMESTAMP", ScanType: scanTypeDateTime}
 	case arrow.DECIMAL128:
-		// Thrift scans DECIMAL into sql.RawBytes; match it even though the kernel
-		// renders the value as an exact fixed-point string (both convert cleanly to
-		// a caller's *string/*[]byte, and reporting the same scan type keeps the
-		// backends indistinguishable).
+		// Match Thrift's sql.RawBytes scan type even though the kernel renders the
+		// value as an exact string — both convert cleanly to a caller's *string/*[]byte.
 		return ColumnTypeInfo{DatabaseTypeName: "DECIMAL", ScanType: scanTypeRawBytes}
```
msrathore-db, line 85: **(P3, doc) DECIMAL scan-type vs value in the new decimal-as-float mode.** With `WithKernelDecimalAsFloat(true)`, `ScanCellCached` now yields `float64` for a top-level DECIMAL, while this arm still reports `ScanType: sql.RawBytes`. That mirrors the Thrift path (RawBytes scan type + float value under `UseArrowNativeDecimal`), so it's not a bug — but this package leans hard on the "reported scan type and scanned value stay in lockstep" contract. Worth a one-line note here that the opt-in decimal-as-float mode is a deliberate, Thrift-matching exception, so a future reader doesn't read it as drift.

connector.go
```diff
@@ -580,6 +602,21 @@ func kernelExperimental(c *config.Config) *config.KernelExperimentalConfig {
...
+// WithKernelDecimalAsFloat makes the kernel path scan top-level DECIMAL columns to
+// a lossy float64 instead of the exact fixed-point string. The kernel still
+// receives native Arrow Decimal128; this only changes how the Go scanner
+// materializes each cell, skipping per-cell string formatting for a cheap scalar.
+// Precision beyond ~15-17 digits is lost, so it is opt-in and off by default;
+// it mirrors the Thrift driver's pre-UseArrowNativeDecimal behavior.
+//
+// EXPERIMENTAL, kernel-only: the default (Thrift) backend rejects this at connect
+// (use WithArrowNativeDecimal there instead).
+func WithKernelDecimalAsFloat(asFloat bool) ConnOption {
```
msrathore-db, line 614: **(P2, stale invariant) This option contradicts a documented parity invariant.** `internal/rows/arrowbased/arrowscan_parity_test.go` (`TestTopLevelDecimalRendering`) documents: *"across every real driver configuration a top-level DECIMAL comes back as the exact string on both backends (verified live)."* `WithKernelDecimalAsFloat` introduces exactly such a config where it does **not** — a top-level kernel DECIMAL comes back as `float64` (I confirmed this live). Please update that comment to carve out this opt-in mode so the invariant no longer reads as violated. (That file isn't part of this PR, so leaving a note here since this is the change that breaks the claim.)

msrathore-db (commented):
Follow-up review of the latest push (`9257e65`). Verified the new `WithKernelDecimalAsFloat` option end-to-end on the kernel path against a live warehouse: default returns the exact string `"99.98"`, the flag returns `float64 99.98`, and the Thrift-path reject gate correctly covers it. Everything builds and all unit + live kernel E2E tests pass. A few non-blocking items below (2 doc/invariant, 1 test-coverage, 1 carried over from the prior push).

connector.go
```diff
@@ -580,6 +602,21 @@ func kernelExperimental(c *config.Config) *config.KernelExperimentalConfig {
...
+	return func(c *config.Config) {
+		kernelExperimental(c).DecimalAsFloat = asFloat
```
msrathore-db, line 616: **(P3, test gap) Add this option to the Thrift-reject test.** `TestWithKernelOptionsRejectedOnThriftPath` in `kernel_experimental_test.go` enumerates each `WithKernel*` option (certs, skip-hostname, proxy, retry, max-chunks) but wasn't extended with `WithKernelDecimalAsFloat`. The generic `KernelExperimental != nil` gate covers it at runtime (verified), but every sibling option is pinned individually — add the case for consistency and to guard the reject path for this option too.

connector.go
mani-mathur-arch: Fixed in 0d33408 — updated the `TestTopLevelDecimalRendering` comment in arrowscan_parity_test.go to carve out `WithKernelDecimalAsFloat` (off by default, so the default-config invariant still holds), so it no longer reads as violated.

internal/arrowscan/coltype.go
mani-mathur-arch: Done in 0d33408 — added a note here that RawBytes scan type + float64 value under decimal-as-float is a deliberate Thrift-matching exception (Thrift does the same under UseArrowNativeDecimal).

connector.go
mani-mathur-arch: Added in 0d33408 — `WithKernelDecimalAsFloat` is now a case in both TestWithKernelOptionsRejectedOnThriftPath and TestWithKernelTLSOptionsSetExperimental, matching the sibling options.

internal/backend/kernel/spog_routing_test.go
mani-mathur-arch: Fixed the comment in 0d33408 (it now correctly states the file is cgo && databricks_kernel-tagged, so the test runs under the kernel tag / nightly, not default PR CI). Kept it behind the tag rather than moving it out for this PR — happy to do the affectedrows.go-style extraction as a follow-up if you'd prefer the predicate in default CI.

internal/backend/kernel/rows.go
```diff
@@ -284,3 +298,92 @@ func (r *kernelRows) nextBatch() error {
...
+			it.r.iterationErr = err
+		}
+	}
+}
+
+// Next returns the next record (caller owns it) or io.EOF when the stream drains.
+func (it *kernelBatchIterator) Next() (arrow.Record, error) {
+	it.fill()
+	if it.err != nil {
+		return nil, it.err
```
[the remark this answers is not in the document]
mani-mathur-arch: Fixed in e62be8e. `Next()` now surfaces the fetch error exactly once, then sets `done` (and clears `err`) so `HasNext()` returns false afterward and a further `Next()` returns `io.EOF` — a log-and-continue loop can no longer spin. `r.iterationErr` is still seeded (via `fill`) so OnClose telemetry still records the failed stream. Added TestBatchIteratorFetchErrorTerminates and updated the sibling test that pinned the old sticky behavior.

PR #417, opened 2026-07-21
Tag CloudFetch, Arrow schema, and result-fetch errors for telemetry

prathamesh-db:
## What

Attaches source-declared error categories at the error sites on the 
result-materialization paths, so telemetry reports a precise
`error_name` instead of the generic `error` fallback:

| Category | Site |
|----------|------|
| `chunk_download_error` | CloudFetch batch download failures (`internal/rows/arrowbased/batchloader.go`) |
| `arrow_schema_parsing_error` | Arrow schema convert / serialize / read failures (`internal/rows/arrowbased/arrowRows.go`) |
| `result_set_error` | result-page fetch (`internal/rows/rowscanner/resultPageIterator.go`) and result-set metadata fetch (`internal/rows/rows.go`) |
| `statement_execution_timeout` | sentinel `WatchTimeout` in the poll loop (`internal/backend/thrift/backend.go`) |

A fetch aborted via the results context (e.g. on Close) is left **untagged**, so
it still classifies as `cancelled`/`timeout` rather than `result_set_error`,
matching the CloudFetch path.

`statement_execution_timeout` is unreachable in production today (the sole
`Watch` call uses `timeout=0`, so the branch never fires and emits no
telemetry); it is tagged and commented so it classifies correctly if a nonzero
poll timeout is ever enabled.

## Why it's safe

- All tags use the existing `WithCategory` chaining, which returns the **same
  concrete pointer**, so `errors.Is`/`errors.As`, the `dbsqlerr.DBError`
  assertion in the Arrow scan path, `Error()` strings, and sentinel identity
  are all unchanged.
- Each live tag reaches `classifyError` via the row-iteration path
  (`Next` → `iterationErr` → `AfterExecute`), verified by the added tests.

## Also: a latent bug fix

`rows.getResultSetSchema` wrapped the wrong (nil) variable on a
`GetResultSetMetadata` failure, dropping the real cause from the error chain.
Fixed to wrap the actual cause. This doesn't change the telemetry category (the
tag wins regardless), but it restores `Cause()`/`Unwrap()`.

---

Jira: PECOBLR-3537


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Release History
 
 ## Unreleased
-- Enrich telemetry error classification with source-declared error categories. No behavior change yet: the categories are defined and read by `classifyError` but not attached at any error source in this change (databricks/databricks-sql-go#414, #415)
+- Enrich telemetry error classification with source-declared error categories, so CloudFetch download, Arrow schema parsing, and result-fetch failures now report a specific `error_name` (`chunk_download_error`, `arrow_schema_parsing_error`, `result_set_error`) instead of the generic `error` (databricks/databricks-sql-go#414, #415)
```
samikshya-db: nit : this is too low level for an external user. They don't need to know intrinsic details of the driver, just high level changes.

samikshya-db (approved):
just 1 nit, LGTM otherwise. Please re-check using E2E before merging

CHANGELOG.md
prathamesh-db: Simplified to a high-level summary

PR #424, opened 2026-07-23
Tag session-close, statement-close, and rate-limit errors for telemetry

prathamesh-db:
## What

Tags the remaining live error sources so telemetry reports a specific
`error_name` instead of the generic `error`:

| Category | Site |
|----------|------|
| `session_closed` | failed `CloseSession` (`connection.go`, DELETE_SESSION telemetry) |
| `statement_closed` | failed `CloseOperation` RPC (`connection.go`, the 3 CLOSE_STATEMENT sites) |
| `rate_limit_exceeded` | HTTP 429 that survives all retries (`internal/client/client.go` retry `errorHandler`) |

All tags are **telemetry-only** and use the existing `WithCategory` chaining
(same concrete pointer), so `errors.Is`/`errors.As`, `Error()` strings, and the
errors returned to callers are unchanged:
- `session_closed` tags a separate telemetry copy; the returned error stays a
  `driver.ErrBadConn` (`database/sql` relies on that for connection-pool eviction).
- `statement_closed` tags only the telemetry argument; the raw close error still
  flows to the caller/rows.
- `rate_limit_exceeded` is tagged in the retry `errorHandler` — the only place the
  429 status is visible (net/http discards the response on a transport error).
  Scoped to 429 exactly, so 503 keeps its existing classification. This covers the
  retry-exhausted failure only; a per-attempt 429 counter metric is a TODO follow-up.

## Notes for reviewers (by design)
- `statement_closed` is only live on the Thrift backend; the kernel backend's close
  never returns an error, so it can't be tagged there.
- Because the innermost source category wins, a close that fails specifically due to
  a 429 reports `rate_limit_exceeded` rather than `session_closed`/`statement_closed`
  — the more specific category.

## Naming note
`session_closed` has no exact JDBC equivalent (JDBC's nearest is `CONNECTION_CLOSED`);
kept Go-specific here.

---
Jira: PECOBLR-3537


samikshya-db (approved):
make sure you test this e2e as discussed - before merging

PR #425, opened 2026-07-23
test(kernel): add a non-blocking Reyden E2E leg to the nightly

mani-mathur-arch:
## What

Add a **non-blocking Reyden E2E leg** to the nightly. Reyden is the SEA reference engine — the same SEA-via-kernel path as the existing kernel E2E suite, just pointed at a read-only reference warehouse over shared Unity Catalog. Rather than duplicate the 18 `TestKernelE2E*` funcs, this **reuses them verbatim** and routes the connection to Reyden when selected.

- **`kernel_reyden_e2e_test.go`** (new): `isReydenLeg()` (`TEST_WAREHOUSE=reyden` + a reyden path in `DATABRICKS_REYDEN_HTTP_PATH`), `reydenHTTPPath()`, and `skipOnReyden()` — a gate for surfaces Reyden genuinely doesn't support. Inert unless the reyden env is set, so every existing run is unaffected.
- **`kernel_e2e_test.go`**: `kernelTestDBWith` routes `WithHTTPPath` to the reyden warehouse on the reyden leg (same host + PAT). The `variant` data-type subcase skips on reyden — it cannot decode VARIANT over inline Arrow (a reyden-side gap; dbsql-sea and Thrift read it fine). No DDL/DML in this suite, so nothing else needs gating.
- **`nightly-e2e.yml`**: a second, **non-blocking** (`continue-on-error`) run step in the existing `kernel-e2e` job re-runs the kernel E2E funcs with `TEST_WAREHOUSE=reyden` — the kernel lib is already built, so it's just another `go test`. The Thrift-vs-kernel parity funcs are excluded (they compare the kernel against Thrift on one warehouse; meaningless cross-warehouse).

## Why

Give the SEA/kernel path continuous coverage against the Reyden reference engine, surfacing reyden-vs-dbsql divergences. `continue-on-error` because Reyden is a reference engine that can be down or diverge — a reyden-side failure must not fail the nightly (it surfaces in the step log). This mirrors the non-blocking Reyden integration leg in the shared driver-test suite.

`DATABRICKS_REYDEN_HTTP_PATH` falls back to the well-known reyden warehouse id on the same host when the secret is unset; if it and `TEST_WAREHOUSE` resolve empty, the suite simply runs against the normal warehouse (`isReydenLeg()==false`) — a harmless no-op.

## Testing

- `gofmt` clean; `go vet` clean on the `-tags databricks_kernel` build.
- Leg-selection (`isReydenLeg`, fail-safe when path unset, case-insensitive) and the `skipOnReyden` gate unit-tested.
- Nightly YAML parses. Live reyden read assertions pending a healthy reyden endpoint (the staging one is currently returning 500s); routing + skip behavior verified locally.

### DIFF
--- a/.github/workflows/nightly-e2e.yml
+++ b/.github/workflows/nightly-e2e.yml
@@ -244,3 +244,24 @@ jobs:
           CGO_ENABLED=1 go test -tags databricks_kernel -count=1 -v -timeout 30m \
             -run 'TestKernelE2E|TestKernelThriftParity|TestKernelParamsVsThrift' \
             -skip 'TestKernelE2EM2M' .
+
+      # Non-blocking Reyden reference-engine leg: reruns the TestKernelE2E* funcs against
+      # the Reyden warehouse (TEST_WAREHOUSE=reyden + DATABRICKS_REYDEN_HTTP_PATH →
+      # kernelTestDBWith routes there; unsupported surfaces self-skip via skipOnReyden).
+      # continue-on-error because Reyden can be down or diverge from dbsql — a reyden-side
+      # failure surfaces in the log, not the nightly. Parity funcs are excluded (they
+      # compare kernel vs Thrift on one warehouse). Inert if the reyden path secret is
+      # unset: isReydenLeg() is false and the suite just re-runs on the normal warehouse.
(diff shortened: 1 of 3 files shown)


msrathore-db (approved):
LGTM

PR #428, opened 2026-07-30
docs: document Thrift vs SEA/kernel parameter parity

mani-mathur-arch:
## What

Documents how every connection/config parameter behaves across the driver's two execution backends — Thrift (default) and the opt-in SEA/kernel backend — so parameter drift between protocols is visible. Follows the [`adbc-drivers/databricks` C# README](https://github.com/adbc-drivers/databricks/blob/main/csharp/README.md#databricks-specific-properties) model referenced in PECOBLR-3737.

- **README.md** — restructured around the two backends:
  - a **protocol-selection** section (`WithUseKernel` / `useKernel=true`, and the fail-loud `ErrKernelNotCompiled` behavior),
  - a **Building** section covering the difference between the pure-Go Thrift build and the cgo + Rust static-lib kernel build (`make kernel-lib` / `build-kernel` / `test-kernel`, cross-compile caveats),
  - **connection-property tables** whose **Protocol** column marks each parameter `Both` / `Thrift-only` / `SEA-only`, and a call-out of every protocol-specific parameter and behavior/default difference.
- **examples/kernel/main.go** — a runnable example of the SEA/kernel backend (`WithUseKernel`, `WithWarehouseID`, and `errors.Is` detection of `ErrKernelNotCompiled` / `ErrNotSupportedByKernel`). Compiles under the default pure-Go build; requires a `-tags databricks_kernel`, `CGO_ENABLED=1` build to actually select the kernel.

## Incidental fixes (found while writing the docs)

- **doc.go** — the U2M default scopes are now at parity with Thrift (the kernel path forwards `oauth.GetScopes`); the prose still described the older `all-apis + offline_access` set.
- **internal/config/config.go** — the `TelemetryBatchSize` / `TelemetryFlushInterval` default comments said `100` / `5s`, but `DefaultConfig()` resolves an unset value to `200` / `30s`. Comment-only correction.

## Notes for reviewers

- Docs + one example + two comment corrections. No behavior change.
- Every documented default was verified against source (`config.go`, `connector.go`, `kernel_config.go`, `Makefile`), not carried over from the old README.

Addresses PECOBLR-3737.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,30 @@
 # Databricks SQL Driver for Go
 
-
 ![http://www.apache.org/licenses/LICENSE-2.0.txt](http://img.shields.io/:license-Apache%202-brightgreen.svg)
 
-## Description
+A [database/sql](https://golang.org/pkg/database/sql) driver for Databricks SQL. It
+connects to Databricks SQL Warehouses and clusters and runs queries through Go's
+standard `database/sql` interface.
 
-This repo contains a Databricks SQL Driver for Go's [database/sql](https://golang.org/pkg/database/sql) package. It can be used to connect and query Databricks clusters and SQL Warehouses.
+## Contents
 
-## Documentation
+- [Quick start](#quick-start)
+- [Choosing a backend (Thrift vs SEA/kernel)](#choosing-a-backend-thrift-vs-seakernel)
+- [Building](#building)
+- [Connecting](#connecting)
+  - [DSN (Data Source Name)](#dsn-data-source-name)
+  - [Connector object](#connector-object)
+- [Connection properties](#connection-properties)
+- [Authentication](#authentication)
+- [Cloud Fetch](#cloud-fetch)
+- [TLS](#tls)
+- [Proxy](#proxy)
+- [Data types](#data-types)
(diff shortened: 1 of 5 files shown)


mani-mathur-arch:
Note on the ADBC reference pattern (for anyone opening the linked C# README):

The [`adbc-drivers/databricks` C# README](https://github.com/adbc-drivers/databricks/blob/main/csharp/README.md#databricks-specific-properties) expresses protocol-differing defaults with **dual columns** (`| Property | Description | Thrift Default | SEA Default |`). This PR uses a **single `Default` column + a `Protocol` column + prose caveats** instead.

This is a deliberate deviation, not an oversight:
- Go's API is option-based (`WithX(...)`), so most parameters have one default that's shared or clearly attributed in prose — dual default columns would be mostly-duplicated cells.
- Per-protocol default/behavior differences (AC #3) are still called out explicitly in prose: `timeout` → `STATEMENT_TIMEOUT`, `maxRows`/`useCloudFetch` inert on kernel, DECIMAL rendering, port-443-only, etc.

Happy to switch to dual default columns if reviewers prefer strict visual parity with the C# reference.

PR #444, opened 2026-08-17
feat(kernel): support federated token providers

vuanhphung:
Reuses `WithFederatedTokenProvider*` on the kernel backend. The driver snapshots one base-provider token for kernel-side PAT federation; `WithFederatedTokenProviderAndClientID` also forwards the SP-wide client ID. Because the kernel cannot refresh the provider, an expired token requires a new connection.

Pins `KERNEL_REV` to the federation-capable kernel. The kernel bump also requires aligning one retry-setter assertion with its clamp behavior.

Validation: `go test ./...`, `make test-kernel`, and `make lint`.


eric-wang-1990:
For Go, we should also reuse-existing FederationProvider to avoid confusion as much as possible. Can we thread that through?
One path is we just get the token from the provider in the driver, then pass it to kernel as PAT with federation. That will fail in the case the token expired.

Long term is support custom token provider in kernel which is out of scope for now

vuanhphung:
@eric-wang-1990 PTAL!

kernel_config.go
```diff
@@ -292,12 +297,26 @@ func resolveKernelProxy(cfg *config.Config, kc *kernel.Config) {
...
 	switch a := cfg.Authenticator.(type) {
+	case *federatedTokenAuthenticator:
+		token, err := a.provider.GetToken(ctx)
+		if err != nil {
+			return kernel.Auth{}, fmt.Errorf("databricks: failed to get a federated token for the kernel backend: %w", err)
+		}
+		if token == nil || token.AccessToken == "" {
+			return kernel.Auth{}, errors.New("databricks: the federated token provider returned an empty token")
+		}
+		return kernel.Auth{Mode: kernel.AuthPAT, Token: token.AccessToken, ClientID: a.clientID}, nil
```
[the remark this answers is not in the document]
vuanhphung: Kernel will do mandatory exchange unless it's same issuer or non-JWT

connector.go
```diff
@@ -567,7 +574,10 @@ func WithFederatedTokenProvider(baseProvider tokenprovider.TokenProvider) ConnOp
 		if baseProvider != nil {
 			// Wrap with federation provider that auto-detects need for token exchange
 			federationProvider := tokenprovider.NewFederationProvider(baseProvider, c.Host)
-			c.Authenticator = tokenprovider.NewAuthenticator(federationProvider)
+			c.Authenticator = &federatedTokenAuthenticator{
```
eric-wang-1990, line 577: Does this change affect existing Go Thrift side token-exchange feature?

vuanhphung: @eric-wang-1990 nope, this is just adding more metadata info. You can think of it like this
```
Before:

c.Authenticator = oldAuthenticator

After:

c.Authenticator = wrapper{
      Authenticator: oldAuthenticator,
      // kernel-only metadata
 }
```

calling c.Authenticator.Authenticate() still delegate to oldAuthenticator

eric-wang-1990 (approved):

PR #447, opened 2026-08-20
feat(kernel): JWT private-key M2M auth via WithJWTPrivateKeyM2M

rahuls-db:
## What

Adds OAuth **machine-to-machine auth with a JWT private-key client assertion** (RFC 7523) on the kernel backend (`WithUseKernel(true)`), via a new `WithJWTPrivateKeyM2M(JWTPrivateKeyM2MConfig{...})` connector option. The kernel signs a short-lived assertion with the service principal's private key instead of sending a client secret and owns the token lifecycle.

Companion to the kernel-side feature (databricks-sql-kernel #249; napi `token_url` in #275) and the parallel databricks-sql-python / databricks-sql-nodejs changes.

## How

- **`auth/oauth/jwtm2m`** — new authenticator satisfying the internal `JWTM2MCredentialsProvider` interface (same structural-interface pattern as M2M/U2M). `Authenticate` returns a clear kernel-only error on the pure-Go Thrift path (JWT signing is the kernel's job).
- **`internal/backend/kernel`** — `AuthJWTM2M` mode + `Auth` JWT fields + `JWTM2MCredentialsProvider`; `setAuth` maps it to the `kernel_session_config_set_auth_m2m_jwt` C setter (8 args incl. `token_url`).
- **`kernel_config.go`** — `resolveKernelAuth` resolves the JWT authenticator to the descriptor (scopes + `token_url` forwarded, unlike the scopes-less shared-secret M2M setter).
- **`connector.go`** — `WithJWTPrivateKeyM2M(JWTPrivateKeyM2MConfig{...})`, a struct option (not positional args) so the many string fields can't be transposed.
- **`KERNEL_REV`** bumped to a build including the JWT C-ABI setter; the `TestSetRetry` max<min case updated to the kernel's newer correct-and-warn behavior.

## Usage

```go
connector, _ := dbsql.NewConnector(
    dbsql.WithServerHostname("adb-….azuredatabricks.net"),
    dbsql.WithHTTPPath("/sql/1.0/warehouses/…"),
    dbsql.WithJWTPrivateKeyM2M(dbsql.JWTPrivateKeyM2MConfig{
        ClientID: "<sp-client-id>",
        KeyFile:  "/path/private_key.pem",
        Kid:      "<kid>",
        TokenURL: "https://login.microsoftonline.com/<tenant>/oauth2/v2.0/token",
        Scopes:   []string{"<databricks-resource-id>/.default"},
    }),
    dbsql.WithUseKernel(true),
)
```

## Testing

- Unit: `jwtm2m` authenticator tests; `resolveKernelAuth` JWT cases in `kernel_config_test.go`.
- cgo: `set_auth_m2m_jwt` marshalling case in `TestSetAuthByMode` (tagged build).
- E2E: env-gated `TestKernelE2EJWTM2MSelect1` (tagged build). **Verified end-to-end** against an Azure Databricks warehouse: `SELECT 1` = 1 via the kernel backend, authenticated by Entra ID with a JWT private-key assertion.

## Not in scope

DSN-string configuration (`authType=oauthm2m-jwt` + key-file/passphrase/tokenUrl query params) is deferred to a follow-up — the programmatic `WithJWTPrivateKeyM2M` is the primary surface (passing a key-file path + passphrase through a DSN query string is awkward, and there is no cross-driver DSN precedent for JWT). Use the option.


internal/backend/kernel/auth.go
```diff
@@ -10,9 +10,10 @@ package kernel
 type AuthMode int
 
 const (
-	AuthPAT AuthMode = iota // personal access token
-	AuthM2M                 // OAuth client-credentials (client id + secret)
-	AuthU2M                 // OAuth user-to-machine (browser/PKCE; kernel-owned flow)
+	AuthPAT    AuthMode = iota // personal access token
+	AuthM2M                    // OAuth client-credentials (client id + secret)
+	AuthU2M                    // OAuth user-to-machine (browser/PKCE; kernel-owned flow)
+	AuthJWTM2M                 // OAuth client-credentials via a JWT private-key client assertion
```
eric-wang-1990, line 16: Same here, we do not need it in Go if it does not support it now.

PR #450, opened 2026-08-21
feat(kernel): forward logs through shared logger

vuanhphung:
Stacked on #444. Depends on databricks/databricks-sql-kernel#278.

Routes Rust kernel tracing records through the driver's existing zerolog sink, so `logger.SetLogOutput` applies equally to Thrift, Go kernel-binding, and Rust kernel logs. A synchronized writer proxy lets existing logger snapshots follow later output changes safely.

The design mirrors Node's kernel-log bridge: Node uses a bounded 4096-record queue and a drain thread to forward batches through a `ThreadsafeFunction` into `DBSQLLogger`. Go copies each synchronous C callback record into its own bounded 4096-record channel; one goroutine blocks when the channel is empty and forwards records immediately when they arrive. There is no polling, timer, or minimum batch size. A full queue drops the newest record rather than blocking a kernel thread; a FIFO barrier supports explicit test flushing.

Adds level mapping, output retargeting, panic containment, drop accounting, and end-to-end same-file coverage. Validated with the default and kernel-tagged Go suites, focused race tests, `go vet`, and golangci-lint.

---

_This PR was created with [GitHub MCP](http://go/mcps)._


internal/backend/kernel/log_callback.go
```diff
@@ -0,0 +1,85 @@
...
+#include <stdlib.h>
+#include "databricks_kernel.h"
+
+// cgo generates this declaration with mutable char pointers; the kernel lends
+// read-only strings, so cast once in the C adapter to match KernelLogCallback.
+void kernelLogTrampoline(char* level, char* target, char* message, void* user_data);
+static KernelLogCallback kernel_log_cb(void) {
+    return (KernelLogCallback)kernelLogTrampoline;
+}
+static void* kernel_handle_to_ctx(uintptr_t h) { return (void*)h; }
```
vikrantpuppala: **High:** `cgo.Handle` is an integer token, not a pointer. The `runtime/cgo` documentation explicitly says it is unsafe to coerce a `Handle` to `unsafe.Pointer`; this helper fabricates a small `void *`, and the exported Go callback receives it in a pointer-typed slot. A GC can detect an invalid pointer and terminate via `runtime.throw`, which the `recover` below cannot catch.

Because this callback is already process-global, the simplest fix is a package-global immutable sink with `NULL` `user_data`. Otherwise allocate a `uintptr_t` cell in C memory and pass its real address. An exact-signature C adapter can also remove the incompatible `char *`/`const char *` function-pointer cast above.

internal/backend/kernel/log_callback.go
```diff
@@ -0,0 +1,85 @@
...
+	// as a dropped diagnostic; logging must never fail a query.
+	defer func() { _ = recover() }()
+	if userData == nil {
+		return
+	}
+	sink, ok := cgo.Handle(uintptr(userData)).Value().(*logSink)
+	if !ok || sink == nil {
+		return
+	}
+	sink.forward(C.GoString(level), C.GoString(target), C.GoString(message))
```
vikrantpuppala: **High design risk:** the linked kernel ABI invokes this callback synchronously on the emitting kernel thread and requires prompt return/no kernel re-entry. `sink.forward` immediately runs zerolog and the arbitrary writer installed through `SetLogOutput` before returning. A slow/network/pipe writer therefore stalls the kernel operation; a writer that persists through this driver can re-enter the kernel ABI.

Please keep the callback minimal: copy the three strings, non-blockingly enqueue an owned record into a bounded Go channel, and drain it from one Go logging goroutine with an explicit drop counter/policy. This preserves arbitrary writers and live retargeting without putting user I/O on a kernel thread.

logger/logger.go
```diff
@@ -14,6 +15,27 @@ type DBSQLLogger struct {
...
+// different goroutines or native kernel threads.
+type sharedOutput struct {
+	mu sync.Mutex
+	w  io.Writer
+}
+
+func (o *sharedOutput) Write(p []byte) (int, error) {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	return o.w.Write(p)
```
vikrantpuppala: **Medium:** this holds the destination-routing mutex while executing user-owned `Write` code. If the writer blocks, every Go/kernel log and `SetLogOutput` blocks behind it; if `Write` calls `SetLogOutput` or logs through this logger, it self-deadlocks. This also means a stuck writer cannot be replaced.

Publish a synchronized writer wrapper under a short-held lock/atomic pointer, then invoke it after releasing the routing lock. Per-destination serialization can live in the wrapper (for example, `zerolog.SyncWriter`) so retargeting never waits on arbitrary user code.

logger/logger.go
```diff
@@ -14,6 +15,27 @@ type DBSQLLogger struct {
...
+}
+
+func (o *sharedOutput) set(w io.Writer) {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	o.w = w
```
vikrantpuppala: **Medium compatibility regression:** previously `SetLogOutput(nil)` called `Logger.Output(nil)` → `zerolog.New(nil)`, which normalizes nil to `io.Discard`. This now stores a nil interface, and the next ordinary Go/Thrift log panics at `o.w.Write` (kernel callback logs happen to be swallowed by the outer recover).

Please normalize nil to `io.Discard` and add a regression test that emits an enabled Go log after `SetLogOutput(nil)`.

logger/logger.go
```diff
@@ -14,6 +15,27 @@ type DBSQLLogger struct {
 	zerolog.Logger
 }
 
+// sharedOutput is the process-wide destination behind every driver logger,
+// including logger values derived before a later SetLogOutput call. Serializing
+// writes keeps Go and kernel-callback records intact when they arrive from
+// different goroutines or native kernel threads.
+type sharedOutput struct {
```
vikrantpuppala, line 33: **Medium compatibility regression:** the root zerolog instance is now constructed once over `*sharedOutput`, which implements only `io.Writer`. Zerolog therefore permanently wraps it as a plain writer. If a caller later supplies a `zerolog.LevelWriter` (for example `SyslogLevelWriter` or `MultiLevelWriter`), it receives `Write` instead of `WriteLevel`, losing severity-aware routing.

Make the stable proxy implement `WriteLevel(zerolog.Level, []byte)` and delegate to the current destination's `zerolog.LevelWriter` when available, falling back to `Write`. Please add a recording `LevelWriter` test.

vikrantpuppala (approved):

PR #457, opened 2026-08-25
feat(kernel): forward request timeout

vuanhphung:
Forward the driver's existing `ClientTimeout` to the kernel request-timeout C ABI introduced in databricks/databricks-sql-kernel#288. Durations are converted to milliseconds, positive sub-millisecond values round up, and zero selects the kernel's 120-second default.

The kernel pin and connection telemetry are updated. Both the normal and cgo kernel test suites pass, and golangci-lint reports no issues.

---

_This PR was created with [GitHub MCP](http://go/mcps)._


eric-wang-1990:
Do you need to update the CONNECTION_PARAMETER file to reflect the new change?

vuanhphung:
Updated `CONNECTION_PARAMETERS.md` in c53cbb4 to document that kernel requests inherit the internal `ClientTimeout` (900s by default), that zero keeps the kernel's 120s default, and that there is currently no DSN/`WithX` option for it.

---

CONNECTION_PARAMETERS.md
eric-wang-1990, line 91: Is this the timeout we are wiring through? Why this is left unchanged?

CONNECTION_PARAMETERS.md
```diff
@@ -95,6 +95,11 @@ Notes for the SEA/kernel backend:
 | `timezone` | `WithSessionParams(timezone=…)` | ✅ | ✅ | | Session time zone (e.g. `America/Los_Angeles`). |
 | `enableMetricViewMetadata` | `WithEnableMetricViewMetadata` | ✅ | ⚠️ | `false` | Enables metric-view metadata (sets `spark.sql.thriftserver.metadata.metricview.enabled=true`). Both paths forward the **identical** conf; the kernel allowlists this key and sends it verbatim — it is **not** rejected driver- or kernel-side. Whether it takes effect on the SEA/kernel path depends on server-side SEA support (a `⚠️` pending confirmation against a live warehouse; PECOBLR-4142 / PECOBLR-4153). |
 
+**Kernel request timeout.** Kernel HTTP requests inherit the driver's internal
```
eric-wang-1990: Let's make sure to put this information in the param table, that is more clear than a paragraph here.

CONNECTION_PARAMETERS.md
vuanhphung: This populates the timeout as a field in execute statement request. It's different from the http client request timeout we want to pass through

CONNECTION_PARAMETERS.md
vuanhphung: Actually I'm thinking of reverting this change. this is an internal timeout of the go driver. it's not exposed as connection param. I think documenting in doc.go is enough. Wdyt?

eric-wang-1990: Yeah if it not configurable we can remove from here.

vuanhphung:
@eric-wang-1990 removed from the md doc again, PTAL!

PR #459, opened 2026-08-27
fix(security): bump golang.org/x/mod to v0.40.0 (CVE-2026-56864/56865)

msrathore-db:
## Problem

The **Security Scan** CI job (OSV-Scanner v2.3.8) fails on two unsuppressed advisories against `golang.org/x/mod@v0.8.0`:

| ID | Alias | Summary | Fixed in |
|---|---|---|---|
| GO-2026-6179 | CVE-2026-56865 | Transparency-log tile verification bypass in `golang.org/x/mod/sumdb/tlog` | x/mod v0.40.0 |
| GO-2026-6180 | CVE-2026-56864 | Unrelated/unauthenticated hashes in `Lookup` in `golang.org/x/mod/sumdb` | x/mod v0.40.0 |

Both are scoreless (UNKNOWN) GHSA/GO advisories on a **non-stdlib** package, which the PR gate treats as blocking (fail-closed). This is pre-existing on `main` — `x/mod` is a long-standing transitive dependency (via `apache/arrow/go/v12`, `golang.org/x/tools`, and `gotestsum`).

## Fix

Bump `golang.org/x/mod` to **v0.40.0** (the fixed version for both advisories).

`x/mod` is a purely transitive/indirect dependency — the driver does **not** import `x/mod` or its vulnerable `sumdb` packages. Verified via:
- `go mod why golang.org/x/mod` → *"main module does not need package golang.org/x/mod"*
- `go list -deps ./...` → `x/mod` absent from the build graph

So there is no real exposure, but a fixed version exists, so we bump rather than add an `[[IgnoredVulns]]` suppression (repo convention prefers a real fix when one is available).

The bump pulls the minimal consistent set of `x/mod`'s transitive indirect requirements forward (`x/sync`, `x/sys`, `x/tools`, plus a new `x/telemetry` indirect). The `go` directive stays at **1.25.0** (no floor increase for consumers).

## Verification

- `go build ./...`, `go vet ./...` — pass
- `go mod tidy` — clean, no further changes
- `go mod verify` — all modules verified
- OSV-Scanner v2.3.8 (`scan source --lockfile=go.mod --config=osv-scanner.toml`): **blocking findings 2 → 0**. GO-2026-6179 / GO-2026-6180 gone; remaining findings are all report-only stdlib advisories (non-blocking by design).


msrathore-db:
Superseded by a same-repo PR so the JFrog-gated checks (Lint, Test and Build) run — fork PRs can't obtain the OIDC token those steps need. The Security Scan fix itself is verified here (green).

Issue #314, opened 2026-01-13
float64 parameters incorrectly mapped to SqlFloat (32-bit)


Mandukhai-Alimaa:
When passing float64 values as query parameters, they are incorrectly mapped to SqlFloat instead of SqlDouble, causing precision loss and overflow/underflow issues.

https://github.com/databricks/databricks-sql-go/blob/main/parameters.go#L151

PR #470, opened 2026-09-03
feat(kernel): send per-statement query tags on the kernel backend

rahuls-db:
## What

Brings the **kernel (SEA) backend** to parity with the Thrift backend for
**per-statement query tags**. The Thrift backend already reads tags from the
request context and sends them via `confOverlay["query_tags"]`; the kernel backend
ignored them because the kernel C ABI had no per-statement tags setter. It now does
(`kernel_statement_set_query_tags`, merged as
databricks-sql-kernel `21504eae`), so this wires it up.

**No new user-facing API** — the same `driverctx.NewContextWithQueryTags(ctx, …)`
that already works on Thrift now also works on the kernel backend.

## How

- **`internal/backend/kernel/operation.go`**: after binding parameters and before
  execute, read `driverctx.QueryTagsFromContext(ctx)`, serialize with the shared
  `querytags.Serialize`, and pass the wire string to `kernel_statement_set_query_tags`.
  Empty/absent → skipped.
- **Interior-NUL guard** (`internal/backend/kernel/querytags_nul.go`):
  `checkQueryTags`/`errQueryTagsNUL` rejects a serialized tags string with an
  interior NUL before the length-less C setter would silently truncate it — the
  query-tags counterpart to `checkQueryText` (`set_sql`) and `checkParamValue`
  (`bind_parameter`), keeping the kernel path at parity with the whole-string Thrift
  path. Pure Go, unit-tested under `CGO_ENABLED=0` (`querytags_nul_test.go`).
- **`KERNEL_REV` + committed header**: bumped to `21504eae` and the C-ABI header
  synced to that rev. The `databricks_kernel` CI leg source-builds the kernel `.a`
  from `KERNEL_REV` (`make kernel-lib`), so this is what lets it link the new setter.
  The header sync also carries the other C-ABI decls that landed since the prior pin
  (max_connections, telemetry circuit breaker, a result-stream canceller type, the
  `kernel_session_test` timeout param); the Go backend calls none of them, so no
  binding code changes were needed.

## Note on the published bindings

`go.mod` still pins the `databricks-sql-kernel-bindings` modules at `v1.0.0` — a
matching bindings release (built from a kernel rev ≥ `21504eae`) is a **fast-follow**
for `go get -tags databricks_kernel` consumers. CI and this repo's tagged tests build
the kernel from `KERNEL_REV` and are unaffected; the pure-Go default build never
touches the kernel.

## Validation (local)

Built the kernel static/shared lib from `21504eae` (the new `KERNEL_REV`) and, with
the synced header, ran the `databricks_kernel`-tagged path against it:
- `go build` / `go test` (`-tags databricks_kernel`, CGO on) — pass.
- `checkQueryTags` guard test under `CGO_ENABLED=0` — pass.
- Default pure-Go `go build ./...` + `gofmt` — clean.
- **Live warehouse E2E**: connected on the kernel backend with statement-level tags
  in context and ran a tagged query end to end — statement id
  `01f1a7e7-5834-15c7-bd90-bb1c7e0f144d` on warehouse `153fb3a25b8051a7`.


internal/backend/kernel/include/databricks_kernel.h
```diff
@@ -656,6 +656,20 @@ KernelStatusCode kernel_statement_bind_parameter(kernel_statement_t* stmt,
...
+ * statement's execution for cost attribution and tracking. `query_tags` is the
+ * serialized wire form: comma-separated `key:value` pairs, with a bare `key`
+ * (no colon) for a valueless tag, e.g. "team:eng,job:etl,production"; backslash
+ * escapes the separators in a value ("\:", "\,", "\\"). This is the same string
+ * a host already builds for the Thrift `confOverlay` path, so it is passed
+ * through unchanged and re-parsed into the native SEA query_tags array at
+ * execute time. An empty string is a no-op; calling again replaces the
+ * statement's tags. `query_tags` must be a NUL-terminated UTF-8 string.
+ */
+KernelStatusCode kernel_statement_set_query_tags(kernel_statement_t* stmt,
```
vuanhphung, line 704: I think you need to bump KERNEL_REV after kernel change is merged

internal/backend/kernel/operation.go
```diff
@@ -78,6 +80,24 @@ func (k *KernelBackend) execute(ctx context.Context, req backend.ExecRequest) (b
 		return &kernelOp{}, fmt.Errorf("kernel: bind params: %w", toStatementError(err))
 	}
 
+	// Per-statement query tags from context — the same source and serializer the
+	// Thrift backend uses; here the serialized wire string goes to the kernel's
+	// per-statement setter instead of Thrift confOverlay. Empty/absent → skip.
+	if queryTags := driverctx.QueryTagsFromContext(ctx); len(queryTags) > 0 {
+		if serialized := querytags.Serialize(queryTags); serialized != "" {
+			tags := newCStr(serialized)
```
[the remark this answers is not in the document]
rahuls-db: Added in 6cea7c5: `checkQueryTags`/`errQueryTagsNUL` in a new untagged `querytags_nul.go` (pure Go, tested under CGO_ENABLED=0 in `querytags_nul_test.go`), mirroring `checkQueryText`/`checkParamValue`. operation.go now calls it on the serialized string and returns `kernel: %w` before `newCStr`, so an interior NUL fails loudly instead of silently truncating — keeping the kernel path at parity with the whole-string Thrift path.

internal/backend/kernel/include/databricks_kernel.h
rahuls-db: Agreed — that's the remaining step. #314 is now merged (kernel main @ 21504eae). The bump is gated on one more thing: the databricks-sql-kernel-bindings modules must be rebuilt at a rev that includes 21504eae and published as a new version (they're still at v1.0.0, built from the old rev dd810d6). Once that lands I'll bump KERNEL_REV -> 21504eae and the go.mod bindings versions to the new release in a follow-up commit here, which is what lets the databricks_kernel CI leg link. Do you know who owns cutting the bindings release?

rahuls-db: Done in 446f9dd: bumped KERNEL_REV to 21504eae (kernel main w/ the merged C ABI) and synced the committed header to that rev. The databricks_kernel CI leg source-builds the kernel from KERNEL_REV (make kernel-lib), so it links the new setter without a bindings release. go.mod bindings stay v1.0.0 — a matching bindings release is a fast-follow for go get -tags databricks_kernel consumers (noted in the description).

vuanhphung (approved):

